Issue #16, opened 2018-01-17
Make the extension work with VS 2017


mikkelbu:
Probably by following the description in https://docs.microsoft.com/en-us/visualstudio/extensibility/how-to-roundtrip-vsixs

Remember to update documentation about building

CharliePoole:
Have you considered going the nuget-only route?

jnm2:
@mikkelbu I found the built-in NuGet + VSIX analyzer template quite finicky and followed the guide for setting up my own from scratch to get away from packages.config. (https://roslyn-analyzers.readthedocs.io/en/latest/how-to-start.html#the-pro-way) It provides the flexibility I've come to demand from the new csproj format. If this sounds interesting, here's what it looks like in one of my analyzer projects:

https://github.com/jnm2/SourceHeaderAnalyzer/blob/22093218482b68c5c4e68b1b244833a175c06ba3/src/SourceHeaderAnalyzer/SourceHeaderAnalyzer.csproj

VSIX project (only for F5 debug): https://github.com/jnm2/SourceHeaderAnalyzer/tree/22093218482b68c5c4e68b1b244833a175c06ba3/src/SourceHeaderAnalyzer.Debug

jnm2:
Oh wait, that's not the best example because I haven't packed it yet. Comparing one of the analyzer projects at work, this is what the analyzer csproj would like like when ready for use with `/t:Pack`:

```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netstandard1.4</TargetFramework>

    <Version>1.0.0</Version>
    <Authors>jnm2</Authors>
    <Copyright>Copyright © 2018 Joseph Musser</Copyright>
    <AssemblyTitle>Foo code analyzers</AssemblyTitle>
    <Description>$(AssemblyTitle)</Description>
    <PackageTags>Foo analyzer</PackageTags>
    <DevelopmentDependency>True</DevelopmentDependency>
    <BuildOutputTargetFolder>analyzers\dotnet</BuildOutputTargetFolder>

    <LangVersion>latest</LangVersion>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.3.2" />
  </ItemGroup>

</Project>
```

ChrisMaddock:
@mikkelbu - this may be interesting for your decision:

https://github.com/wachulski/nunit-migrator/issues/19

PR #19, opened 2018-03-15
New project format and netstandard for analyzer project

JohanLarsson:


mikkelbu:
Hi @JohanLarsson thanks for the contribution. I've been going a bit back and forth on whether to use the new project format (see e.g. https://github.com/nunit/nunit.analyzers/issues/16), but I think that I've come to the conclusion that the new format is the way to go. Hopefully, I'll have time to look at this PR tomorrow evening, as the last week was completely packed.

JohanLarsson:
If nothing else the new format diffs nicer. I moved to new format for all my analyzer projects and it has worked ok. Only issue really is the vsix that needs a few hax.
I have not tested if `dotnet pack` detects and packs an analyzer project correctly, would be nice if it did, then the nuget stuff could be removed.

jnm2:
@JohanLarsson I've been using `dotnet pack` which works great with this. Example at https://github.com/nunit/nunit.analyzers/issues/16#issuecomment-358645894.

src/nunit.analyzers/nunit.analyzers.csproj
```diff
@@ -1,127 +1,18 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <MinimumVisualStudioVersion>11.0</MinimumVisualStudioVersion>
```
mikkelbu: Should this still be in the project-file?

src/nunit.analyzers.playground/nunit.analyzers.playground.csproj
```diff
@@ -1,68 +1,24 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\packages\NUnit.3.10.1\build\NUnit.props" Condition="Exists('..\packages\NUnit.3.10.1\build\NUnit.props')" />
```
mikkelbu: I had to remove this line and the `Target` element below (with the name `EnsureNuGetPackageBuildImports`) otherwise I got complaints about "This project references NuGet package(s) that are missing on this computer. ...The missing file is ..\packages\NUnit.3.10.1\build\NUnit.props." when compiling.

src/nunit.analyzers/nunit.analyzers.csproj
JohanLarsson: Not sure, that is why I left it there.

jnm2: This file will only open in VS2017 or later, so it could be removed.

PR #24, opened 2018-03-18
Add .editorconfig fix #22.

### LINKED ISSUE #22: Add editorconfig
Depending on #20

JohanLarsson:
A couple of defaults, indent_style = tab makes Visual Studio pick up that this sln uses tabs.

Personally I prefer:
indent_style = space
indent_size = 4

But convention >> preference.
Reason I prefer spaces is that it renders nicer when reading code on github in the browser.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -16,21 +16,31 @@ charset = utf-8
 [*.{proj,csproj,vcxproj,xproj,json,config,nuspec,xml}]
 indent_size = 2
 
-
+# Sort using and Import directives with System.* appearing first
+dotnet_sort_system_directives_first = true
 
 [*]
 # https://github.com/nunit/docs/wiki/Coding-Standards#namespace-class-structure-interface-enumeration-and-method-definitions
-csharp_indent_braces = false
-csharp_new_line_before_catch = true
+csharp_new_line_before_open_brace = all
 csharp_new_line_before_else = true
+csharp_new_line_before_catch = true
 csharp_new_line_before_finally = true
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_within_query_expression_clauses = true
 
 # https://github.com/nunit/docs/wiki/Coding-Standards#spaces
 csharp_space_after_cast = false
+csharp_space_after_colon_in_inheritance_clause = true
 csharp_space_after_comma = true
+csharp_space_after_dot = false
 csharp_space_after_keywords_in_control_flow_statements = true
+csharp_space_after_semicolon_in_for_statement = true
 csharp_space_around_binary_operators = before_and_after
 csharp_space_around_declaration_statements = do_not_ignore
+csharp_space_before_colon_in_inheritance_clause = true
+csharp_space_before_comma = false
+csharp_space_before_dot = false
 csharp_space_before_open_square_brackets = false
 csharp_space_before_semicolon_in_for_statement = false
 csharp_space_between_empty_square_brackets = false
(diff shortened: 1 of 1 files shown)


mikkelbu:
Hi @JohanLarsson thanks for the PR, but my plan was to use - more or less - the same .editorconfig as the nunit-project (https://github.com/nunit/nunit/blob/master/.editorconfig) and then make all the code satisfy the rules.
Probably I'll change the suggestions to either warning or errors in the .editorconfig-file (as there is less code to update in this project). Also I noticed that you have some tags - like `csharp_style_pattern_matching_over_is_with_cast_check` - that I did not know about, so perhaps we should also include those.

CharliePoole:
@JohanLarsson One question... if you prefer spaces and our coding standard says to use spaces, why use tabs? To be fair, I don't know if that standard applies to this project, however.

JohanLarsson:
I just saw tabs in this project so I added an editorconfig with tabs. Did not know about the editorconfig in the main project.

jnm2:
Tabs render weird in the browser and the git CLI. Because of this, when I'm making changes to files that have tabs, I just save each file in the editor first to trigger the .editorconfig whitespace conversion and commit that. Then the actual changes go in a separate commit so that review is easier.

I almost wonder if it's worth doing that now along with the .editorconfig addition? The only downside is that it's another line of noise in git blames.

JohanLarsson:
From what I have seen this project is still in pretty early stages, no releases etc. I say add stylecop and editorconfig and fix all the formatting. After that style will be uniform and diffs will have minimal noise.
I can do the work and PR it, no problem.

mikkelbu:
I don't worry about git blame, since we only have 22 commits 😄, so I much agree with @JohanLarsson. The only thing I worry about is the interplay between editorconfig and stylecop (I have not used StyleCopAnalyzers yet and only know stylecop from back in the days) - how easy is it to keep these in sync? and are there differences between what can be configured?

JohanLarsson:
@mikkelbu It works very well, I have that combo for all things. Keeping in sync is not an issue as both are pretty much configure once and add to source control.

JohanLarsson:
I'll update it with the one in the main project + the new tags.

mikkelbu:
@JohanLarsson Ok. Sounds good (and please also change the "suggestions" in the configuration to "errors" if it does not require too much work - otherwise we will do this later).

JohanLarsson:
Ok, started over and committed the defaults from main project, then sorted & added tags in a second commit.

> and please also change the "suggestions" in the configuration to "errors"
Does this mean you want all the code fixed to be according to the editorconfig?

I can do it no problem but think it is better in separate pull request(s).

.editorconfig
```diff
@@ -0,0 +1,110 @@
...
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+charset = utf-8
+
+[*.{proj,csproj,vcxproj,xproj,json,config,nuspec,xml}]
+indent_size = 2
+
+# Sort using and Import directives with System.* appearing first
+dotnet_sort_system_directives_first = true
```
mikkelbu: The key `dotnet_sort_system_directives_first` also appear on line 86. I don't think it makes sense here?

.editorconfig
```diff
@@ -0,0 +1,110 @@
...
+
+[*]
+# https://github.com/nunit/docs/wiki/Coding-Standards#namespace-class-structure-interface-enumeration-and-method-definitions
+csharp_new_line_before_open_brace = all
+csharp_new_line_before_else = true
+csharp_new_line_before_catch = true
+csharp_new_line_before_finally = true
+csharp_new_line_before_members_in_object_initializers = true
+csharp_new_line_before_members_in_anonymous_types = true
+csharp_new_line_within_query_expression_clauses = true
```
mikkelbu: I think that this should be `csharp_new_line_between_query_expression_clauses` (see e.g. dotnet/corefx#26767)

.editorconfig
```diff
@@ -0,0 +1,110 @@
...
+csharp_space_before_open_square_brackets = false
+csharp_space_before_semicolon_in_for_statement = false
+csharp_space_between_empty_square_brackets = false
+csharp_space_between_method_call_empty_parameter_list_parentheses = false
+csharp_space_between_method_call_name_and_opening_parenthesis = false
+csharp_space_between_method_call_parameter_list_parentheses = false
+csharp_space_between_method_declaration_empty_parameter_list_parentheses = false
+csharp_space_between_method_declaration_name_and_open_parenthesis = false
+csharp_space_between_method_declaration_parameter_list_parentheses = false
+csharp_space_between_parentheses = false
```
mikkelbu: As far as I could google (MicrosoftDocs/visualstudio-docs#513), this has the same effect as `none` (as in the original file), but let us just keep `false` 😄

.editorconfig
```diff
@@ -0,0 +1,110 @@
...
+csharp_indent_case_contents = true
+csharp_indent_switch_labels = true
+csharp_indent_labels = flush_left
+
+# Always use "this." and "Me." when applicable; let StyleCop Analyzers provide the warning and fix
+dotnet_style_qualification_for_field = true:suggestion
+dotnet_style_qualification_for_property = true:suggestion
+dotnet_style_qualification_for_method = true:suggestion
+dotnet_style_qualification_for_event = true:suggestion
+
```
mikkelbu: Actually, I prefer to leave `this.` out if it is not necessary. So can we go for `false:suggestion` instead?

mikkelbu:
> > and please also change the "suggestions" in the configuration to "errors"
>
> Does this mean you want all the code fixed to be according to the editorconfig?
Only if it is not too much work (otherwise we will do it in smaller chunks

> I can do it no problem but think it is better in separate pull request(s).
I think you are right 👍 .

.editorconfig
JohanLarsson: I scanned the code and `this.` seems to be used, do you still want it changed? Personally I prefer `this.` but that is not very important.

mikkelbu: I don't feel that strongly about this. So if the original code was written with `this.` then let us keep it as it is.

Issue #48, opened 2018-06-12
Async test method must have non-generic Task return type when no result is expected


mikkelbu:
We should warn about the following problem

**Message: Async test method must have non-generic Task return type when no result is expected**
```csharp
    [Test]
    public async Task<int> AsyncGenericTask()
    {
      return await Task.FromResult(1);
    }
```
and
```csharp
    [Test]
    public Task<int> NonAsyncGenericTask()
    {
      return Task.FromResult(1);
    }
```
and
**

PR #69, opened 2018-08-19
Added warning: Async test method must have non-void return type

### LINKED ISSUE #7: Feature request: Look for async void tests
Since it's not allowed to have `async void` tests, there should be an analyzer that looks for these tests and autofixes them to be `async Task`.

304NotModified:
Fixes #7

### DIFF
--- a/src/nunit.analyzers.tests/Targets/TestCaseUsage/TestCaseUsageAnalyzerTestsAnalyzeWhenReturnTypeIsAsyncVoid.cs
+++ b/src/nunit.analyzers.tests/Targets/TestCaseUsage/TestCaseUsageAnalyzerTestsAnalyzeWhenReturnTypeIsAsyncVoid.cs
@@ -1,0 +1,14 @@
+using System.Threading.Tasks;
+using NUnit.Framework;
+
+namespace NUnit.Analyzers.Tests.Targets.TestCaseUsage
+{
+    class TestCaseUsageAnalyzerTestsAnalyzeWhenReturnTypeIsAsyncVoid
+    {
+        [Test]
+        public async void AsyncVoid()
+        {
+            await Task.Delay(1); // To avoid warning message
+        }
+    }
+}
--- a/src/nunit.analyzers/Constants/NunitFrameworkConstants.cs
+++ b/src/nunit.analyzers/Constants/NunitFrameworkConstants.cs
@@ -27,6 +27,7 @@ namespace NUnit.Analyzers.Constants
         public const string FullNameOfTypeTestCaseAttribute = "NUnit.Framework.TestCaseAttribute";
 
         public const string NameOfTestCaseAttribute = "TestCaseAttribute";
+        public const string NameOfTestAttribute = "TestAttribute";
         public const string NameOfTestCaseAttributeExpectedResult = "ExpectedResult";
 
         public const string AssemblyQualifiedNameOfTypeAssert =
--- a/src/nunit.analyzers/Constants/TestCaseUsageAnalyzerConstants.cs
+++ b/src/nunit.analyzers/Constants/TestCaseUsageAnalyzerConstants.cs
@@ -5,6 +5,7 @@ namespace NUnit.Analyzers.Constants
         internal const string ExpectedResultTypeMismatchMessage = "The ExpectedResult value cannot be assigned to the return type, {0}";
         internal const string NotEnoughArgumentsMessage = "There are not enough arguments provided from the TestCaseAttribute for the method";
         internal const string SpecifiedExpectedResultForVoidMethodMessage = "Cannot specify ExpectedResult when the method returns void";
+        internal const string AsyncVoidMessage = "Async test method must have non-void return type";
         internal const string Title = "Find Incorrect TestCaseAttribute Usage";
         internal const string TooManyArgumentsMessage = "There are too many arguments provided from the TestCaseAttribute for the method";
         internal const string ParameterTypeMismatchMessage = "The value of the argument at position {0} cannot be assigned to the argument {1}";
--- a/src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs
+++ b/src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs
@@ -52,45 +52,53 @@ namespace NUnit.Analyzers.TestCaseUsage
                     var attributeNode = (AttributeSyntax)context.Node;
                     var attributeSymbol = context.SemanticModel.GetSymbolInfo(attributeNode).Symbol;
 
+                    var isTestCase = false;
                     if (testCaseType.ContainingAssembly.Identity == attributeSymbol?.ContainingAssembly.Identity &&
-                        NunitFrameworkConstants.NameOfTestCaseAttribute == attributeSymbol?.ContainingType.Name)
+                        (isTestCase = NunitFrameworkConstants.NameOfTestCaseAttribute == attributeSymbol?.ContainingType.Name)
+                        ||  NunitFrameworkConstants.NameOfTestAttribute == attributeSymbol?.ContainingType.Name)
                     {
                         context.CancellationToken.ThrowIfCancellationRequested();
 
                         var methodSymbol = context.SemanticModel.GetDeclaredSymbol(methodNode);
-                        var methodParameters = methodSymbol.GetParameterCounts();
-                        var methodRequiredParameters = methodParameters.Item1;
-                        var methodOptionalParameters = methodParameters.Item2;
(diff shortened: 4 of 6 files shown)


src/nunit.analyzers.tests/Targets/TestCaseUsage/TestCaseUsageAnalyzerTestsAnalyzeWhenReturnTypeIsAsyncVoid.cs
```diff
@@ -0,0 +1,14 @@
+using System.Threading.Tasks;
+using NUnit.Framework;
+
+namespace NUnit.Analyzers.Tests.Targets.TestCaseUsage
+{
+    class TestCaseUsageAnalyzerTestsAnalyzeWhenReturnTypeIsAsyncVoid
+    {
+        [Test]
```
304NotModified: note, this isn't a `TestCase`, but a `Test`

src/nunit.analyzers/Constants/TestCaseUsageAnalyzerConstants.cs
```diff
@@ -5,6 +5,7 @@ internal static class TestCaseUsageAnalyzerConstants
         internal const string ExpectedResultTypeMismatchMessage = "The ExpectedResult value cannot be assigned to the return type, {0}";
         internal const string NotEnoughArgumentsMessage = "There are not enough arguments provided from the TestCaseAttribute for the method";
         internal const string SpecifiedExpectedResultForVoidMethodMessage = "Cannot specify ExpectedResult when the method returns void";
+        internal const string AsyncVoidMessage = "Async test method must have non-void return type";
```
304NotModified: Not sure if this should be in **TestCase**UsageAnalyzerConstants

src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs
```diff
@@ -52,45 +52,53 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context)
                     var attributeNode = (AttributeSyntax)context.Node;
                     var attributeSymbol = context.SemanticModel.GetSymbolInfo(attributeNode).Symbol;
 
+                    var isTestCase = false;
                     if (testCaseType.ContainingAssembly.Identity == attributeSymbol?.ContainingAssembly.Identity &&
-                        NunitFrameworkConstants.NameOfTestCaseAttribute == attributeSymbol?.ContainingType.Name)
+                        (isTestCase = NunitFrameworkConstants.NameOfTestCaseAttribute == attributeSymbol?.ContainingType.Name)
+                        ||  NunitFrameworkConstants.NameOfTestAttribute == attributeSymbol?.ContainingType.Name)
```
304NotModified: maybe this could be simplified, now checking for TestCase + Test attributes

src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs
```diff
@@ -52,45 +52,53 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context)
...
+                                context.CancellationToken.ThrowIfCancellationRequested();
+                                TestCaseUsageAnalyzer.AnalyzePositionalArgumentsAndParameters(context,
+                                    attributePositionalArguments, methodSymbol.Parameters);
+                                context.CancellationToken.ThrowIfCancellationRequested();
+                                TestCaseUsageAnalyzer.AnalyzeNamedArguments(context,
+                                    attributeNamedArguments, methodSymbol);
+                            }
                         }
+
+                        AnalyzeReturnType(context, methodSymbol);
```
304NotModified: this will run with Test+TestCase

src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs
```diff
@@ -52,45 +52,53 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context)
...
+                            }
+                            else
+                            {
+                                context.CancellationToken.ThrowIfCancellationRequested();
+                                TestCaseUsageAnalyzer.AnalyzePositionalArgumentsAndParameters(context,
+                                    attributePositionalArguments, methodSymbol.Parameters);
+                                context.CancellationToken.ThrowIfCancellationRequested();
+                                TestCaseUsageAnalyzer.AnalyzeNamedArguments(context,
+                                    attributeNamedArguments, methodSymbol);
+                            }
```
304NotModified: nothing changed in this block above (lines 66-98)

304NotModified:
any opinion on this one?

jnm2:
I'm happy about this one. Would an error be more appropriate, since you know for sure that it will error at runtime?

src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs
```diff
@@ -128,6 +136,22 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context)
...
+            if (methodSymbol.IsAsync)
+            {
+
+                var methodReturnValueType = methodSymbol.ReturnType;
+
+                if (methodReturnValueType.SpecialType == SpecialType.System_Void)
+                {
+                    context.ReportDiagnostic(Diagnostic.Create(
+                        TestCaseUsageAnalyzer.CreateDescriptor(
+                            TestCaseUsageAnalyzerConstants.AsyncVoidMessage),
```
mikkelbu: Shouldn't this also be mentioned in `SupportedDiagnostics`?

src/nunit.analyzers.tests/Targets/TestCaseUsage/TestCaseUsageAnalyzerTestsAnalyzeWhenReturnTypeIsAsyncVoid.cs
mikkelbu: Perhaps we will - at some time - need a test for both cases. But for now I think it is fine

mikkelbu:
Hi @304NotModified
Thanks for the contribution, and again sorry about the delay, but I've only had time to very little the last copuple of weeks. Overall, I think it looks good, but I prefer that all the async functionality is placed in a separate analyzer (to keep the classes smaller). As I noted in the issue (#7) I've also started working on one of the other async tasks (#48), and I hoped to be done with it 10 days ago, but since then no much has happened 😞. 

But my "plan" is as follows:
* Create a new async analyzer for both `Test` and `TestCase` attributes (to begin with mostly a copy of the existing `TestCase` analyzer, then we can always improve on it - since some of the result type checking will be different from non-async methods to async methods)
* Add you changes on top of the new analyzer (I think we can use the test and the body of the analyzer - `AnalyzeReturnType` - more or less directly). 

I've also added some minor comments in the code, but please don't spend time on them before I've pushed the skeleton for #48

src/nunit.analyzers/Constants/NunitFrameworkConstants.cs
```diff
@@ -27,6 +27,7 @@ public static class NunitFrameworkConstants
         public const string FullNameOfTypeTestCaseAttribute = "NUnit.Framework.TestCaseAttribute";
 
         public const string NameOfTestCaseAttribute = "TestCaseAttribute";
+        public const string NameOfTestAttribute = "TestAttribute";
```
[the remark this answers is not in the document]
304NotModified: 👍 nice!

304NotModified:
> Thanks for the contribution, and again sorry about the delay, but I've only had time to very little the last copuple of weeks.

No problem! It's the nature of open source IMO 😄 

> Create a new async analyzer for both Test and TestCase attributes (to begin with mostly a copy of the existing TestCase analyzer, then we can always improve on it - since some of the result type checking will be different from non-async methods to async methods)

Could we create subclasses etc? e.g. a base class for the sharing functionality and the a subclass for Test attribute and and for testcase attribute?


> Add your changes on top of the new analyzer (I think we can use the test and the body of the analyzer - AnalyzeReturnType - more or less directly).

If I need to rebase stuff, no problem. Just let me know

mikkelbu:
@304NotModified Once more I've left this hanging for much too long, and I'm sorry about that.
I think that the easiest solution would be to extend the method `AnalyzeExpectedResult` to also handle that there is no `ExpectedResultAttribute`, but that the return type is not null.
I'm unsure how much can be reapplied from this PR, and I can fully understand if you think that your time is better spent somewhere else given the lack of feedback from my side, but I'll try to increase the activity on this project again.

304NotModified:
When I found time, I will create a new PR with the changes!

PR #71, opened 2018-08-19
Small cleanup,

304NotModified:
removed unused variables, unused init values, move to c#7

note: all done by Resharper

### DIFF
--- a/src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
+++ b/src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
@@ -46,8 +46,9 @@ namespace NUnit.Analyzers.ClassicModelAssertUsage
             {
                 var invocationNode = (InvocationExpressionSyntax)context.Node;
 
-                if (context.SemanticModel.GetSymbolInfo(
-                        invocationNode.Expression).Symbol is IMethodSymbol invocationSymbol && invocationSymbol.ContainingType.IsAssert())
+                var symbol = context.SemanticModel.GetSymbolInfo(invocationNode.Expression).Symbol;
+
+                if (symbol is IMethodSymbol invocationSymbol && invocationSymbol.ContainingType.IsAssert())
                 {
                     context.CancellationToken.ThrowIfCancellationRequested();
 
--- a/src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
+++ b/src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
@@ -45,10 +45,9 @@ namespace NUnit.Analyzers.ClassicModelAssertUsage
             if (methodNode != null && !methodNode.ContainsDiagnostics)
             {
                 var invocationNode = (InvocationExpressionSyntax)context.Node;
-                var invocationSymbol = context.SemanticModel.GetSymbolInfo(
-                    invocationNode.Expression).Symbol as IMethodSymbol;
 
-                if (invocationSymbol != null && invocationSymbol.ContainingType.IsAssert())
+                if (context.SemanticModel.GetSymbolInfo(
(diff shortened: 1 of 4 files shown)


src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs
```diff
@@ -18,18 +18,12 @@ public sealed class TestCaseUsageAnalyzer
...
-            {
-                return ImmutableArray.Create(
-                    TestCaseUsageAnalyzer.CreateDescriptor(TestCaseUsageAnalyzerConstants.ExpectedResultTypeMismatchMessage),
-                    TestCaseUsageAnalyzer.CreateDescriptor(TestCaseUsageAnalyzerConstants.NotEnoughArgumentsMessage),
-                    TestCaseUsageAnalyzer.CreateDescriptor(TestCaseUsageAnalyzerConstants.SpecifiedExpectedResultForVoidMethodMessage),
-                    TestCaseUsageAnalyzer.CreateDescriptor(TestCaseUsageAnalyzerConstants.ParameterTypeMismatchMessage),
-                    TestCaseUsageAnalyzer.CreateDescriptor(TestCaseUsageAnalyzerConstants.TooManyArgumentsMessage));
-            }
-        }
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
```
304NotModified: note: move to expression

src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs
```diff
@@ -135,11 +127,11 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context)
             var symbol = position >= methodParameter.Length ?
                 methodParameter[methodParameter.Length - 1] : methodParameter[position];
 
-            ITypeSymbol type = null;
+            ITypeSymbol type;
 
             if (symbol.IsParams)
             {
-                type = (symbol.Type as IArrayTypeSymbol).ElementType;
+                type = ((IArrayTypeSymbol)symbol.Type).ElementType;
```
304NotModified: better type cast error then nullref IMO

mikkelbu: Agree

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
```diff
@@ -45,10 +45,9 @@ private static void AnalyzeInvocation(SyntaxNodeAnalysisContext context)
             if (methodNode != null && !methodNode.ContainsDiagnostics)
             {
                 var invocationNode = (InvocationExpressionSyntax)context.Node;
-                var invocationSymbol = context.SemanticModel.GetSymbolInfo(
-                    invocationNode.Expression).Symbol as IMethodSymbol;
 
-                if (invocationSymbol != null && invocationSymbol.ContainingType.IsAssert())
+                if (context.SemanticModel.GetSymbolInfo(
+                        invocationNode.Expression).Symbol is IMethodSymbol invocationSymbol && invocationSymbol.ContainingType.IsAssert())
```
mikkelbu: I found the original code easier to read. I'm unsure if placing `if (context.SemanticModel.GetSymbolInfo(invocationNode.Expression).Symbol is IMethodSymbol invocationSymbol &&` on one line and the rest on the next line would make it easier. But still I find that several member accesses, two method calls, pattern matching, and a boolean operator to be too much "work" inside an `if`-statement.

304NotModified:
@mikkelbu  I've updated the PR, it's now more readable. Let me know what you think. 

If I need to revert the changes in ClassicModelAssertUsageAnalyzer, that's also OK with me.

mikkelbu:
@304NotModified I think it looks good. Sorry about the delay, but RL only allowed me little time to do nunit work the last couple of weeks, so I mostly spent time on smaller tasks.

Ps. Currently, the build is failing, but from the log it seems that appveyor have upgraded to msbuild 15.8 and that requires an `Installation` element within the `PackageManifest`. I've a PR ready to solve this.

Issue #70, opened 2018-08-19
Simplify test in this project


304NotModified:
Hi,

I was trying to add a new fix + test, but IMO adding a testcase in this project is pretty complicated

In my opinion:

1. The test classes names are really too long
![image](https://user-images.githubusercontent.com/5808377/44313662-a8999380-a40c-11e8-8b10-8650e24aa1f4.png)

1. The tests in `TestCaseUsageAnalyzerTests` are using to much `nameof` 
1. The `BasePath` in `TestCaseUsageAnalyzerTests` is a bit of magic
1. The testcases are duplicates and hard to read, even in the simple case:
![image](https://user-images.githubusercontent.com/5808377/44313675-f6ae9700-a40c-11e8-8cc8-0ad675ba313d.png)

1. The `RunAnalysisAsync` should not accept an `Action<ImmutableArray<Diagnostic>>`, but a expectedDiagnostics (of type `IEnumerable<Diagnostic>`)

mikkelbu:
Hi @304NotModified 
I wholeheartedly agree that the tests are too difficult to read, both due to the technical details, but also since the data and the assertions are separated into separate files.

I don't know the original reason for this, but I much prefer the solutions chosen in other roslyn-projects  where the data is just strings inside the test. Of cause one loses the intellisense, but this is still better than the current solution. I don't think there (currently) are any best practises for writing tests of analyzers/code fixes, but I know that @JohanLarsson has Gu.Roslyn.Asserts, and MS have several different variants (see e.g. https://github.com/dotnet/roslyn-analyzers/issues/1574#issuecomment-392066890).

I had planned to look into this after the first milestone, but perhaps we should take he battle now (so far I've lived with the pain in the few contributions I've made).

304NotModified:
Hi @mikkelbu 

Thanks for your reaction.

I do agree that strings are better, but I think there could be some improvement with the current approach (splitted testcases from tests). 

If you like, I will try to refactor a bit so it's more readable.

304NotModified:
made #73, please give your opinion :)

JohanLarsson:
@304NotModified using Gu.Roslyn.Asserts tests look like [this](https://github.com/GuOrg/Gu.Analyzers/blob/master/Gu.Analyzers.Test/GU0082IdenticalTestCaseTests/Diagnostics.cs#L13-L31):

```cs
        [Test]
        public void TestCaseAttributeAndParameter()
        {
            var testCode = @"
[...]

PR #73, opened 2018-08-21
[WIP] Refactor tests tryout

304NotModified:
This is a tryout to continue the discussion in #70, with some more concrete examples.

These are 2 directions we could go now, to improve the readability of the tests.

What I have done:

- moved the testcase to a folder and made the name shorter
- created a helper to get the filename from a generic (so it's easy to move from test to testcase and visa versa)
- created 2 options we could go to (or a mix of course)
   1. option 1 is a small change compared to the current testcode
   2. option 2 refactors also the helpers and so less duplicate testcode is needed. Also using C#7 tuples makes stuff easier to read

note: this assumes that the class name and the filename are the same. I think that's even best practice.

### DIFF
--- a/src/nunit.analyzers.tests/Extensions/AttributeArgumentSyntaxExtensionsTests.cs
+++ b/src/nunit.analyzers.tests/Extensions/AttributeArgumentSyntaxExtensionsTests.cs
@@ -1,10 +1,12 @@
 using System;
+using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using NUnit.Analyzers.Extensions;
+using NUnit.Analyzers.Tests.Targets.Extensions;
 using NUnit.Framework;
 
 namespace NUnit.Analyzers.Tests.Extensions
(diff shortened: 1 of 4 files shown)


src/nunit.analyzers.tests/Extensions/AttributeArgumentSyntaxExtensionsTests.cs
```diff
@@ -201,22 +203,43 @@ public async Task CanAssignToWhenParameterIsTimeSpanAndArgumentIsInvalidString()
                 rootAndModel.Item2);
         }
 
+        private async static Task<(AttributeArgumentSyntax Syntax, ITypeSymbol Symbol, SemanticModel Model)> GetAttributeSyntaxAsync<T>()
```
304NotModified: note: this is just a duplicate for demo purpose.

304NotModified:
Please share your thoughts on this refactor, thanks!

304NotModified:
> Please share your thoughts on this refactor, thanks!

Polite bump @mikkelbu. This would help me writing new code and so fixing #69

mikkelbu:
Hi @304NotModified 

I like the tuples, and I think that we should replace all the old unamed tuples with named tuples throughout the code base (to get rid of `Item1`...).

Looking at your examples I prefer option 2, but I think that the two examples are only working by mistake (I could be wrong as it is past midnight, and I'm _very_ tired). The reason for why I think that there is a problem is that the class `CanAssignToWhenArgumentIsImplicitlyTypedArrayAndNotAssignable` should not be compiled (or at least it used not to be compiled, the file was only copied to the output), so we cannot use the name as a type parameter in e.g. `await GetAttributeSyntaxAsync<CanAssignToWhenArgumentIsImplicitlyTypedArrayAndNotAssignable>()`.

The problem is that we have the following change in the csproj
```diff
-    <Compile Remove="Targets\Extensions\AttributeArgumentSyntaxExtensionsTestsCanAssignToWhenArgumentIsImplicitlyTypedArrayAndNotAssignable.cs" />
+    <Compile Remove="Targets\Extensions\CanAssignToWhenArgumentIsImplicitlyTypedArrayAndNotAssignable.cs" />
```
but the location of the file is now - in the folder `AttributeArgumentSyntaxExtensionsTests` - `Targets\Extensions\AttributeArgumentSyntaxExtensionsTests\CanAssignToWhenArgumentIsImplicitlyTypedArrayAndNotAssignable.cs`, so we have removed a non-existing file from compilation, and as a result `Targets\Extensions\AttributeArgumentSyntaxExtensionsTests\CanAssignToWhenArgumentIsImplicitlyTypedArrayAndNotAssignable.cs` will be compiled. I find it a bit strange that msbuild does not complain about `Compile Remove` of a non-existing file, but I don't know much about this.

I'll try to take a closer look tomorrow, but now I need to get some sleep.

304NotModified:
Superseded by #105  :)

mikkelbu:
Wow. I'm too slow 😄. I had started to write that I would close this due to #105, but got sidetracked by a late night snack, and you beat me to it.

PR #82, opened 2018-11-30
Add Analyzer for TestCaseSource(StringConstant)

### LINKED ISSUE #80: Suggestion: Analyzer Should Fix TestCaseSource(string) usage to use nameof()
Thank you for this tool.

We have several Test Cases that look similar to this:

```
[TestCaseSource( "_SetupIsNotEqualToNoEntryErrorBehaviorTestCases" )]
```

These were written before the introduction of [nameof()](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/nameof) and if they were to be re-written today would look like this:

```
[TestCaseSource( nameof(_SetupIsNotEqualToNoEntryErrorBehaviorTestCases) )]
```

This way Intellisense would operate correctly (F12/Renaming/Find Usages/etc).

I realize that you would be required to use a compiler/IDE that supports C# 6.0 (that usually means Visual Studio 2015) but it would encourage others to keep current.

Thank you for your consideration.

aolszowka:
Fixes #80

This at very least adds the Analyzer I'll see if I can get a Fixer Next.

### DIFF
--- a/src/nunit.analyzers.playground/Program.cs
+++ b/src/nunit.analyzers.playground/Program.cs
@@ -27,9 +27,14 @@ namespace NUnit.Analyzers.Playground
             Program.Foo(22);
         }
 
-        public static void Foo(short a) { }
+        public static void Foo(short a)
+        {
+        }
 
         [TestCase(3d, "a value", 123, ExpectedResult = 22)]
-        public int RunTest(double a, string b, int c) { return 22; }
+        public int RunTest(double a, string b, int c)
+        {
+            return 22;
+        }
     }
 }
--- a/src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
+++ b/src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
@@ -1,4 +1,3 @@
-using System.Collections.Generic;
 using System.Collections.Immutable;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
@@ -22,7 +21,6 @@ namespace NUnit.Analyzers.TestCaseSourceUsage
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
             CreateDescriptor(TestCaseSourceUsageConstants.ConsiderNameOfInsteadOfStringConstantMessage));
-
 
         public override void Initialize(AnalysisContext context)
         {
@@ -55,9 +53,9 @@ namespace NUnit.Analyzers.TestCaseSourceUsage
                 {
                     var stringConstant = ((LiteralExpressionSyntax)firstArgument).Token.ValueText;
                     context.ReportDiagnostic(Diagnostic.Create(
-                        CreateDescriptor( string.Format(TestCaseSourceUsageConstants.ConsiderNameOfInsteadOfStringConstantMessage, stringConstant)),
+                        CreateDescriptor(string.Format(TestCaseSourceUsageConstants.ConsiderNameOfInsteadOfStringConstantMessage, stringConstant)),
                         attributeNode.GetLocation(),
-                        ImmutableDictionary.Create<string,string>().Add("StringConstant", stringConstant)
+                        ImmutableDictionary.Create<string, string>().Add("StringConstant", stringConstant)
                         ));
                 }
             }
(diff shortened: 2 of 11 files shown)


src/nunit.analyzers.playground/Program.cs
```diff
@@ -27,9 +27,14 @@ static void Main(string[] args)
...
+        public static void Foo(short a)
+        {
+        }
 
         [TestCase(3d, "a value", 123, ExpectedResult = 22)]
-        public int RunTest(double a, string b, int c) { return 22; }
+        public int RunTest(double a, string b, int c)
+        {
+            return 22;
+        }
```
mikkelbu: This is fine. I think that we at some point will remove the "playground" from the project, but don't spend time reverting this 😄 .

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
```diff
@@ -0,0 +1,66 @@
+using System.Collections.Generic;
```
mikkelbu: I don't think that this is used.

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
```diff
@@ -0,0 +1,66 @@
...
+                TestCaseSourceUsageConstants.ConsiderNameOfInsteadOfStringConstantAnalyzerTitle,
+                message,
+                Categories.Usage,
+                DiagnosticSeverity.Warning,
+                true);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+            CreateDescriptor(TestCaseSourceUsageConstants.ConsiderNameOfInsteadOfStringConstantMessage));
+
+
```
mikkelbu: Please remove one of the empty lines

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
```diff
@@ -0,0 +1,66 @@
...
+            if (testCaseSourceType.ContainingAssembly.Identity == attributeSymbol?.ContainingAssembly.Identity &&
+                NunitFrameworkConstants.NameOfTestCaseSourceAttribute == attributeSymbol?.ContainingType.Name)
+            {
+                context.CancellationToken.ThrowIfCancellationRequested();
+
+                var arguments = attributeNode.ArgumentList.Arguments;
+
+                // If the First Argument is a String Constant we're in trouble
+                var firstArgument = arguments.FirstOrDefault().Expression;
```
mikkelbu: Actually, I thought that this would fail for `TestCaseSource()`, but for this `attributeSymbol` is null. I was worried about whether this would throw a NRE.

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
```diff
@@ -0,0 +1,66 @@
...
+                var firstArgumentIsStringConstant = firstArgument.Kind() == SyntaxKind.StringLiteralExpression;
+
+                if (firstArgumentIsStringConstant)
+                {
+                    var stringConstant = ((LiteralExpressionSyntax)firstArgument).Token.ValueText;
+                    context.ReportDiagnostic(Diagnostic.Create(
+                        CreateDescriptor( string.Format(TestCaseSourceUsageConstants.ConsiderNameOfInsteadOfStringConstantMessage, stringConstant)),
```
mikkelbu: Please remove space before `string.Format`.

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
```diff
@@ -0,0 +1,66 @@
...
+                        attributeNode.GetLocation(),
+                        ImmutableDictionary.Create<string,string>().Add("StringConstant", stringConstant)
```
mikkelbu: Please insert space between "," and "string".

mikkelbu:
I've also edited your original message such that it will auto close the issue when this PR is merged into master.

src/nunit.analyzers.playground/Program.cs
aolszowka: Sorry about this; could we maybe get an .editorconfig in here to fix this up? My workflow totally hosed up a bunch of these files (sorry!)

mikkelbu: Don't be sorry. It is more my fault than yours. There is a .editorconfig in the project, but I think that I left the playground project and the integrationtests project out in my cleanup as I planned to remove them at some time, as I don't think they provide much value (and actually the integrationtests project is not being build at the moment).

PR #103, opened 2019-01-04
Transfer ClassicModelAssertUsage tests to Gu.Roslyn.Asserts

mikkelbu:
One test `AnalyzeWhenDiagnosticIssuesExist` was removed as it did
not provide much value and was difficult to translate without using
`IgnoredErrorsAttribute` to allow compiler errors in the tested code.

Relates to #70


src/nunit.analyzers.tests/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzerTests.cs
```diff
@@ -53,148 +51,173 @@ public void VerifySupportedDiagnostics()
...
-                Array.Empty<string>());
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    public sealed class AnalyzeWhenThatIsUsed
+    {
+        public void Test()
+        {
+            Assert.That(true, Is.True);
+        }
+    }");
+            AnalyzerAssert.Valid<ClassicModelAssertUsageAnalyzer>(testCode);
```
JohanLarsson: Nitpick: I usually use the overload `AnalyzerAssert.Valid(analyzer, testCode);` here. Where is a field. Reason I prefer this overload is that the tests are more copy paste friendly that way. 😃

src/nunit.analyzers.tests/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzerTests.cs
```diff
@@ -53,148 +51,173 @@ public void VerifySupportedDiagnostics()
         }
 
         [Test]
-        public async Task AnalyzeWhenThatIsUsed()
+        public void AnalyzeWhenThatIsUsed()
         {
-            await TestHelpers.RunAnalysisAsync<ClassicModelAssertUsageAnalyzer>(
-                $"{ClassicModelAssertUsageAnalyzerTests.BasePath}{(nameof(this.AnalyzeWhenThatIsUsed))}.cs",
-                Array.Empty<string>());
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
```
JohanLarsson: Another nitpick: I prefer adding namespace and usings to the string literal over calling this helper. Reason for this I like tests self contained where possible.

src/nunit.analyzers.tests/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzerTests.cs
JohanLarsson: Copy paste friendly as you only need to update the field in one place. No biggie keeping it as-is is fine of course.

mikkelbu: Got it :+1:

Issue #113, opened 2019-03-02
Detect scenarios where expected and actual are switched


jnm2:
For example, if the `actual` argument to `Assert.AreEqual` is constant?

https://github.com/nunit/nunit/pull/3172#discussion_r261806723

mikkelbu:
This could be done together with - or as part of - #40.

jnm2:
Oh, #40 is what I was trying to find! This sounded familiar. Okay for me to close as covered by #40?

mikkelbu:
Closed as a duplicate of #40.

PR #114, opened 2019-03-03
Add analyzer for const actual value

### LINKED ISSUE #40: Warning if literal or const value is provided as actual value
In most cases (actually can't figure out when it should actually be otherwise) const value shouldn't be provided as actual value to assert.
However it can often happen when `actual` and `expected` parameter are mixed places. 
That mostly applies to classic assertion model, but might be helpful for constraint model as well.

Dreamescaper:
(fix #40 #113)

Warning if `actual` value provided to Assert method is literal or const.
Applies for all methods in Assert class, having 'actual' parameter.

### DIFF
--- a/src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
@@ -7,8 +7,7 @@ namespace NUnit.Analyzers.Tests.ActualConstValueUsage
 {
     public class ActualConstValueUsageAnalyzerTests
     {
-        private readonly DiagnosticAnalyzer analyzer = new ActualConstValueUsageAnalyzer();
-
+        private static readonly DiagnosticAnalyzer analyzer = new ActualConstValueUsageAnalyzer();
 
         [Test]
         public void AnalyzeWhenLiteralArgumentIsProvidedForAreEqual()
@@ -20,7 +19,7 @@ namespace NUnit.Analyzers.Tests.ActualConstValueUsage
                     Assert.AreEqual(expected, ↓1);
                 }");
 
-            AnalyzerAssert.Diagnostics(this.analyzer, testCode);
+            AnalyzerAssert.Diagnostics(analyzer, testCode);
         }
 
         [Test]
@@ -32,7 +31,7 @@ namespace NUnit.Analyzers.Tests.ActualConstValueUsage
                     Assert.That(↓true, Is.EqualTo(false));
                 }");
 
-            AnalyzerAssert.Diagnostics(this.analyzer, testCode);
+            AnalyzerAssert.Diagnostics(analyzer, testCode);
         }
 
         [Test]
@@ -46,7 +45,7 @@ namespace NUnit.Analyzers.Tests.ActualConstValueUsage
                     Assert.AreEqual(expected, ↓actual);
                 }");
 
-            AnalyzerAssert.Diagnostics(this.analyzer, testCode);
+            AnalyzerAssert.Diagnostics(analyzer, testCode);
         }
 
         [Test]
@@ -60,7 +59,7 @@ namespace NUnit.Analyzers.Tests.ActualConstValueUsage
                     Assert.That(↓actual, Is.EqualTo(expected));
                 }");
 
-            AnalyzerAssert.Diagnostics(this.analyzer, testCode);
+            AnalyzerAssert.Diagnostics(analyzer, testCode);
         }
 
         [Test]
@@ -78,7 +77,7 @@ namespace NUnit.Analyzers.Tests.ActualConstValueUsage
                 }
             }");
 
-            AnalyzerAssert.Diagnostics(this.analyzer, testCode);
+            AnalyzerAssert.Diagnostics(analyzer, testCode);
         }
 
         [Test]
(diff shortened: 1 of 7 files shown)


src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
```diff
@@ -0,0 +1,138 @@
+using Gu.Roslyn.Asserts;
+using Microsoft.CodeAnalysis.Diagnostics;
+using NUnit.Analyzers.ConstActualValueUsage;
+using NUnit.Framework;
+
+namespace NUnit.Analyzers.Tests.ActualConstValueUsage
+{
+    public class ActualConstValueUsageAnalyzerTests
+    {
+        private readonly DiagnosticAnalyzer analyzer = new ActualConstValueUsageAnalyzer();
```
JohanLarsson: Why not static? Also one new line after is enough.

JohanLarsson:
There is a c ode fix opportunity here but that can be added later of course.

src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
Dreamescaper: fixed

Dreamescaper:
Yes, decided to tackle that separately

src/nunit.analyzers/Constants/ConstActualValueUsageAnalyzerConstants.cs
```diff
@@ -0,0 +1,9 @@
+namespace NUnit.Analyzers.Constants
+{
+    internal static class ConstActualValueUsageAnalyzerConstants
+    {
+        public const string Title = "Actual value should not be constant";
+        public const string Message = "Actual value should not be constant - " +
+            "possible that actual and expected values are switched places";
```
mikkelbu: I'm not a native English speaker, but I think the sentence sounds a bit odd.
```suggestion
            "perhaps the actual and expected values have switched places";
```

src/nunit.analyzers/ConstActualValueUsageAnalyzer/ConstActualValueUsageAnalyzer.cs
```diff
@@ -0,0 +1,64 @@
+using System.Collections.Immutable;
```
mikkelbu: To be more consistent, we should remove `Analyzer` from the folder name.

src/nunit.analyzers/ConstActualValueUsageAnalyzer/ConstActualValueUsageAnalyzer.cs
```diff
@@ -0,0 +1,64 @@
...
+                {
+                    context.ReportDiagnostic(Diagnostic.Create(
+                        descriptor,
+                        argumentSyntax.GetLocation()));
+                }
+            }
+        }
+    }
+}
+
```
mikkelbu: When I look at the file in VS then there are to trailing empty lines. I think we should remove one of them.

src/nunit.analyzers/ConstActualValueUsageAnalyzer/ConstActualValueUsageAnalyzer.cs
```diff
@@ -0,0 +1,64 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using NUnit.Analyzers.Constants;
+using NUnit.Analyzers.Extensions;
+
+namespace NUnit.Analyzers.ConstActualValueUsage
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class ActualConstValueUsageAnalyzer : DiagnosticAnalyzer
```
mikkelbu: I think we should rename the class to follow the file name.
```suggestion
    public class ConstActualValueUsageAnalyzer.cs : DiagnosticAnalyzer
```

src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
```diff
@@ -0,0 +1,137 @@
+using Gu.Roslyn.Asserts;
+using Microsoft.CodeAnalysis.Diagnostics;
+using NUnit.Analyzers.ConstActualValueUsage;
+using NUnit.Framework;
+
+namespace NUnit.Analyzers.Tests.ActualConstValueUsage
+{
+    public class ActualConstValueUsageAnalyzerTests
```
mikkelbu: I think we should rename the class to follow the file name.
```suggestion
    public class ConstActualValueUsageAnalyzerTests
```

src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
```diff
@@ -0,0 +1,137 @@
...
+                public void Test()
+                {
+                    string actual = ""act"";
+                    Assert.That(actual, Is.EqualTo(expected));
+                }
+            }");
+
+            AnalyzerAssert.Valid(analyzer, testCode);
+        }
+    }
```
mikkelbu: I think we should also add tests that cover when the arguments are named.

src/nunit.analyzers/Constants/ConstActualValueUsageAnalyzerConstants.cs
Dreamescaper: fixed

Dreamescaper:
@mikkelbu 
Thanks for review! Updated PR with suggested changes.

PR #117, opened 2019-03-18
Code fix for ConstActualValueUsageAnalyzer

Dreamescaper:
Simply switch 'actual' and 'expexted' arguments places. 

Same as for analyzer itself, I wanted to make this code fix pattern-based, meaning that code fix is applicable for any Assert method that has 'actual' and 'expected' parameters (or 'actual' and 'constraint' method argument with 'expected' parameter, e.g. `Assert.That(act, Is.EqualTo(exp))`).

However, there are two things I'm not sure about, so would like to discuss:

1) Naming of Constraint parameter is not unified across Assert methods - it might be 'expression' or 'expr' or 'constraint' (maybe should be unified in NUnit itself?). The only possible option in current context, however, is 'expression', so that's what I used.

2) Not sure if it should be applicable for non-symmetric asserts, as it will change functionality. E.g. `Assert.That("abc", Does.StartsWith(someVar))`. Should we suggest code fix here? I'm a bit hesitant, as swapping arguments will lead to functional changes. Should we do that?


src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
```diff
@@ -125,13 +125,13 @@ private static string GetQualifiedTypeName(ITypeSymbol targetType)
         {
             // Note that this does not take into account generics,
             // so if that's ever added to attributes this will have to change.
-            var namespaces = new List<string>();
+            var namespaces = new Stack<string>();
```
Dreamescaper: Not related to current PR, just thing that I mentioned. 
Namespaces go in reverse order if getting them using ContainingNamespace (e.g. "A.B.C.SomeClass" -> "C.B.A"), therefore we need to reverse them again.

mikkelbu:
Thanks for the contribution @Dreamescaper. First I'll try to answer your questions and then I will take a closer look at the code (so far I've only tested it, to see how it behaves).

> However, there are two things I'm not sure about, so would like to discuss:
>
> Naming of Constraint parameter is not unified across Assert methods - it might be 'expression' or 'expr' or 'constraint' (maybe should be unified in NUnit itself?). The only possible option in current context, however, is 'expression', so that's what I used.

I can live with that 😄. We can always improve on the code (in this repository or in the NUnit repository) later on.

> Not sure if it should be applicable for non-symmetric asserts, as it will change functionality. E.g. Assert.That("abc", Does.StartsWith(someVar)). Should we suggest code fix here? I'm a bit hesitant, as swapping arguments will lead to functional changes. Should we do that?

I very much like that you have chosen the pattern-based approach as it covers a lot of different cases without too much complexity in the analyzer or in the fixer. That being said I think that for the fixer we should restrict it a little so that it only matches a constraint that starts with the class `Is`, e.g. `Assert.That(____, Is._____(_____))` in "option 2" in the code. Then we can always generalise to cover more cases - or restrict it and remove cases - later on. However, we will still have non-symmetric situations using `Is`, like `Is.GreaterThan(expected)`.

It is also not only non-symmetric cases that are problematic, for statements such as `Assert.That("1231", Is.EqualTo(expected1).And.EqualTo(expected2));` the fixer will exchange `"1231"` and `expected2`.

src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageCodeFixTests.cs
```diff
@@ -0,0 +1,81 @@
...
+                {
+                    int expected = 5;
+                    Assert.AreEqual(1, expected);
+                }");
+
+            AnalyzerAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode,
+                fixTitle: CodeFixConstants.SwapArgumentsDescription);
+        }
+
+
```
mikkelbu: Please remove one of the empty lines.

src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
mikkelbu: I think you are right. As far as I can remember we only hit this code for classes in the `System` namespace, so have not noticed discovered the problem.

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageCodeFix.cs
```diff
@@ -0,0 +1,102 @@
...
+                var constraintExpression = invocationSyntax.ArgumentList.Arguments[1].Expression as InvocationExpressionSyntax;
+
+                if (constraintExpression == null)
+                    return false;
+
+                var constraintMethod = semanticModel.GetSymbolInfo(constraintExpression).Symbol as IMethodSymbol;
+
+                if (constraintMethod == null)
+                    return false;
+
```
mikkelbu: I think that we should return here if `constraintExpression` does not match `Is.____(expected)` for some method name `____`.

Dreamescaper: Do you think we shouldn't support cases like `Assert.That(smth, Is.Not.EqualTo(123))`?

mikkelbu: I think it is okay also to support this if it does not make the code too complex. As far as I can tell `Is` has the following methods that are symmetric `EqualTo`, `SameAs`, and `SamePath`, and each of these can be "prefixed" by zero or more `.Not` (but I think it is ok that the code only handles zero or one).

Dreamescaper:
Updated PR.
Now code fix is limited only for cases:
Assert.AreEqual, Assert.AreNotEqual, Assert.AreSame, Assert.AreNotSame,
Is.EqualTo, Is.Not.EqualTo, Is.SameAs, Is.Not.SameAs, Is.SamePath, Is.Not.SamePath

Updated and added corresponding tests

PR #133, opened 2019-05-18
Warning if actual value is same as expected

### LINKED ISSUE #129: Warning if actual value is same as expected
Fixed very useful test today, something like that:
```
// Arrange
var types = /* some code */;

// Act
var actualTypes = /* some code */

// Assert
Assert.That(types, Is.EqualTo(types));
```

Probably we should show warning in such cases.

I would suggest to apply analyzer for any constraint with 'expected' parameter - don't think we have any tricky case where it might make sense.

Not sure about arguments though. Should we limit verification by variables only? Or compare the whole expression?

_Technically_ something like that might be correct:
```
int i = 0;
int getNext() => i++;
Assert.That(getNext(), Is.Not.EqualTo(getNext()));
```
Very rare case, however. 
Maybe make sense to make it generic - for the whole expression, and to adjust depending on feedback, if needed.

Dreamescaper:
fix #129


Dreamescaper:
Merged suggested chagnes, added tests for multiple combined constraints

PR #136, opened 2019-05-25
Refactor assertion analyzers

### LINKED ISSUE #134: Constraint analyzers have a lot of repeating code
Constraints analyzers (classic assert analyzers, IgnoreCaseUsage, ConstActaulValue etc) have a lot of similar code (check if assert, get actual/expected arguments etc).

Probably makes sense to
a) Create some `AssertExpressionHelper` with commonly used methods
b) Create abstract class `AssertionAnalyzer` with existing check for assert.

Dreamescaper:
Fix #134

Added abstract class `BaseAssertionAnalyzer`, which has abstract method `AnalyzeAssertInvocation(SyntaxNodeAnalysisContext context, InvocationExpressionSyntax assertExpression, IMethodSymbol methodSymbol);` to avoid checks that expression is actually assert in each analyzer.

Added helper class `AssertExpressionHelper` with methods to get actual and constraint arguments, get expected values from constraint expression, split constraint by operators.

### DIFF
--- a/src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
@@ -29,7 +29,7 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
                 public void Test()
                 {
                     int expected = 5;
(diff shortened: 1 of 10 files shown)


mikkelbu:
Thanks for the work @Dreamescaper, I'll take a look at it in the end of the week.

PR #137, opened 2019-05-25
Partition the analyzers into categories

### LINKED ISSUE #84: Partition the analyzers into categories and create identifier ranges for each category
Currently, the analyzer identifiers are just consecutively numbered with no meaning behind a given number. Instead we should create ranges for each area of analyzers and reassign the identifiers - similar to what most other analyzer projects do.

**Tasks:**
* Partition the existing analyzers into groups (and perhaps also take into account the existing open issues)
* Create a range of identifiers for each group and reassign identifiers to the analyzers

Dreamescaper:
Fix #84

Changed Categories to `Structure` and `Assertion`.

Changed analyzer identifiers according to rule: `NUnitxxxx`
Structure - Range from 1000 to 1999. 
Assertion - Range from 2000 to 2999.


mikkelbu:
Thanks for the work @Dreamescaper, I'll take a look at it in the end of the week.

Issue #96, opened 2018-12-21
Document implemented analyzers and fixers


mikkelbu:
We can probably get inspiration from the other projects:
* https://dotnetanalyzers.github.io/StyleCopAnalyzers/
* https://xunit.github.io/xunit.analyzers/rules/
* https://github.com/DotNetAnalyzers/ReflectionAnalyzers
* https://github.com/GuOrg/Gu.Analyzers

To begin with we should just keep it simple. Then later we can improve on UI and automation

JohanLarsson:
I use a nasty hack for docs: https://github.com/DotNetAnalyzers/AspNetCoreAnalyzers/blob/master/AspNetCoreAnalyzers.Tests/Documentation/Tests.cs

A couple of tests that check that docs exists and generate stub docs if missing. Not pretty but have found it good enough.

mikkelbu:
Thanks for the link, at least it could be useful in the beginning.

JohanLarsson:
I can PR it if you want it like that.

mikkelbu:
Hi @JohanLarsson 
Do you have time for the PR? Or if you don't have time, can you send me the tests or upload them somewhere (this would also help with subsequent updates to the analyzers).

JohanLarsson:
I'll try to find time this week.

mikkelbu:
Closed by #139

PR #139, opened 2019-06-02
Tests checking documentation.

JohanLarsson:
If a new analyzer is added with no documentation the test MissingDocs fails and creates a stub doc Id.md.generated.
Remove the .generated suffix and add it to the solution folder /docs
Not sure if the extra step with .generated suffix solves a problem but that is how I originally wrote it for some reason.

Currently the analyzers do not have Description, the stub docs have a `## Description` with an empty value and the test for description is set to [Explicit]. If no description is by design we should remove the test and the noise from the stub docs.

#96

We create help links in the analyzers so that clicking on the id in the VS error list navigates to the documentation. I can PR that later.

The code in this PR is not awesome, @jnm2 said it was fascinating, he is a polite man :)

### DIFF
--- a/documentation/NUnit1001.md
+++ b/documentation/NUnit1001.md
@@ -1,0 +1,51 @@
+# NUnit1001
+## Find Incorrect TestCaseAttribute Usage
+
+| Topic    | Value
+| :--      | :--
+| Id       | NUnit1001
+| Severity | Error
+| Enabled  | True
+| Category | Structure
+| Code     | [TestCaseUsageAnalyzer](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs)
+
+
+## Description
+
+
+
+## Motivation
+
+ADD MOTIVATION HERE
+
+## How to fix violations
+
+ADD HOW TO FIX VIOLATIONS HERE
+
+<!-- start generated config severity -->
+## Configure severity
+
+### Via ruleset file.
+
+Configure the severity per project, for more info see [MSDN](https://msdn.microsoft.com/en-us/library/dd264949.aspx).
+
+### Via #pragma directive.
+```C#
+#pragma warning disable NUnit1001 // Find Incorrect TestCaseAttribute Usage
+Code violating the rule here
+#pragma warning restore NUnit1001 // Find Incorrect TestCaseAttribute Usage
+```
+
+Or put this at the top of the file to disable all instances.
+```C#
+#pragma warning disable NUnit1001 // Find Incorrect TestCaseAttribute Usage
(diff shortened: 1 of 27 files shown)


documentation/NUnit1001.md
```diff
@@ -0,0 +1,51 @@
...
+| Topic    | Value
+| :--      | :--
+| Id       | NUnit1001
+| Severity | Error
+| Enabled  | True
+| Category | Structure
+| Code     | [TestCaseUsageAnalyzer](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs)
+
+
+## Description
```
JohanLarsson: No description provided in the descriptors. I included this header with empty value in the generated stubs.

src/nunit.analyzers.tests/DocumentationTests.cs
```diff
@@ -0,0 +1,335 @@
...
+        {
+            var expected = $"## {descriptorInfo.Descriptor.Title}";
+            var actual = descriptorInfo.DocumentationFile.AllLines
+                                       .Skip(1)
+                                       .First()
+                                       .Replace("`", string.Empty);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Explicit("No descriptions. Remove this test and update doc stubs if we don't want descriptions.")]
```
JohanLarsson: This test should be removed or `[Explicit]` should be removed depending on if we want description.

documentation/NUnit1001.md
mikkelbu: I think it is fine. We should have descriptions in this project

src/nunit.analyzers.tests/DocumentationTests.cs
mikkelbu: We should add descriptions and then remove `[Explicit]`.

mikkelbu:
Thank you very much for the contribution @JohanLarsson. This will save us a lot of work. Would you mind if I add a commit to this PR that reformats the code in `DocumentationTests.cs`, so that it follows the rest of the project (I think that this would be much faster than writing a lot of nit-pick comments about syntax). Alternative I can also merge this PR as it is, and then subsequently change the code a little.

Ps. I think you are being too hard on the code :). I was fearing something much more "interesting" from your descriptions :D.

JohanLarsson:
Of course I don't mind, I copy pasted, forgot about project conventions, sorry about that.

jnm2:
@JohanLarsson Oh dear, I meant "fascinating" more literally... as in, I'm interested in  this kind of workflow, the idea of autogenerating and maintaining docs via the test suite. I'm so sorry for giving the wrong impression!

PR #144, opened 2019-06-22
feat: Add motivation and examples for classical assertions

mikkelbu:
And correct title, message, and description of NUnit2005 and NUnit2006.

### DIFF
--- a/documentation/NUnit2001.md
+++ b/documentation/NUnit2001.md
@@ -16,11 +16,29 @@ Consider using the constraint model, Assert.That(expr, Is.False), instead of the
 
 ## Motivation
 
-ADD MOTIVATION HERE
+The classic Assert model contains less flexibility and reads more clumsy than the constraint model,
+so this analyzer marks usages of `Assert.False` from the classic Assert model.
+
+```csharp
+[Test]
+public void Test()
+{
+    Assert.False(expression);
+}
+```
 
 ## How to fix violations
 
-ADD HOW TO FIX VIOLATIONS HERE
+The analyzer comes with a code fix that will replace `Assert.False(expression)` with
+`Assert.That(expression, Is.False)`. So the code block above will be changed into.
+
+```csharp
+[Test]
+public void Test()
+{
+    Assert.That(expression, Is.False);
+}
+```
 
 <!-- start generated config severity -->
 ## Configure severity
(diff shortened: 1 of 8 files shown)


documentation/NUnit2001.md
```diff
@@ -16,11 +16,29 @@ Consider using the constraint model, Assert.That(expr, Is.False), instead of the
 
 ## Motivation
 
-ADD MOTIVATION HERE
+The classic Assert model contains less flexibility and reads more clumsy than the constraint model,
```
JohanLarsson: I would not agree that `Assert.False(expression)` is clumsy compared to `Assert.That(expression, Is.False)`, the opposite to me. Maybe change to something about being consistent with which API to write in a test suite. Also this rule is default disabled right?

jnm2: I agree, 'clumsy' isn't maybe the word. We do consider them legacy APIs though.

`Assert.False` actually is a call to `Assert.That(actual, Is.False)` and I personally don't mind it. I do personally dislike `AreEqual` and on top of that I continually see people mix up the expected and actual parameters.

JohanLarsson: Are there plans on removing them in a future version? Probably good to check usage statistics before doing it.

jnm2: I don't believe there are any plans to remove them.

mikkelbu: > I would not agree that Assert.False(expression) is clumsy compared to Assert.That(expression, Is.False), the opposite to me. Maybe change to something about being consistent with which API to use in a test suite?

"Clumsy" was my poor attempt at writing that often `Assert.True(expression)` ends up reading like Yoda expressions. I'll try to see if I can come up with a better description - and I'm open for suggestions.

> Also this rule is default disabled right? I've been thinking about adding rules + fixes for refactoring in the other direction. Been doing it manually a couple of times.

It is a warning, but it is enabled as default (currently none of the analyzers are disabled as default). Perhaps we should have analyzers, code fixes in both directions and then people can select the one that matches their setup (but I'm unsure if people would ever turn on an analyzer, or it just will be wasted work).

Regarding removing the classical model, then I think that originally (around 3.0 or before) there were plans to remove the classical model. This is no longer the case, but most work is done on the constraint model now.

JohanLarsson: Agreed default disabled making it opt-in has somewhat poor discoverability.
We can maybe use the github API to figure out if classical model is more widely used? It is rare to find usage of `Assert.that` in my experience.
Another option is to make it default disabled to start with and change it later if we find solid data.

JohanLarsson: Maybe a new issue is better as this is already merged. I'll create one.

PR #146, opened 2019-07-01
Add analyzer and code fix for Is.EqualTo usage

Dreamescaper:
PR adds analyzer and code fix for cases, when comparison via `==` or `!=` operators or via `.Equals` method is used instead of `Is.EqualTo` or `Is.Not.EqualTo` constraints.


Dreamescaper:
@mikkelbu 
Hi! Would you be able to take a look at this one?

mikkelbu:
Hi @Dreamescaper 
Thanks for the ping. I've been on vacation for 3 weeks - and have been busy the weeks before and after the vacation, so I did not have much time to do anything except minimal work.

Your contribution is as usual much appreciated and I hope to take a look at it tomorrow or during the weekend.

Dreamescaper:
@mikkelbu 
Thanks for update, no rush.
Hope you had great vacation!

src/nunit.analyzers/ConstraintUsage/EqualToConstraintUsageAnalyzer.cs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                if (!(constraint is MemberAccessExpressionSyntax memberAccess
+                       && memberAccess.Expression is IdentifierNameSyntax identifierSyntax
+                       && identifierSyntax.Identifier.Text == NunitFrameworkConstants.NameOfIs
+                       && (memberAccess.Name.Identifier.Text == NunitFrameworkConstants.NameOfIsTrue
+                           || memberAccess.Name.Identifier.Text == NunitFrameworkConstants.NameOfIsFalse)))
+                {
+                    return;
+                }
+                else if (memberAccess.Name.Identifier.Text == NunitFrameworkConstants.NameOfIsFalse)
```
mikkelbu: I think that `memberAccess` can be null.

src/nunit.analyzers/ConstraintUsage/EqualToConstraintUsageAnalyzer.cs
```diff
@@ -0,0 +1,140 @@
...
+                    && (IsStaticObjectEquals(unarySyntax.Operand, context)
+                        || IsInstanceObjectEquals(unarySyntax.Operand, context)))
+            {
+                shouldReport = true;
+                negated = !negated;
+            }
+
+            if (shouldReport && !negated)
+            {
+                context.ReportDiagnostic(Diagnostic.Create(
```
mikkelbu: I think this can be written on one line (and similarly below)

src/nunit.analyzers/ConstraintUsage/EqualToConstraintUsageAnalyzer.cs
```diff
@@ -0,0 +1,140 @@
...
+
+            var shouldReport = false;
+
+            if (actual.IsKind(SyntaxKind.EqualsExpression)
+                || IsStaticObjectEquals(actual, context)
+                || IsInstanceObjectEquals(actual, context))
+            {
+                shouldReport = true;
+            }
+            else if (actual.IsKind(SyntaxKind.NotEqualsExpression)
```
mikkelbu: I think extracting some of the expressions into variables and/or perhaps adding "()" would help a lot on the readability.

src/nunit.analyzers/ConstraintUsage/EqualToConstraintUsageAnalyzer.cs
Dreamescaper: Not sure. Why? It is checked in 'if' above

Dreamescaper:
@mikkelbu 
Thanks! Hope you had enough time to sleep :)

Updated PR based on your review.

I also not sure about two separate descriptors anymore. It would require all condition analyzers (i.e. to replace bool assert with constraints, I think there might be a lot of them) to do same, which would be a bit of a pain.
I would rather use single descriptor with parameterized message. Not sure if it would work well with our documentation tests thought.

Wdyt?

mikkelbu:
@Dreamescaper I had time to sleep, but not enough 😄. 

I have thought a bit about two separate descriptors vs. a single descriptor with parameterized message. Do I understand the problem correctly that you want to avoid the duplication in defining two almost identical descriptors. As far as I can tell we will still need to have the same logic and tests as we have now to determine which message to show, but is there more (duplication) that I overlook?

The problem is not only the documentation and the documentation tests, but also VS as it will only then show one entry as there is only one descriptor, so it will not be possible to distinguish the two "cases", e.g. to enable one and disable another, and the title must be the same (and more general).

Given the above I think I prefer to have two separate descriptors, but I feel that I might overlook something.

Dreamescaper:
> Do I understand the problem correctly that you want to avoid the duplication in defining two almost identical descriptors.

Mostly yes. Also, similar analyzers, that could be added in future (e.g. for string methods Contains, StartsWith, EndsWith etc), would have to do same.


> so it will not be possible to distinguish the two "cases", e.g. to enable one and disable another

I'd say it's actually better. I highly doubt that someone will disable Is.EqualTo analyzer, but leave Is.Not.EqualTo enabled intentionally.

Anyway, I'm fine with both ways.

mikkelbu:
Then let us go with the one analyzer approach. With respect to the parameterized message, then we don't show that in the documentation, so I don't think that would be a problem.

Dreamescaper:
> Then let us go with the one analyzer approach.

Ok, will update PR later then.

> With respect to the parameterized message, then we don't show that in the documentation, so I don't think that would be a problem.

Not sure I understand this part. Could you clarify? What we don't show in documentation?

mikkelbu:
For instance the message(format) of `NUnit1003` is `"There are not enough arguments provided by the TestCaseAttribute. Expected '{0}', but got '{1}'."`, but this is not shown in https://github.com/nunit/nunit.analyzers/blob/master/documentation/NUnit1003.md. We only show the title and the description in the documentation as it is now (and these are fixed for the existing analyzers).

But perhaps we should also show the message(format) at some point.

Dreamescaper:
Updated PR to use single parameterized descriptor. 

Previously I checked exact descriptorId in CodeFixProvider to understand whether expresion is negated, now I put Properties dictionary with suggested constraint to Diagnostic for this purpose.

mikkelbu:
Thanks @Dreamescaper. It looks good I'll be merging the contribution now. Sorry for the long delay, but I've not had enough time to keep up with all the NUnit work.

PR #151, opened 2019-08-31
String boolean methods analyzer

### LINKED ISSUE #150: Analyzers and refactorings to use string constraints instead of methods
E.g. instead of 
`Assert.That("123".Contains("12"))`
should be
`Assert.That("123", Does.Contain("12"))`

Other cases:
`StartsWith` -> `Does.StartWith`
`EndsWith` -> `Does.EndWith`

Dreamescaper:
fixes #150 

Extracted common code from `EqualConstraintUsageAnalyzer` and `EqualConstraintUsageCodeFix` to `BaseConditionConstraintAnalyzer` and `BaseConditionConstraintCodeFix` respectively.

Supported constraints:
`Does.Contain`, `Does.StartWith`, `Does.EndWith`, and corresponding `Does.Not.*` constraints.

### DIFF
--- a/src/nunit.analyzers.tests/ConstraintsUsage/StringConstraintUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/ConstraintsUsage/StringConstraintUsageAnalyzerTests.cs
@@ -10,10 +10,20 @@ namespace NUnit.Analyzers.Tests.ConstraintsUsage
     {
         private static readonly DiagnosticAnalyzer analyzer = new StringConstraintUsageAnalyzer();
 
-        [TestCase(nameof(string.Contains), AnalyzerIdentifiers.StringContainsConstraintUsage, "Does.Contain")]
-        [TestCase(nameof(string.StartsWith), AnalyzerIdentifiers.StringStartsWithConstraintUsage, "Does.StartWith")]
-        [TestCase(nameof(string.EndsWith), AnalyzerIdentifiers.StringEndsWithConstraintUsage, "Does.EndWith")]
-        public void AnalyzeStringBooleanMethodPositiveAssert(string method, string analyzerId, string suggestedConstraint)
+        private static readonly object[] PositiveAssertData = new[] {
+            new [] { nameof(string.Contains), AnalyzerIdentifiers.StringContainsConstraintUsage, "Does.Contain" },
+            new [] { nameof(string.StartsWith), AnalyzerIdentifiers.StringStartsWithConstraintUsage, "Does.StartWith" },
+            new [] { nameof(string.EndsWith), AnalyzerIdentifiers.StringEndsWithConstraintUsage, "Does.EndWith" }
+        };
+
+        private static readonly object[] NegativeAssertData = new[] {
+            new [] { nameof(string.Contains), AnalyzerIdentifiers.StringContainsConstraintUsage, "Does.Not.Contain" },
+            new [] { nameof(string.StartsWith), AnalyzerIdentifiers.StringStartsWithConstraintUsage, "Does.Not.StartWith" },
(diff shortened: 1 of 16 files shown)


src/nunit.analyzers.tests/ConstraintsUsage/StringConstraintUsageCodeFixTests.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+        private static readonly DiagnosticAnalyzer analyzer = new StringConstraintUsageAnalyzer();
+        private static readonly CodeFixProvider fix = new StringConstraintUsageCodeFix();
+
+        [TestCase(nameof(string.Contains), AnalyzerIdentifiers.StringContainsConstraintUsage, "Does.Contain")]
+        [TestCase(nameof(string.StartsWith), AnalyzerIdentifiers.StringStartsWithConstraintUsage, "Does.StartWith")]
+        [TestCase(nameof(string.EndsWith), AnalyzerIdentifiers.StringEndsWithConstraintUsage, "Does.EndWith")]
+        public void AnalyzeStringBooleanMethodPositiveAssert(string method, string analyzerId, string suggestedConstraint)
+        {
+            var code = TestUtility.WrapInTestMethod($@"Assert.True(""abc"".{method}(""ab""));");
```
mikkelbu:
```suggestion
            var code = TestUtility.WrapInTestMethod($@"Assert.True(↓""abc"".{method}(""ab""));");
```

src/nunit.analyzers.tests/ConstraintsUsage/StringConstraintUsageCodeFixTests.cs
```diff
@@ -0,0 +1,39 @@
...
+
+            var fixedCode = TestUtility.WrapInTestMethod($@"Assert.That(""abc"", {suggestedConstraint}(""ab""));");
+
+            AnalyzerAssert.CodeFix(analyzer, fix, ExpectedDiagnostic.Create(analyzerId), code, fixedCode);
+        }
+
```
mikkelbu: Perhaps we should also have tests that cover the codefix when the code before the fix is using `Assert.That(...)` instead of `Assert.True(...)`, e.g. `Assert.That("...".StartsWith("..."))`.

src/nunit.analyzers.tests/ConstraintsUsage/StringConstraintUsageAnalyzerTests.cs
```diff
@@ -0,0 +1,54 @@
...
+        [TestCase(nameof(string.EndsWith), AnalyzerIdentifiers.StringEndsWithConstraintUsage, "Does.EndWith")]
+        public void AnalyzeStringBooleanMethodPositiveAssert(string method, string analyzerId, string suggestedConstraint)
+        {
+            var testCode = TestUtility.WrapInTestMethod($@"Assert.True(↓""abc"".{method}(""ab""));");
+
+            var diagnostic = ExpectedDiagnostic.Create(analyzerId,
+                string.Format(StringConstraintUsageConstants.Message, suggestedConstraint));
+            AnalyzerAssert.Diagnostics(analyzer, diagnostic, testCode);
+        }
+
```
mikkelbu: Same comment as for the fixer. Do we need a test when the context is `Assert.That(...)` and not `Assert.True`?

Dreamescaper: Wouldn't harm. Will add a bit later.

Dreamescaper:
Added more tests. I wanted to make them parametrized to decrease copy-paste, but haven't came up with anything readable.

PR #156, opened 2019-10-11
Add DateTimeOffset parameter check and some tests

### LINKED ISSUE #148: Analyzer update needed for string->datetime offset conversion in testcaseattribute
NUnit Version: 3.13

Support was added for string->datetime offset conversion in https://github.com/nunit/nunit/issues/3053

But I've noticed I'm getting an analyzer info message about an invalid conversion when using it. We should update the analyzers to be aware of the new type conversion.

Assuming we want to go ahead with this change, I'd actually be happy to try to implement too. Would be a much-wanted introduction for me to Roslyn analyzers.

stevenaw:
Fixes #148 

This PR adds support for detecting valid string -> DateTimeOffset conversions in TestCase.

### DIFF
--- a/src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
@@ -114,12 +114,12 @@ namespace NUnit.Analyzers.Tests.TestCaseUsage
         [TestCaseSource(nameof(SpecialConversions))]
         public void AnalyzeWhenArgumentIsSpecialConversion(string value, Type targetType)
         {
-            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing($@"
     public sealed class AnalyzeWhenArgumentIsSpecialConversion
-    {
-        [TestCase(""" + value + @""")]
-        public void Test(" + targetType.Name + @" a) { }
-    }");
+    {{
+        [TestCase(""{value}"")]
+        public void Test({targetType.Name} a) {{ }}
+    }}");
             AnalyzerAssert.Valid<TestCaseUsageAnalyzer>(testCode);
         }
 
--- a/src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
@@ -15,6 +15,16 @@ namespace NUnit.Analyzers.Tests.TestCaseUsage
     public sealed class TestCaseUsageAnalyzerTests
     {
         private DiagnosticAnalyzer analyzer = new TestCaseUsageAnalyzer();
+        
+        private static IEnumerable<TestCaseData> SpecialConversions
+        {
+            get
+            {
+                yield return new TestCaseData("2019-10-10", typeof(DateTime));
+                yield return new TestCaseData("23:59:59", typeof(TimeSpan));
+                yield return new TestCaseData("2019-10-10", typeof(DateTimeOffset));
+            }
+        }
 
         [Test]
         public void VerifySupportedDiagnostics()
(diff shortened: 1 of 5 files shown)


src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
```diff
@@ -64,6 +64,11 @@ internal static bool CanAssignTo(this AttributeArgumentSyntax @this, ITypeSymbol
                     {
                         canConvert = TimeSpan.TryParse(argumentValue as string, out _);
                     }
+                    else if (argumentType.SpecialType == SpecialType.System_String &&
+                        model.Compilation.GetTypeByMetadataName(typeof(DateTimeOffset).FullName).IsAssignableFrom(targetType))
+                    {
+                        canConvert = DateTimeOffset.TryParse(argumentValue as string, out _);
```
stevenaw: Note to reviewer: I used `TryParse` here, but I think the underlying NUnit implementation uses `TypeConverter.CanConvertFrom`. Please let me know if you'd prefer I update it.

Reference code: https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Internal/ParamAttributeTypeConversions.cs#L138-L143

src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
```diff
@@ -51,6 +51,29 @@ public void VerifySupportedDiagnostics()
                 $"{TestMethodUsageAnalyzerConstants.SpecifiedExpectedResultForVoidMethodMessage} is missing.");
         }
 
+
+        private static IEnumerable<TestCaseData> SpecialConversions
```
stevenaw: Note to reviewer: These have been duplicated across test files because I couldn't find a good place to centralize them. Please let me know if you have any suggestions, or think its ok as-is.

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -100,6 +110,30 @@ public sealed class AnalyzeWhenArgumentIsCorrect
...
+    public sealed class AnalyzeWhenArgumentIsSpecialConversion
+    {
+        [TestCase(""" + value + @""")]
+        public void Test(" + targetType.Name + @" a) { }
+    }");
+            AnalyzerAssert.Valid<TestCaseUsageAnalyzer>(testCode);
+        }
+        
+        [Test]
+        public void AnalyzeWhenArgumentIsVerySpecialConversion()
```
mikkelbu: This test looks the same as the `TestCaseData("2019-10-10", typeof(DateTimeOffset))` case above?

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -15,6 +15,16 @@ namespace NUnit.Analyzers.Tests.TestCaseUsage
...
     {
         private DiagnosticAnalyzer analyzer = new TestCaseUsageAnalyzer();
+        
+        private static IEnumerable<TestCaseData> SpecialConversions
+        {
+            get
+            {
+                yield return new TestCaseData("2019-10-10", typeof(DateTime));
+                yield return new TestCaseData("23:59:59", typeof(TimeSpan));
+                yield return new TestCaseData("2019-10-10", typeof(DateTimeOffset));
```
mikkelbu: Can you also provide a testcase where we have time and offset for the `DateTimeOffset`?

src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
mikkelbu: I think it is fine as it is, but perhaps also add a testcase with a time and offset for `DateTimeOffset`.

src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
mikkelbu: Actually, I don't know the exact differences between `DateTimeOffset.TryParse` and `TypeConverter.CanConvertFrom`, so I'm unsure about this. But I think it will be the safest to have the same implementation, if it is not too cumbersome. Also I think we should use `CultureInfo.InvariantCulture`.

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
stevenaw: Oh you're right, thanks for spotting this! I'd been rushing to get this PR up before my laptop battery died, and I must've missed cleaning up this earlier attempt at a test.

src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
stevenaw: Thanks for the feedback. I think I have this working now using `TypeConverter` + `CultureInfo.InvariantCulture` :)

stevenaw:
Thanks for the review @mikkelbu 
I think I've covered all the feedback. I've also updated the tests I added to use string interpolation (instead of concatenation) since that seems to be what most of the tests do. I've also added 2 test cases to the existing `AttributeArgumentSyntaxExtensionsTests` to cover the positive/negative `DateTimeOffset`-related tests

PR #161, opened 2019-10-26
feat: Add analyzer and codefix for `Assert.AreSame`

### LINKED ISSUE #153: Constraint model suggestion for Assert.AreSame
```cs
Assert.AreSame(expected, actual);
```

Expected suggestion, since Assert.False has a constraint model suggestion:
```cs
Assert.That(actual, Is.SameAs(expected));
```

mikkelbu:
Fixes #153

### DIFF
--- a/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreSameClassicModelAssertUsageCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreSameClassicModelAssertUsageCodeFixTests.cs
@@ -1,0 +1,95 @@
+using System.Collections.Immutable;
+using Gu.Roslyn.Asserts;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Diagnostics;
+using NUnit.Analyzers.ClassicModelAssertUsage;
+using NUnit.Analyzers.Constants;
+using NUnit.Framework;
+
+namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
+{
+    [TestFixture]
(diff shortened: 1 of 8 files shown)


src/nunit.analyzers.tests/ClassicModelAssertUsage/AreSameClassicModelAssertUsageCodeFixTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        [Test]
+        public void VerifyAreSameFixWithMessageAndParams()
+        {
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+        public void TestMethod()
+        {{
+            var expected = new object();
+            var actual = new object();
+
+            ↓Assert.AreSame(expected, actual, ""message"", Guid.NewGuid());
```
JohanLarsson: An alternative is to put the diagnostic on the identifier. Not saying it is better, just an alternative. If the assert is multiline there will be a bunch of squiggle when having it on the expression. But we can start like this and see if it feels ergonomic.

JohanLarsson:
Looks good.

src/nunit.analyzers.tests/ClassicModelAssertUsage/AreSameClassicModelAssertUsageCodeFixTests.cs
mikkelbu: I had not considered how this looks when having a multiline assert (almost all of my asserts are on a single line 😄). I think I'll keep it as it is - for now - as this align with the other analyzers, but it is worth to consider, if we should change them all.

Issue #123, opened 2019-04-02
Generalise TestMethodUsageAnalyzer to support custom awaitables


mikkelbu:
We should generalise the work done in #122, so that it also supports custom awaitables - see e.g. nunit/nunit#3095. See in #49 for useful information.

PR #167, opened 2019-11-02
Add SameAsIncompatibleTypesAnalyzer

### LINKED ISSUE #154: Warning when SameAs compares expressions of incompatible types
```cs
class Foo { }
class Bar { }

var foo = new Foo();
var bar = new Bar();

// Should warn because there is no way that the same instance can be of type Foo and type Bar
Assert.That(foo, Is.SameAs(bar));
```

I think something like bidirectional IsAssignableFrom semantics are desired but I'm not quite sure.

Dreamescaper:
Fixes #154

### DIFF
--- a/src/nunit.analyzers.tests/DocumentationTests.cs
+++ b/src/nunit.analyzers.tests/DocumentationTests.cs
@@ -176,9 +176,9 @@ namespace NUnit.Analyzers.Tests
                 var length = 0;
                 while (startIndex + length < doc.Length)
                 {
-                    if (doc[startIndex + length] == '\n' &&
-                        doc.Length > startIndex + length &&
-                        doc[startIndex + length +1] != '|')
+                    if (doc.Length > startIndex + length &&
+                        doc[startIndex + length] == '\n' &&
+                        doc.ElementAtOrDefault(startIndex + length + 1) != '|')
                     {
                         length++;
                         return length;
--- a/src/nunit.analyzers.tests/SameAsIncompatibleTypes/SameAsIncompatibleTypesAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/SameAsIncompatibleTypes/SameAsIncompatibleTypesAnalyzerTests.cs
@@ -15,20 +15,20 @@ namespace NUnit.Analyzers.Tests.SameAsIncompatibleTypes
         [Test]
         public void AnalyzeWhenIncompatibleTypesProvided()
         {
-            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing($@"
-    class A {{ }}
-    class B {{ }}
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    class A { }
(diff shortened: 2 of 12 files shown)


Dreamescaper:
Forgot about delegate and Tasks actual types handling. Will add a bit later.

Dreamescaper:
Updated PR.
Added support for any awaitable, therefore will submit a PR for https://github.com/nunit/nunit.analyzers/issues/123 after this one merged (decided not to add it to this PR, even though not much changes there).

mikkelbu:
Thanks for another great contribution @Dreamescaper. I'll probably have time to review this one of the following days. I just merged the other PR, so there are some conflicts - minor as it is due to the same identifier used in both PRs.

Dreamescaper:
Sure, no rush!
Will resolve conflicts later.

src/nunit.analyzers.tests/DocumentationTests.cs
```diff
@@ -176,9 +176,9 @@ int TableLength()
                 var length = 0;
                 while (startIndex + length < doc.Length)
                 {
-                    if (doc[startIndex + length] == '\n' &&
-                        doc.Length > startIndex + length &&
-                        doc[startIndex + length +1] != '|')
+                    if (doc.Length > startIndex + length &&
```
Dreamescaper: It crashed with index out of bound before, if there was no empty line at the end in 'index.md'

mikkelbu: I can remember I added an empty line at some point as a quick solution, but I never came back to correct the code. So thanks for correcting this :)

src/nunit.analyzers.tests/SameAsIncompatibleTypes/SameAsIncompatibleTypesAnalyzerTests.cs
```diff
@@ -0,0 +1,329 @@
...
+    public class SameAsIncompatibleTypesAnalyzerTests
+    {
+        private static readonly DiagnosticAnalyzer analyzer = new SameAsIncompatibleTypesAnalyzer();
+        private static readonly ExpectedDiagnostic expectedDiagnostic =
+            ExpectedDiagnostic.Create(AnalyzerIdentifiers.SameAsIncompatibleTypes);
+
+        [Test]
+        public void AnalyzeWhenIncompatibleTypesProvided()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing($@"
```
mikkelbu: I think you can remove the $-part from all strings in this file and then also avoid the `{{` and `}}`. I can also do this and push an update to your branch, if you don't mind?

Dreamescaper: Sure I don't mind, thanks!

mikkelbu:
Thanks for the contribution @Dreamescaper

Issue #163, opened 2019-10-31
Feature: Assert.IsNull / Assert.IsNotNull Classic to Fluent


jcurl:
Thanks for the good analyzer. In my old project, I'm migrating to new style before upgrading to NUnit 3.x (no fixed date yet, but when I'm ready). 

A see no warnings for Assert.IsNull, Assert.IsNotNull.

## Expected
`Assert.IsNull(foo)` => `Assert.That(foo, Is.Null)`
`Assert.IsNotNull(foo)` => `Assert.That(foo.Is.Not.Null)`
`Assert.Null(foo)` => `Assert.That(foo, Is.Null)`
`Assert.NotNull(foo)` => `Assert.That(foo.Is.Not.Null)`

mikkelbu:
Sounds like an analyzer to me :).

jcurl:
I've had a shot at this, but I'm not familiar enough with Roslyn to do the code fix for Assert.That(obj, Is.Not.Null). 

Please take a look at 35819be800b8388fd720e48c22a5cd33d29d8ffe

Dreamescaper:
@jcurl 
I think it should be similar as existing `AreNotEqualClassicModelAssertUsageCodeFix`, 
but probably with something similar as that:
```csharp
            arguments.Insert(2, SyntaxFactory.Argument(
                    SyntaxFactory.MemberAccessExpression(
                        SyntaxKind.SimpleMemberAccessExpression,
                        SyntaxFactory.MemberAccessExpression(
                            SyntaxKind.SimpleMemberAccessExpression,
                            SyntaxFactory.IdentifierName(NunitFrameworkConstants.NameOfIs),
                            SyntaxFactory.IdentifierName(NunitFrameworkConstants.NameOfIsNot)),
                        SyntaxFactory.IdentifierName(NunitFrameworkConstants.NameOfIsNull))));
```

(haven't checked, but should work)

or even that:
```csharp
            arguments.Insert(2, SyntaxFactory.Argument(
                SyntaxFactory.ParseExpression($"{NameOfIs}.{NameOfIsNot}.{NameOfIsNull}")));
```

[...]

Issue #145, opened 2019-06-23
Make analyzers suggesting refactoring to the constraint model default disabled?


JohanLarsson:
For example `NUnit2001` warns on `Assert.False(expression)` and offers refactoring to `Assert.That(expression, Is.False)`.

From what I understand this is mostly a style choice and while opinionated analyzers are nice for consistency it may be poor ergonomics to push users away from their preferred API.

Downside to making it default disabled is that it will not be easily discoverable. Upside is that it makes the style choice configurable and opt-in.

We can perhaps use the GitHub API to query for usage statistics on the different APIs?

Dreamescaper:
I agree.
Unless we want to deprecate classic assertion model, it shouldn't be warning. 
I wouldn't disable it by default either. I would suggest making it Info.

mikkelbu:
There was a lot of discussion about the severity of the "ClassicModelAssertUsage" analyzers in #9 (search for Info in the comments), but I did not give it much thought when I took over the project (also, I am probably biased as I was advocating the constraint based model at my old job), so I just left the analyzers as a warning.

I agree with @Dreamescaper and think to begin with we should change the severity to Info and leave the analyzer enabled. This still makes it easy to use the fixer and the effect on the IDE is not too intrusive (just 3 dots). If/when we get more feedback then we can consider this choice.

Ps. We should also change the documentation of the analyzers as commented in https://github.com/nunit/nunit.analyzers/pull/144#discussion_r296458983.

Regarding querying the GitHub API for usage statistics then I think that it could be interesting, but I'm quite sure that this will be too large a task for me in any near future 😄.

OsirisTerje:
I am not sure I agree on this.
The default model is the constraint model, using the classic model means you add another layer.  
For the simple example above, I agree it doesnt matter much, but when you look at Assert.AreEqual syntax, the constraint model is less prone to user errors.  
IMHO I believe we *should* give out recommendations.   From courses and workshops I have run on this, many devs don't even know about the constraint model.  The analyzer then points them in the right direction, and the majority (in my experience) prefers it. 
 If a user disagree, it is easy to reconfigure anyway.

pantosha:
Any news? 
It seems that this bug is only one stopper for the first version releasing. It will be very helpful to resolve it and publish the first version at nuget.com.

mikkelbu:
@pantosha I hope to have something ready within a couple of weeks (my schedule is quite busy at the moment), so until then you can use the myget feed.

[...]

PR #168, opened 2019-11-03
Add Analyzers for Assert.[Is][Not]Null(expr)

jcurl:
Took the Assert.IsFalse() implementation and adapted it for Assert.Null, Assert.IsNull, Assert.NotNull and Assert.IsNotNull.

All unit test cases pass. Compiled and installed the VSIX into VS2019 and the analyzers are detected and code fixes work.

Issue: https://github.com/nunit/nunit.analyzers/issues/163

### DIFF
--- a/src/nunit.analyzers.tests/ClassicModelAssertUsage/IsNotNullAndNotNullClassicModelAssertUsageCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/ClassicModelAssertUsage/IsNotNullAndNotNullClassicModelAssertUsageCodeFixTests.cs
@@ -1,0 +1,91 @@
+using System.Collections.Immutable;
+using Gu.Roslyn.Asserts;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Diagnostics;
+using NUnit.Analyzers.ClassicModelAssertUsage;
+using NUnit.Analyzers.Constants;
+using NUnit.Framework;
+
+namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
+{
+    [TestFixture]
+    public sealed class IsNotNullAndNotNullClassicModelAssertUsageCodeFixTests
+    {
+        private static readonly DiagnosticAnalyzer analyzer = new ClassicModelAssertUsageAnalyzer();
+        private static readonly CodeFixProvider fix = new IsNotNullAndNotNullClassicModelAssertUsageCodeFix();
+
+        [Test]
+        public void VerifyGetFixableDiagnosticIds()
+        {
+            var fix = new IsNotNullAndNotNullClassicModelAssertUsageCodeFix();
+            var ids = fix.FixableDiagnosticIds.ToImmutableArray();
+
+            Assert.That(ids.Length, Is.EqualTo(2), nameof(ids.Length));
+            Assert.That(ids, Contains.Item(AnalyzerIdentifiers.IsNotNullUsage), nameof(AnalyzerIdentifiers.IsNotNullUsage));
+            Assert.That(ids, Contains.Item(AnalyzerIdentifiers.NotNullUsage), nameof(AnalyzerIdentifiers.NotNullUsage));
+        }
+
+        [TestCase("IsNotNull", AnalyzerIdentifiers.IsNotNullUsage)]
+        [TestCase("NotNull", AnalyzerIdentifiers.NotNullUsage)]
+        public void VerifyIsNotNullAndNotNullFixes(string assertion, string diagnosticId)
+        {
+            var expectedDiagnostic = ExpectedDiagnostic.Create(diagnosticId);
+
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+        public void TestMethod()
+        {{
+            object obj = null;
+            ↓Assert.{assertion}(obj);
+        }}");
+            var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void TestMethod()
+        {
+            object obj = null;
+            Assert.That(obj, Is.Not.Null);
+        }");
+            AnalyzerAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: CodeFixConstants.TransformToConstraintModelDescription);
+        }
+
+        [TestCase("IsNotNull", AnalyzerIdentifiers.IsNotNullUsage)]
+        [TestCase("NotNull", AnalyzerIdentifiers.NotNullUsage)]
+        public void VerifyIsNotNullAndNotNullFixesWithMessage(string assertion, string diagnosticId)
+        {
+            var expectedDiagnostic = ExpectedDiagnostic.Create(diagnosticId);
+
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+        public void TestMethod()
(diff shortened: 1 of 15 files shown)


src/nunit.analyzers/nunit.analyzers.csproj
```diff
@@ -5,7 +5,10 @@
     <TargetFramework>netstandard1.6</TargetFramework>
     <PackageTargetFallback>portable-net45+win8</PackageTargetFallback>
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
-    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
jcurl: I guess I should remove this. For some reason there were a lot of "errors" in the test case generated by SonarLint, which would be warnings otherwise. So I was looking for a way to turn it off.

src/nunit.analyzers.tests/ClassicModelAssertUsage/IsNotNullAndNotNullClassicModelAssertUsageCodeFixTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        private static readonly DiagnosticAnalyzer analyzer = new ClassicModelAssertUsageAnalyzer();
+        private static readonly CodeFixProvider fix = new IsNotNullAndNotNullClassicModelAssertUsageCodeFix();
+
+        [Test]
+        public void VerifyGetFixableDiagnosticIds()
+        {
+            var fix = new IsNotNullAndNotNullClassicModelAssertUsageCodeFix();
+            var ids = fix.FixableDiagnosticIds.ToImmutableArray();
+
+            Assert.That(ids.Length, Is.EqualTo(2), nameof(ids.Length));
```
Dreamescaper: you assert it in shorter way:
```csharp
Assert.That(ids, Is.EquivalentTo(new[] {AnalyzerIdentifiers.IsNotNullUsage, AnalyzerIdentifiers.NotNullUsage}));
```

no need to log `nameof(...)` in message - in case of failure NUnit will log that on its own

src/nunit.analyzers.tests/ClassicModelAssertUsage/IsNotNullAndNotNullClassicModelAssertUsageCodeFixTests.cs
```diff
@@ -0,0 +1,91 @@
...
+            Assert.That(ids, Contains.Item(AnalyzerIdentifiers.NotNullUsage), nameof(AnalyzerIdentifiers.NotNullUsage));
+        }
+
+        [TestCase("IsNotNull", AnalyzerIdentifiers.IsNotNullUsage)]
+        [TestCase("NotNull", AnalyzerIdentifiers.NotNullUsage)]
+        public void VerifyIsNotNullAndNotNullFixes(string assertion, string diagnosticId)
+        {
+            var expectedDiagnostic = ExpectedDiagnostic.Create(diagnosticId);
+
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
```
Dreamescaper: you can use `TestUtility.WrapInTestMethod` to make it a bit clearer

src/nunit.analyzers.tests/ClassicModelAssertUsage/IsNullAndNullClassicModelAssertUsageCodeFixTests.cs
```diff
@@ -0,0 +1,91 @@
...
+        private static readonly DiagnosticAnalyzer analyzer = new ClassicModelAssertUsageAnalyzer();
+        private static readonly CodeFixProvider fix = new IsNullAndNullClassicModelAssertUsageCodeFix();
+
+        [Test]
+        public void VerifyGetFixableDiagnosticIds()
+        {
+            var fix = new IsNullAndNullClassicModelAssertUsageCodeFix();
+            var ids = fix.FixableDiagnosticIds.ToImmutableArray();
+
+            Assert.That(ids.Length, Is.EqualTo(2), nameof(ids.Length));
```
Dreamescaper: same here

src/nunit.analyzers.tests/ClassicModelAssertUsage/IsNotNullAndNotNullClassicModelAssertUsageCodeFixTests.cs
jcurl: I used `IsTrueAndTrueClassicModelAssertUsageCodeFixTests.cs` as a template, where it is also there. Should this change be made in both places?

src/nunit.analyzers.tests/ClassicModelAssertUsage/IsNotNullAndNotNullClassicModelAssertUsageCodeFixTests.cs
jcurl: I used `IsTrueAndTrueClassicModelAssertUsageCodeFixTests.cs` as a template. Should the change be made in both places?

src/nunit.analyzers.tests/ClassicModelAssertUsage/IsNullAndNullClassicModelAssertUsageCodeFixTests.cs
jcurl: I used `IsTrueAndTrueClassicModelAssertUsageCodeFixTests.cs` as a template. Should the change be made in both places?

jcurl:
Is there anything that I still need to do for this review?

mikkelbu:
@jcurl I'll have to take a look at it :), but I've been very busy the last week. I'll review it now.

jcurl:
Thanks for responding, and fully understand. If you want me to implement the comment in the last patch, please let me know.

src/nunit.analyzers.tests/ClassicModelAssertUsage/IsNotNullAndNotNullClassicModelAssertUsageCodeFixTests.cs
mikkelbu: I agree with @Dreamescaper that the assertion could be expressed more clearly - and as a consequence of this also give a better error message. But as this seems to be a more general problem I think is is fine to keep the code as it is in this PR. I'll create another issue to solve all of the occurrences of this problem (or perhaps make it part of #145)

src/nunit.analyzers.tests/ClassicModelAssertUsage/IsNotNullAndNotNullClassicModelAssertUsageCodeFixTests.cs
mikkelbu: Same comment as above :). I agree with @Dreamescaper, but I also think it is better to solve all the occurrences in one go (and not part of this addition).

src/nunit.analyzers.tests/ClassicModelAssertUsage/IsNullAndNullClassicModelAssertUsageCodeFixTests.cs
mikkelbu: Same comment as in `IsNotNullAndNotNullClassicModelAssertUsageCodeFixTests.cs`.

PR #170, opened 2019-11-24
Add support for any awaitables in TestMethodUsageAnalyzer

### LINKED ISSUE #123: Generalise TestMethodUsageAnalyzer to support custom awaitables

Dreamescaper:
fixes #123

### DIFF
--- a/src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
@@ -302,7 +302,7 @@ namespace NUnit.Analyzers.Tests.TestCaseUsage
         public void AnalyzeWhenTestMethodHasCustomAwaitableReturnTypeAndExpectedResult()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-    public sealed class AnalyzeWhenTestMethodHasGenericTaskReturnTypeAndExpectedResult
+    public sealed class AnalyzeWhenTestMethodHasCustomAwaitableReturnTypeAndExpectedResult
     {
         [TestCase(ExpectedResult = 1)]
         public CustomAwaitable GenericTaskTestCaseWithExpectedResult()
@@ -330,7 +330,7 @@ namespace NUnit.Analyzers.Tests.TestCaseUsage
                 string.Format(TestMethodUsageAnalyzerConstants.ExpectedResultTypeMismatchMessage, typeof(int).Name));
 
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-    public sealed class AnalyzeWhenTestMethodHasGenericTaskReturnTypeAndExpectedResult
+    public sealed class AnalyzeWhenTestMethodHasCustomAwaitableReturnTypeAndExpectedResultIsIncorrect
     {
         [TestCase(ExpectedResult = '1')]
         public CustomAwaitable GenericTaskTestCaseWithExpectedResult()
--- a/src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
@@ -351,6 +351,36 @@ namespace NUnit.Analyzers.Tests.TestCaseUsage
         }
 
         [Test]
+        public void AnalyzeWhenTestMethodHasValueTaskReturnTypeAndExpectedResult()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                [TestCase(ExpectedResult = 1)]
+                public async ValueTask<int> GenericTaskTestCaseWithExpectedResult()
+                {
+                    return 1;
+                }");
+
+            AnalyzerAssert.Valid<TestMethodUsageAnalyzer>(testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenTestMethodHasValueTaskReturnTypeAndExpectedResultIsIncorrect()
+        {
+            var expectedDiagnostic = ExpectedDiagnostic.Create(
+                AnalyzerIdentifiers.TestMethodExpectedResultTypeMismatchUsage,
+                string.Format(TestMethodUsageAnalyzerConstants.ExpectedResultTypeMismatchMessage, typeof(int).Name));
+
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                [TestCase(ExpectedResult = '1')]
+                public async ValueTask<int> GenericTaskTestCaseWithExpectedResult()
+                {
+                    return 1;
+                }");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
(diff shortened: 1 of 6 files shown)


Dreamescaper:
I wanted to add tests for `ValueTask` as well, since it's the most popular non-`Task` awaitable type (I think), but I'm not really sure how to do that, since `ValueTask` is not available in `netstandard1.6`.

(Problem is only with test, analyzer itself supports `ValueTask` due to pattern-based approach)

mikkelbu:
@Dreamescaper Currently, the testproject targets `net461`, but I cannot see a problem in either bumping it to a newer framework version and referencing `System.Threading.Tasks.Extensions` or to e.g. bump it to `netcoreapp2.1`.

Dreamescaper:
@mikkelbu 
I've tried updating it to `netcoreapp2.1`, but half of tests failed. Probably something is wrong with how `Gu.Roslyn.Asserts` creates workspace, but not sure.

Will look into that later.

Dreamescaper:
@mikkelbu 
Updated test project to `netcoreapp2.1`, added tests for `ValueTask`.

src/nunit.analyzers.tests/TestUtility.cs
```diff
@@ -1,9 +1,7 @@
-using System.Linq;
-using System.Threading.Tasks;
 using Gu.Roslyn.Asserts;
 using NUnit.Framework;
 
-[assembly: MetadataReferences(typeof(Assert), typeof(object), typeof(Task), typeof(Enumerable))]
```
Dreamescaper: Previously most tests failed with issues reported by `Gu.Roslyn.Asserts` similar to:
>  CS0012 The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'.

Let's be honest, I have no idea why it failed, and why this change fixed that.

Dreamescaper:
I probably need help with cake script, as it fails to run tests now.

Dreamescaper:
Had to update cake script, since cannot use NUnit Console Runner with netcoreapp project. Let me know if that's ok.

mikkelbu:
Hi @Dreamescaper As usual I've not had much spare time this week, but I'll take a look at this during the weekend.

src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
```diff
@@ -298,6 +298,88 @@ public Task<int> GenericTaskTestCaseWithExpectedResult()
             AnalyzerAssert.Valid<TestMethodUsageAnalyzer>(testCode);
         }
 
+        [Test]
+        public void AnalyzeWhenTestMethodHasCustomAwaitableReturnTypeAndExpectedResult()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    public sealed class AnalyzeWhenTestMethodHasGenericTaskReturnTypeAndExpectedResult
```
mikkelbu: Nitpick
```suggestion
    public sealed class AnalyzeWhenTestMethodHasCustomAwaitableReturnTypeAndExpectedResult
```

src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
```diff
@@ -298,6 +298,88 @@ public Task<int> GenericTaskTestCaseWithExpectedResult()
...
+
+        [Test]
+        public void AnalyzeWhenTestMethodHasCustomAwaitableReturnTypeAndExpectedResultIsIncorrect()
+        {
+            var expectedDiagnostic = ExpectedDiagnostic.Create(
+                AnalyzerIdentifiers.TestMethodExpectedResultTypeMismatchUsage,
+                string.Format(TestMethodUsageAnalyzerConstants.ExpectedResultTypeMismatchMessage, typeof(int).Name));
+
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    public sealed class AnalyzeWhenTestMethodHasGenericTaskReturnTypeAndExpectedResult
```
mikkelbu:
```suggestion
    public sealed class AnalyzeWhenTestMethodHasCustomAwaitableReturnTypeAndExpectedResultIsIncorrect
```

src/nunit.analyzers.tests/TestUtility.cs
mikkelbu: As far as I can remember one needs to mention the classes used via `MetadataReferences`, so perhaps a `typeof(Object)` would also have solved the issue (although it seems from https://github.com/dotnet/standard/issues/542#issuecomment-530451163 that this is a general problem). @JohanLarsson can probably answer it better than I.

That being said the README of Gu.Roslyn.Asserts - https://github.com/GuOrg/Gu.Roslyn.Asserts/blob/master/README.md -  states `Below should not be needed (i.e. adding MetadataReferences), prefer TransitiveMetadataReferences as it updates as the resolution graph changes.`, so I think your solution is sound.

Dreamescaper:
> I like that we can use Roslyn to determine whether something is awaitable  (I thought we would have had to implement this logic ourselves)

Nope, that's not Roslyn. I did that in previous PR https://github.com/nunit/nunit.analyzers/pull/167 (that's why decided to fix this issue next thing).

Dreamescaper:
@mikkelbu 
Lack of tests is probably due to difference in how appveyor displays results from NUnit and TRX. In latter case it shows only test name, not full name. Might be that we have some duplicates in this case. Not sure if there's any way to change that.
In build log there is `Passed: 673` entry.

mikkelbu:
> > I like that we can use Roslyn to determine whether something is awaitable  (I thought we would have had to implement this logic ourselves)
> 
> Nope, that's not Roslyn. I did that in previous PR #167 (that's why decided to fix this issue next thing).
I should have remembered. The last many weeks my mind has been full with my daytime work, so most of the reviewing and open source work has been from day-to-day (where I cannot remember what I did the day before).

You are right that it is an issue in how appveyor displays the tests. The log of master says 669 and the log of the PR says 673 which corresponds to the 4 added tests.

I've extracted the tests and then "problem" is that we have tests like `NUnit.Analyzers.Tests.ConstraintsUsage.StringConstraintUsageCodeFixTests.AnalyzeStringBooleanMethod_AssertIsFalse("StartsWith","NUnit2012","Does.Not.StartWith")` and `NUnit.Analyzers.Tests.ConstraintsUsage.StringConstraintUsageAnalyzerTests.AnalyzeStringBooleanMethod_AssertIsFalse("StartsWith","NUnit2012","Does.Not.StartWith")` which differs in class name, but the method name is the same, so this is only reported once when tests are run using `dotnet test`. I can live with that 😄.

JohanLarsson:
Sorry I missed this thread.

Many times it is enough to do:
```cs
[assembly: TransitiveMetadataReferences(typeof(TypeInTestProject))]
```

It brings in the same references as the test project has, brings in nugets etc.

That can fail if the test project does not use `ValueTask` anywhere for example. If that happens one can specify more. There is also overloads for the asserts where one can pass in explicit metadata references but that can get noisy.

PR #171, opened 2019-11-25
Add EqualToIncompatibleTypesAnalyzer

### LINKED ISSUE #39: Warning if actual type does not match expected type
Currently there's no type checks for actual and expected results, so it's quite a common mistake (at least for me), which an analyzer could help to avoid.

Basic case: 
`Assert.That(arg1, Is.EqualTo(arg2))`
arg1 and arg2 should be of same type (at runtime).
From analyzer point of view the easiest way, I believe, is to check ClassifyConversion between typeSymbols.
Exceptions: between number types (e.g. int-double is fine), string and numbers. Anything else?
Case when custom comparer is provided should be skipped as well.

Other possible Asserts:
Collection asserts (contains, subset, superset etc), probably others

Dreamescaper:
fixes #39

Added analyzer for asserts using EqualConstraint for actual and expected types, which could not be possibly considered equal.

That became a bit more complex than I thought. 
I had to ignore equal asserts with All/Some/None/Property suffixes (at least for now), as they change actual type for constraint. 
(In fact, currently it is not too hard to make it account for those suffixes properly, but I decided not to make this PR even more complicated).
Besides, ignored asserts with custom EqualityComparer provided.

In order to do that, had to rework `AssertExpressionHelper`. We already had methods to split constraint by binary operators ('|', '&'), I had to add split to smaller constraint "parts" - by `.And` and `.Or` methods, and to be able to get those "suffixes" and "prefixes" for each part.

It's a bit tricky to work with those "parts", as there is no way to express half of an expression in Roslyn.

(I'm thinking, maybe we should add some abstraction on top of `ExpressionSyntax`, something like `ConstraintExpressionSyntax`, and to encapsulate those things there to make it easier to work with, but probably not in this PR)


src/nunit.analyzers/Constants/NunitFrameworkConstants.cs
```diff
@@ -63,9 +66,13 @@ public static class NunitFrameworkConstants
...
         public const string NameOfExpressionParameter = "expression";
 
+        public const string NameOfConstraintExpressionAnd = "And";
+        public const string NameOfConstraintExpressionOr = "Or";
+        public const string NameOfConstraintExpressionWith = "With";
+
         public const string NameOfIgnoreCase = "IgnoreCase";
+        public const string NameOfUsing = "Using";
 
-        public const string AssemblyQualifiedNameOfTypeAssert =
```
Dreamescaper: It was used only for assembly name check anyway

src/nunit.analyzers/Helpers/AssertExpressionHelper.cs
```diff
@@ -92,5 +119,105 @@ public static IEnumerable<ExpressionSyntax> SplitConstraintByOperators(Expressio
...
+
+            yield return constraintExpression;
+        }
+
+        private static (ExpressionSyntax expectedArgument, IMethodSymbol constraintMethod)? GetExpectedArgumentFromConstraintPart(
+            ExpressionSyntax constraintPart,
+            SemanticModel semanticModel,
+            CancellationToken cancellationToken)
+        {
+            // Each constraint part might have only one method, accepting 'expected' argument.
```
mikkelbu: I'm unsure about the might in the statement.

src/nunit.analyzers/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzer.cs
```diff
@@ -0,0 +1,268 @@
...
+                }
+
+                // Streams
+                if (IsStream(actualType, actualFullName) && IsStream(expectedType, expectedFullName))
+                    return true;
+
+                // IEquatables
+                if (IsIEquatable(actualType, expectedType) || IsIEquatable(expectedType, actualType))
+                    return true;
+            }
```
mikkelbu: Should we move the content of this if-statement up inside the `if (namedActualType != null && namedExpectedType != null)` expression on L127, so we don't have this minimal duplication?

src/nunit.analyzers/Helpers/AssertExpressionHelper.cs
```diff
@@ -52,32 +53,58 @@ internal static class AssertExpressionHelper
...
-                    var symbol = semanticModel.GetSymbolInfo(invocation, cancellationToken).Symbol;
-
-                    if (symbol is IMethodSymbol methodSymbol
-                        && methodSymbol.Parameters.Length == 1
-                        && methodSymbol.Parameters[0].Name == NunitFrameworkConstants.NameOfExpectedParameter)
-                    {
-                        var argument = invocation.ArgumentList.Arguments[0];
-                        expectedArguments.Add((argument.Expression, methodSymbol));
-                    }
+                    expectedArguments.Add(expected.Value);
```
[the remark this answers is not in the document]
Dreamescaper: Previously we split expression only by `&`/`|` operators, now - by `.And`/`.Or` properties as well.
Because of that I assume there might be only one (or none) expected argument per part.
Should have same result, if I haven't missed anything

src/nunit.analyzers/Helpers/AssertExpressionHelper.cs
```diff
@@ -52,32 +53,58 @@ internal static class AssertExpressionHelper
...
+                .Where(e => e is MemberAccessExpressionSyntax || e is InvocationExpressionSyntax)
+                .TakeWhile(e => !ReturnsConstraint(e, semanticModel));
+        }
+
+        public static IEnumerable<ExpressionSyntax> GetConstraintExpressionSuffixes(
+            ExpressionSyntax constraintExpression, SemanticModel semanticModel)
+        {
+            // e.g. Has.Property("Prop").Not.EqualTo("1").IgnoreCase
+            // -->
+            // Has.Property("Prop").Not.EqualTo("1").IgnoreCase
```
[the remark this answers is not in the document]
Dreamescaper: Probably not the best example, but yes.
Method will return either `MemberAccessExpression` (in case of property), or `InvocationExpression` (in case of method).

In this case `Name` property of expression would be `IgnoreCase`, but the whole expressions would be, well, the whole expression.

src/nunit.analyzers/Helpers/AssertExpressionHelper.cs
Dreamescaper: e.g. Is.Not.Null` has none

src/nunit.analyzers/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzer.cs
Dreamescaper: I tried to preserve same order as it goes in constraint itself, but probably KeyValuePair case is order independent

src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
```diff
@@ -0,0 +1,600 @@
...
+    class B { }
+
+    public class Tests
+    {
+        [Test]
+        public void AnalyzeWhenIncompatibleTypesProvided_WithNegatedAssert()
+        {
+            var actual = new A();
+            var expected = new B();
+            Assert.That(actual, Is.Not.EqualTo(↓expected));
```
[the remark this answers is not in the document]
Dreamescaper: Yes, probably description is wrong in this case.

Assert will always pass, but most likely it's invalid. Why would you assert something that always passes?

I can suggest either
a) change description to something more appropriate, like `therefore assertion is invalid`
b) create separate diagnostic for negative case

src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
```diff
@@ -0,0 +1,600 @@
...
+
+        [Test]
+        public void AnalyzeWhenIncompatibleTypesProvided_WithLocalFunctionActualValue()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    class A { }
+    class B { }
+
+    public class Tests
+    {{
```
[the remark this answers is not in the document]
Dreamescaper: Curious. Even there is compilation error, is successfully parses part where diagnostic expected, therefore test passes.
Fixed

Dreamescaper:
Replied to comments, applied suggestions and fixes.


> Perhaps sometime we should have tests that tests these methods more directly?

Makes sense. In separate PR?

WDYT about this instead of `AssertExpressionHelper`?
> (I'm thinking, maybe we should add some abstraction on top of ExpressionSyntax, something like ConstraintExpressionSyntax, and to encapsulate those things there to make it easier to work with.

I could submit wip PR with some examples later to show the idea, and if agreed - apply it for other analyzers, and add tests.

src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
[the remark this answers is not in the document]
mikkelbu: I think a) is fine as I don't think the (development/maintainability) cost of a separate diagnostic is worth it in this case.

src/nunit.analyzers/Helpers/AssertExpressionHelper.cs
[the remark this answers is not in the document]
mikkelbu: Sounds reasonable

src/nunit.analyzers/Helpers/AssertExpressionHelper.cs
mikkelbu: Instead of "Each constraint part might have only one method, accepting 'expected' argument." could we write "Each constraint part will at most have one method accepting an 'expected' argument."

src/nunit.analyzers/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzer.cs
mikkelbu: Ok. Then I think it is to leave it as it is.

src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
```diff
@@ -0,0 +1,644 @@
...
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenActualIsMatchingNullableType()
+        {
+            var testCode = TestUtility.WrapInTestMethod(@"
+            int? actual = 5;
+            string expected = ""5"";
+            Assert.That(actual, Is.EqualTo(expected));");
```
mikkelbu:
```suggestion
            Assert.That(actual, Is.EqualTo(↓expected));");
```

src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
```diff
@@ -0,0 +1,644 @@
...
+            var expected = new B();
+            Assert.That(actual, Is.Not.EqualTo(↓expected));
+        }
+    }");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenActualIsMatchingNullableType()
```
mikkelbu:
```suggestion
        public void AnalyzeWhenActualIsIncompatibleNullableType()
```

Dreamescaper:
Addressed comments.

PR #173, opened 2019-12-15
fix: Change severity of most classic assertions and improve documentation

### LINKED ISSUE #145: Make analyzers suggesting refactoring to the constraint model default disabled?

mikkelbu:
Also improves assertion on the number of diagnostic ids, as mentioned in
https://github.com/nunit/nunit.analyzers/pull/168#discussion_r341980254.

Fixes #145

### DIFF
--- a/src/nunit.analyzers.tests/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzerTests.cs
@@ -30,7 +30,7 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
                     $"{diagnostic.Id} : {nameof(DiagnosticDescriptor.MessageFormat)}");
                 Assert.That(diagnostic.Category, Is.EqualTo(Categories.Assertion),
                     $"{diagnostic.Id} : {nameof(DiagnosticDescriptor.Category)}");
-                Assert.That(diagnostic.DefaultSeverity, Is.EqualTo(DiagnosticSeverity.Warning),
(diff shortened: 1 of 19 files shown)


src/nunit.analyzers.tests/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzerTests.cs
```diff
@@ -30,7 +30,7 @@ public void VerifySupportedDiagnostics()
                     $"{diagnostic.Id} : {nameof(DiagnosticDescriptor.MessageFormat)}");
                 Assert.That(diagnostic.Category, Is.EqualTo(Categories.Assertion),
                     $"{diagnostic.Id} : {nameof(DiagnosticDescriptor.Category)}");
-                Assert.That(diagnostic.DefaultSeverity, Is.EqualTo(DiagnosticSeverity.Warning),
+                Assert.That(diagnostic.DefaultSeverity, Is.EqualTo(DiagnosticSeverity.Warning).Or.EqualTo(DiagnosticSeverity.Info),
```
Dreamescaper: alternative:

```suggestion
                Assert.That(diagnostic.DefaultSeverity, Is.AnyOf(DiagnosticSeverity.Warning, DiagnosticSeverity.Info),
```

Dreamescaper:
LGTM

PR #187, opened 2020-02-20
Fix EqualTo diagnostic when errors present

Dreamescaper:
Don't show incompatible types diagnostic when type info is missing.

This validation was already present, but didn't work for recursive checks (e.g. when we have array or list with unknown element type).

### DIFF
--- a/src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
@@ -770,8 +770,19 @@ namespace NUnit.Analyzers.Tests.EqualToIncompatibleTypes
         public void NoDiagnosticWhenActualIsErrorType()
         {
             var testCode = TestUtility.WrapInTestMethod(@"
-                var actual = 3;
-                var expected = HaveNoIdea();
+                var actual = HaveNoIdea();
+                var expected = 3;
+                Assert.That(actual, Is.EqualTo(expected));");
+
+            AnalyzerAssert.NoAnalyzerDiagnostics(analyzer, testCode);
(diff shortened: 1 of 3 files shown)


src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
```diff
@@ -754,5 +754,38 @@ public void NoDiagnosticWhenObjectComparedToInt()
...
+                Assert.That(actual, Is.EqualTo(expected));");
+
+            AnalyzerAssert.NoAnalyzerDiagnostics(analyzer, testCode);
+        }
+
+        [Test]
+        public void NoDiagnosticWhenActualIsErrorType()
+        {
+            var testCode = TestUtility.WrapInTestMethod(@"
+                var actual = 3;
```
mikkelbu: actual should be `HaveNoIdea()` (otherwise this is the same test as above).

src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
```diff
@@ -754,5 +754,38 @@ public void NoDiagnosticWhenObjectComparedToInt()
...
+                var expected = HaveNoIdea();
+                Assert.That(actual, Is.EqualTo(expected));");
+
+            AnalyzerAssert.NoAnalyzerDiagnostics(analyzer, testCode);
+        }
+
+        [Test]
+        public void NoDiagnosticWhenExpectedIsErrorTypeArray()
```
mikkelbu: Should we also have a test when actual is ErrorTypeArray?

Dreamescaper:
Thanks, fixed!

PR #190, opened 2020-03-22
TestCaseUsage - do not warn if type is unknown

### LINKED ISSUE #186: False error when Uri is parameter and string is argument
Next code leads to error, but it works.

    [TestCase( "http://www.site.com" )]
    public void Test(Uri value) {
    }

_The value of the argument at position '0' of type string cannot be assigned to the argument 'value' of type System.Uri_

Dreamescaper:
fixes #186 

Change is very simple - if type is unknown in runtime - do not show warnings.
We can add more supported cases, but I believe that this behavior (do not warn if don't know whether types can be converted) should remain anyway.

(Note - I added test case with Uri, but it does not reproduce issue itself, probably because of how we create test project)

### DIFF
--- a/src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
@@ -199,13 +199,13 @@ namespace NUnit.Analyzers.Tests.TestCaseUsage
         public void AnalyzeArgumentIsStringConvertedToEnum()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-    class AnalyzeArgumentIsIntConvertedToEnum
+    class AnalyzeArgumentIsStringConvertedToEnum
     {
         public enum TestEnum { A,B,C }
 
 
-        [TestCase(""A"")]
-        [TestCase(""B"")]
+        [TestCase(↓""A"")]
+        [TestCase(↓""B"")]
         public void Test(TestEnum e) { }
     }");
             AnalyzerAssert.Diagnostics<TestCaseUsageAnalyzer>(
@@ -217,12 +217,12 @@ namespace NUnit.Analyzers.Tests.TestCaseUsage
         public void AnalyzeArgumentIsEnumConvertedToString()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-    class AnalyzeArgumentIsEnumConvertedToInt
+    class AnalyzeArgumentIsEnumConvertedToString
     {
         public enum TestEnum { A,B,C }
 
-        [TestCase(TestEnum.A)]
-        [TestCase(TestEnum.B)]
+        [TestCase(↓TestEnum.A)]
+        [TestCase(↓TestEnum.B)]
         public void Test(string e) { }
     }");
             AnalyzerAssert.Diagnostics<TestCaseUsageAnalyzer>(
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -194,6 +195,41 @@ public enum TestEnum { A,B,C }
             AnalyzerAssert.Valid<TestCaseUsageAnalyzer>(testCode);
         }
 
+        [Test]
+        public void AnalyzeArgumentIsStringConvertedToEnum()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    class AnalyzeArgumentIsIntConvertedToEnum
```
mikkelbu:
```suggestion
    class AnalyzeArgumentIsStringConvertedToEnum
```

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -194,6 +195,41 @@ public enum TestEnum { A,B,C }
...
+    {
+        public enum TestEnum { A,B,C }
+
+
```
mikkelbu: Please remove one of the empty lines

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -194,6 +195,41 @@ public enum TestEnum { A,B,C }
...
+        [TestCase(""A"")]
+        [TestCase(""B"")]
```
mikkelbu:
```suggestion
        [TestCase(↓""A"")]
        [TestCase(↓""B"")]
```

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -194,6 +195,41 @@ public enum TestEnum { A,B,C }
...
+            AnalyzerAssert.Diagnostics<TestCaseUsageAnalyzer>(
+                ExpectedDiagnostic.Create(AnalyzerIdentifiers.TestCaseParameterTypeMismatchUsage),
+                testCode);
+        }
+
+        [Test]
+        public void AnalyzeArgumentIsEnumConvertedToString()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    class AnalyzeArgumentIsEnumConvertedToInt
```
mikkelbu:
```suggestion
    class AnalyzeArgumentIsEnumConvertedToString
```

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -194,6 +195,41 @@ public enum TestEnum { A,B,C }
...
+    {
+        public enum TestEnum { A,B,C }
+
+        [TestCase(TestEnum.A)]
+        [TestCase(TestEnum.B)]
```
mikkelbu:
```suggestion
        [TestCase(↓TestEnum.A)]
        [TestCase(↓TestEnum.B)]
```

Dreamescaper:
Updated

mikkelbu:
Looks good. Thanks for the contribution @Dreamescaper

PR #198, opened 2020-04-09
feat: Handle type conversion conservatively without loading types

### LINKED ISSUE #196: Avoid loading new assemblies into the VS process
Issue raised based on comments from @jnm2 and @JohanLarsson in https://gitter.im/GuOrg/Gu.Roslyn.Asserts?at=5e793252b08f3e0abf8c2224.

So instead of loading types and possibly new assemblies into the VS process we handle intrinsic type converters directly and will handle custom type converters in a subsequent issue - #197. If a converter (from string) exists then we assume that it is able to convert to the target type. We loose the ability to mark actual string values that does not satisfy the correct format as failures, but that is the price to pay using static analysis.

Extract from Gitter.
> This is not a good idea:
https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs#L154
this could load new assemblies into the VS process
Anything in the GAC or VS directories
...

mikkelbu:
Fixes #196 

A consequence of this is that we cannot determine the result from `Convert.ChangeType`, so we will assume that it is always successful.

### DIFF
--- a/src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
+++ b/src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
@@ -1,7 +1,3 @@
-using System;
-using System.Collections.Generic;
-using System.ComponentModel;
-using System.Globalization;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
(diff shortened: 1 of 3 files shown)


src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
```diff
@@ -69,35 +59,23 @@ internal static class AttributeArgumentSyntaxExtensions
                     {
                         canConvert = argumentType.SpecialType == SpecialType.System_String;
                     }
-
-                    if (canConvert)
+                    // Intrinsic type converters
+                    // https://github.com/dotnet/runtime/blob/master/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/ReflectTypeDescriptionProvider.cs
+                    else if (targetType.ContainingNamespace?.Name == "System" && (
+                        targetType.Name == "CultureInfo" ||
+                        targetType.Name == "DateTimeOffset" ||
```
siprbaum, line 66: Looking at the linked file for the Intrinsic type converters also lists
`DateTime` and `Version`. Should those also be added?

mikkelbu: `DateTime` was already covered by the previous `else if` case, but I think we could add `Version` (but I'll be doing it in a newly created PR based on the suggestions in https://github.com/nunit/nunit.analyzers/issues/196#issuecomment-611788092

mikkelbu:
Replaced by #199

PR #199, opened 2020-04-10
fix: Handle intrinsic type converters by hardcoded converters

### LINKED ISSUE #196: Avoid loading new assemblies into the VS process

mikkelbu:
Fixes #196

Note that `VersionConverter` and `CultureInfoConverter` does not exist in netstandard1.6.

### DIFF
--- a/src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
+++ b/src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
@@ -11,6 +11,28 @@ namespace NUnit.Analyzers.Extensions
 {
     internal static class AttributeArgumentSyntaxExtensions
     {
+        private static readonly IReadOnlyList<Type> ConvertibleTypes = new List<Type>
+        {
+            typeof(short),
+            typeof(byte),
+            typeof(long),
+            typeof(sbyte),
+            typeof(double),
+            typeof(decimal),
+            typeof(DateTime),
+        };
+
+        // Intrinsic type converters for types that are not SpecialTypes (and supported in netstandard1.6)
+        // https://github.com/dotnet/runtime/blob/master/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/ReflectTypeDescriptionProvider.cs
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
```diff
@@ -138,45 +145,39 @@ private static bool TryChangeType(ITypeSymbol targetType, object argumentValue)
...
             }
 
-            return $"{string.Join(".", namespaces)}.{typeName}";
-        }
+            var typeConverter = targetType.typeConverter.Value;
+            if (typeConverter.CanConvertFrom(argumentValue.GetType()))
+            {
+                try
+                {
+                    typeConverter.ConvertFrom(null, CultureInfo.InvariantCulture, argumentValue);
```
jnm2, line 171: Keep in mind that both `CanConvertFrom` and `ConvertFrom` may behave differently depending on which version of .NET Framework or .NET Core the analyzer is executing on. The framework and version that the analyzer is executing on will usually not match the version and framework that the tests execute on.

mikkelbu: Thanks for the warning. I'll keep that in mind. My plan is to go with this initially, but if we run into differences then we can choose the safer approach and just assume that strings can always be converted into the mentioned types.

Dreamescaper:
LGTM

PR #202, opened 2020-04-18
chore: Add nuget badge and update download section

mikkelbu:


Dreamescaper:
We're on Nuget? Yuy!!!)

mikkelbu:
For 2 days, but so far no downloads - if I can trust the counter to be updated), so I probably need to do some advertising 😄

PR #208, opened 2020-04-24
Analyzers List: Add icon for "enabled by default"

SeanKilleen:
To make it clear which analyzers are turned on by default (as of now, all of them).

Thinking it may be useful for the future when adding analyzers that are not enabled by default.

> image


mikkelbu:
I had been thinking the same, but had not done anything about it, so the contribution is much appreciated.

PR #213, opened 2020-04-26
Feat/support all testcase source ctors

### LINKED ISSUE #204: Extend TestCaseSourceUsesStringAnalyzer to handle all constructors of TestCaseSourceAttribute
The existing analyzer should handle all the constructors of the attribute:
* `TestCaseSourceAttribute(Type sourceType)`
* `TestCaseSourceAttribute(Type sourceType, string sourceName, object[] methodParams)`
* `TestCaseSourceAttribute(Type sourceType, string sourceName)`
* `TestCaseSourceAttribute(string sourceName)` (only ctor handled now)
* `TestCaseSourceAttribute(string sourceName, object[] methodParams)`

The attribute should satisfy the following requirements. The `sourceName` is a string representing the name of the source used to provide test cases. It has the following characteristics:
* It may be a field, property or method in the test class. (already handled now if one does not use `nameof`)
* It must be static. This is a change from NUnit 2.x.
* It must return an IEnumerable or a type that implements IEnumerable. For fields an array is generally used. For properties and methods, you may return an array or implement your own iterator.

The individual items returned by the enumerator must be compatible with the signature of the method on which the attribute appears. 

We should also check the following "error cases":
* "You have specified a data source field but also given a set of parameters. Fields cannot take parameters, please revise the 3rd parameter passed to the TestCaseSourceAttribute and either remove it or specify a method.";
* "You have specified a data source property but also given a set of parameters. Properties cannot take parameters, please revise the 3rd parameter passed to the TestCaseSource attribute and either remove it or specify a method.";
* "You have given the wrong number of arguments to the method in the TestCaseSourceAttribute, please check the number of parameters passed in the object is correct in the 3rd parameter for the TestCaseSourceAttribute and this matches the number of parameters in the target method and try again.";

See documentation at https://github.com/nunit/docs/wiki/TestCaseSource-Attribute

mikkelbu:
Fixes #204 (most of it - the rest will be descoped to another issue)


src/nunit.analyzers/Constants/TestCaseSourceUsageConstants.cs
```diff
@@ -5,5 +5,17 @@ internal static class TestCaseSourceUsageConstants
         internal const string ConsiderNameOfInsteadOfStringConstantAnalyzerTitle = "TestCaseSource should use nameof operator to specify target.";
         internal const string ConsiderNameOfInsteadOfStringConstantMessage = "Consider using nameof({0}) instead of \"{0}\".";
         internal const string ConsiderNameOfInsteadOfStringConstantDescription = "TestCaseSource should use nameof operator to specify target.";
+
+        internal const string SourceTypeNotINumerableTitle = "Source type does not implement IEnumerable.";
+        internal const string SourceTypeNotINumerableMessage = "Source type '{0}' does not implement IEnumerable.";
+        internal const string SourceTypeNotINumerableDescription = "The source type must implement IEnumerable in order to provide test cases.";
```
Dreamescaper, line 11: INumerable - typo

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
```diff
@@ -53,47 +83,235 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context)
...
+                // TODO Report parameters passed to field or property
+            }
+        }
+
+        private static SourceAttributeInformation ExtractInfoFromAttribute(
+            SyntaxNodeAnalysisContext context,
+            AttributeSyntax attributeSyntax)
+        {
+            var attributePositionalAndNamedArguments = attributeSyntax.GetArguments();
+            var positionalArguments = attributePositionalAndNamedArguments.Item1;
```
Dreamescaper, line 192: Can we change this method to return ValueTuple?...

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
```diff
@@ -53,47 +83,235 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context)
...
+            int? numMethodParams = null;
+            if (positionalArguments.Length > sourceNameIndex + 1)
+            {
+                numMethodParams = GetNumberOfParametersToMethod(positionalArguments[1]);
+            }
+
+            return new SourceAttributeInformation(sourceType, sourceName, syntaxNode, isStringLiteral, numMethodParams);
+        }
+
+        private static Tuple<SyntaxNode, string, bool> GetSyntaxStringConstantAndType(
```
Dreamescaper, line 260: Can we use ValueTuple here?
`(SyntaxNode syntaxNode, string sourceName, bool isLiteral)`

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
```diff
@@ -53,47 +83,235 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context)
...
+                        case IFieldSymbol field:
+                            ReportIfSymbolNotIEnumerable(field.Type);
+                            ReportIfParametersSupplied(attributeInfo.NumberOfMethodParameters);
+                            break;
+                        case IMethodSymbol method:
+                            ReportIfSymbolNotIEnumerable(method.ReturnType);
+
+                            if (method.Parameters.Length != attributeInfo.NumberOfMethodParameters)
+                            {
+                                // TODO Report mismatch of parameters
```
Dreamescaper, line 162: That's pending that another issue, that you mentioned, right?

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
mikkelbu: We can. Actually, I tried to do it at some point, but I got some strange compile errors. I could blame my preview VS 😄 , but I think I just did something wrong. It is ValueTuple now (and I'll make another PR get rid of the remaining Item1, ...).

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
mikkelbu: Yes we can

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
mikkelbu: Yes. All the TODOs will be solved by #214. I just wanted to do it in steps.

mikkelbu:
Thanks for the review @Dreamescaper, both for typo and for the comments regarding ValueTuple 👍

PR #217, opened 2020-04-29
Feat/add diagnostics testcasesource

### LINKED ISSUE #214: Add diagnostics and tests for all cases of TestCaseSourceUsesStringAnalyzer
This issue finishes the work related to #204.

The attribute should satisfy the following requirements. The sourceName is a string representing the name of the source used to provide test cases. It has the following characteristics:

* It may be a field, property or method in the test class. (already handled now if one does not use nameof)
* It must be static. This is a change from NUnit 2.x.
* It must return an IEnumerable or a type that implements IEnumerable. For fields an array is generally used. For properties and methods, you may return an array or implement your own iterator.
* The individual items returned by the enumerator must be compatible with the signature of the method on which the attribute appears.

We should also check the following "error cases":

"You have specified a data source field but also given a set of parameters. Fields cannot take parameters, please revise the 3rd parameter passed to the TestCaseSourceAttribute and either remove it or specify a method.";
"You have specified a data source property but also given a set of parameters. Properties cannot take parameters, please revise the 3rd parameter passed to the TestCaseSource attribute and either remove it or specify a method.";
"You have given the wrong number of arguments to the method in the TestCaseSourceAttribute, please check the number of parameters passed in the object is correct in the 3rd parameter for the TestCaseSourceAttribute and this matches the number of parameters in the target method and try again.";
See documentation at https://github.com/nunit/docs/wiki/TestCaseSource-Attribute

mikkelbu:
Closes #214

### DIFF
--- a/src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
@@ -70,27 +70,27 @@ namespace NUnit.Analyzers.Tests.TestCaseSourceUsage
         [TestCase("private static TestCaseData[] TestCases() => new TestCaseData[0];")]
         public void FixWhenStringLiteral(string testCaseMember)
         {
-            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing($@"
     public class AnalyzeWhenStringConstant
-    {
-        private static readonly TestCaseData[] TestCases = new TestCaseData[0];
+    {{
+        {testCaseMember}
 
         [TestCaseSource(↓""TestCases"")]
         public void Test()
-        {
-        }
-    }").AssertReplace("private static readonly TestCaseData[] TestCases = new TestCaseData[0];", testCaseMember);
+        {{
+        }}
+    }}");
 
-            var fixedCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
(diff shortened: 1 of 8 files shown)


src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
```diff
@@ -136,5 +136,284 @@ public void Test()
...
+
+        [TestCase("private static readonly TestCaseData[] TestCases = new TestCaseData[0];")]
+        [TestCase("private static TestCaseData[] TestCases => new TestCaseData[0];")]
+        [TestCase("private static TestCaseData[] TestCases() => new TestCaseData[0];")]
+        public void AnalyzeWhenSourceDoesProvideIEnumerable(string testCaseMember)
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    public class AnalyzeWhenSourceDoesProvideIEnumerable
+    {
+        private static readonly TestCaseData[] TestCases = new TestCaseData[0];
```
Dreamescaper, line 283: (suggestion, feel free to ignore)
imho, would be a bit more clear if you'd use something like `{testCaseSourceMember}` to make clear that this is a placeholder.
I was a bit confused for a moment why do you replace valid code.
(well, and because I didn't know what `AssertReplace` method does 😄 )

mikkelbu: I think the suggestion makes sense, so I've corrected all occurrences of this. Thanks 👍

PR #223, opened 2020-05-01
Add Nullable annotations

### LINKED ISSUE #222: Use nullable reference types
While we have no public API to annotate, using nullable reference types still will make code easier to write (would be clear whether I am required to do null checks), and safer from NullReferenceExceptions.

Since nullable attributes are not available for netstandard1.6, we need to add them. We can use:
- https://github.com/manuelroemer/Nullable - development-only package with nullable attributes
- https://github.com/tunnelvisionlabs/ReferenceAssemblyAnnotator - nullable attributes + black magic to add annotations to reference assemblies

Also, would be great to have nullability warnings from Roslyn, but it requires to updates Roslyn version to 3+. Not sure whether it's a great idea, as it means that VS2017 support is dropped.

Dreamescaper:
Resolves #222 .

Enabled nullable nullable reference types, added annotations where needed, fixed warnings.
Used https://github.com/manuelroemer/Nullable package (development-only dependency) for nullable attributes.

(If not suitable for any reason - feel free to close)

### DIFF
--- a/src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageCodeFix.cs
+++ b/src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageCodeFix.cs
@@ -71,7 +71,7 @@ namespace NUnit.Analyzers.ConstActualValueUsage
         }
 
         private static bool TryFindArguments(SemanticModel semanticModel, InvocationExpressionSyntax invocationSyntax,
-            out ExpressionSyntax expectedArgument, out ExpressionSyntax actualArgument)
+            out ExpressionSyntax? expectedArgument, out ExpressionSyntax? actualArgument)
         {
             expectedArgument = null;
             actualArgument = null;
--- a/src/nunit.analyzers/ConstraintUsage/SomeItemsConstraintUsageAnalyzer.cs
+++ b/src/nunit.analyzers/ConstraintUsage/SomeItemsConstraintUsageAnalyzer.cs
@@ -56,8 +56,8 @@ namespace NUnit.Analyzers.ConstraintUsage
         {
             return symbol is IMethodSymbol methodSymbol
                 && methodSymbol.Name == "Contains"
-                && (methodSymbol.IsInterfaceImplementation(typeof(ICollection<>)!.FullName!)
-                    || methodSymbol.IsInterfaceImplementation(typeof(ICollection)!.FullName!));
+                && (methodSymbol.IsInterfaceImplementation(typeof(ICollection<>).FullName!)
+                    || methodSymbol.IsInterfaceImplementation(typeof(ICollection).FullName!));
         }
     }
 }
--- a/src/nunit.analyzers/ConstraintUsage/SomeItemsConstraintUsageAnalyzer.cs
+++ b/src/nunit.analyzers/ConstraintUsage/SomeItemsConstraintUsageAnalyzer.cs
@@ -26,7 +26,7 @@ namespace NUnit.Analyzers.ConstraintUsage
 
         public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(descriptor);
 
-        protected override (DiagnosticDescriptor descriptor, string suggestedConstraint) GetDiagnosticData
+        protected override (DiagnosticDescriptor? descriptor, string? suggestedConstraint) GetDiagnosticData
             (SyntaxNodeAnalysisContext context, ExpressionSyntax actual, bool negated)
         {
             if (actual is InvocationExpressionSyntax invocationExpression
@@ -44,7 +44,7 @@ namespace NUnit.Analyzers.ConstraintUsage
             return (null, null);
         }
 
-        private bool IsLinqContains(ISymbol symbol)
+        private bool IsLinqContains(ISymbol? symbol)
         {
             return symbol is IMethodSymbol methodSymbol
                && methodSymbol.IsExtensionMethod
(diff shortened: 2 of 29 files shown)


src/nunit.analyzers/SameActualExpectedValue/SameActualExpectedValueAnalyzer.cs
```diff
@@ -38,7 +38,7 @@ public class SameActualExpectedValueAnalyzer : BaseAssertionAnalyzer
             {
                 context.ReportDiagnostic(Diagnostic.Create(
                     descriptor,
-                    expected.GetLocation(),
+                    expected!.GetLocation(),
```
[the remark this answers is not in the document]
Dreamescaper: In this case we know for sure that `expected` is not null because of line 35:
`.Where(e => actualExpression.IsEquivalentTo(e));`
(`IsEquivalent` returns false if input is null)

src/nunit.analyzers/SameAsIncompatibleTypes/SameAsIncompatibleTypesAnalyzer.cs
```diff
@@ -60,7 +60,7 @@ public class SameAsIncompatibleTypesAnalyzer : BaseAssertionAnalyzer
                 {
                     context.ReportDiagnostic(Diagnostic.Create(
                         descriptor,
-                        expectedArgumentExpression.GetLocation()));
+                        expectedArgumentExpression!.GetLocation()));
```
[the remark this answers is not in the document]
Dreamescaper: Line 42: 
`.Where(ex => ex != null)`
We know for sure that `expectedArgumentExpression` is not null, it's just compiler is not smart enough to know that as well.

src/nunit.analyzers/ConstraintUsage/SomeItemsConstraintUsageAnalyzer.cs
```diff
@@ -44,20 +44,20 @@ public class SomeItemsConstraintUsageAnalyzer : BaseConditionConstraintAnalyzer
...
 
-        private bool IsCollectionContains(ISymbol symbol)
+        private bool IsCollectionContains(ISymbol? symbol)
         {
             return symbol is IMethodSymbol methodSymbol
                 && methodSymbol.Name == "Contains"
-                && (methodSymbol.IsInterfaceImplementation(typeof(ICollection<>).FullName)
-                    || methodSymbol.IsInterfaceImplementation(typeof(ICollection).FullName));
+                && (methodSymbol.IsInterfaceImplementation(typeof(ICollection<>)!.FullName!)
+                    || methodSymbol.IsInterfaceImplementation(typeof(ICollection)!.FullName!));
```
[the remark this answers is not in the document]
Dreamescaper: I used `ReferenceAssemblyAnnotator` package originally, therefore FullName was annotated as nullable. Then I switched to `Nullable`, so annotation disappeared.
I decided to leave `!`, since might be needed anyway in future.

Still, suppression after `typeof` is definitely redundant, so removed it.

src/nunit.analyzers/SameAsIncompatibleTypes/SameAsIncompatibleTypesAnalyzer.cs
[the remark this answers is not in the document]
mikkelbu: My fault, I confused the C# type system with other type systems - where it would propagate the non-null property. Of course the typing of `Where` does not change the type of the elements, as it is just a filtering.

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageCodeFix.cs
```diff
@@ -71,7 +71,7 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
         }
 
         private static bool TryFindArguments(SemanticModel semanticModel, InvocationExpressionSyntax invocationSyntax,
-            out ExpressionSyntax expectedArgument, out ExpressionSyntax actualArgument)
+            out ExpressionSyntax? expectedArgument, out ExpressionSyntax? actualArgument)
```
jnm2, line 74: Sorry I'm late. Is this a case where `[NotNullWhen(true)]` would apply on each argument?

src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
```diff
@@ -55,7 +56,7 @@ internal static class AttributeArgumentSyntaxExtensions
             }
             else
             {
-                var targetType = GetTargetType(target);
+                ITypeSymbol? targetType = GetTargetType(target);
```
jnm2, line 59: The need for this is disappearing (has disappeared already?) in a newer version of the C# compilers, treated as a bugfix. The inferred type of `var` is always tracked as nullable now for the purpose of future assignments. This doesn't lose the tracked not-null state that the initial value gets.

src/nunit.analyzers/Extensions/ExpressionSyntaxExtensions.cs
```diff
@@ -78,7 +78,7 @@ public static string GetName(this ExpressionSyntax expression)
                 case IdentifierNameSyntax identifierName:
                     return identifierName.Identifier.Text;
                 default:
-                    return null;
+                    return "";
```
jnm2, line 81: Why isn't it better for the caller to explicitly handle the possibility that there was no name?

src/nunit.analyzers/SameActualExpectedValue/SameActualExpectedValueAnalyzer.cs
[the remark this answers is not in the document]
jnm2: > 
> 
> Personally, I would prefer that we don't have too many `!`s, but on the other hand it would also be annoying to have too many unnecessary null checks just to avoid `!`.

Debug.Assert or equivalent can help fill the gap too. If you were using ReferenceAssemblyAnnotator, that would be able to replace the `!`.

src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
Dreamescaper: Yeah, it compiled perfectly fine for me with VS preview, but failed in Appveyor build.

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageCodeFix.cs
Dreamescaper: Yep, my bad.

src/nunit.analyzers/Extensions/ExpressionSyntaxExtensions.cs
Dreamescaper: I suppose you're right, will submit separate PR.

src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
jnm2: Good to know.

PR #236, opened 2020-05-06
feat: Handle more cases in CodeFix correctly

### LINKED ISSUE #227: TestCaseSource: Fix codefix for nameof so that it also works when the source is in another class
Consider the following setup
```csharp
    public class MyClass
    {
        [TestCaseSource(typeof(AnotherClass), "MyTestSource")]
        public bool Value_ShouldEqual_One(int value)
        {
            return value == 1;
        }
    }

    public class AnotherClass
    {
        public static IEnumerable MyTestSource
        {
            get
            {
                yield return new TestCaseData(1).Returns(true);
                yield return new TestCaseData(0).Returns(false);
            }
        }
    }
```
Our codefix suggests to replace `"MyTestSource"` with `nameof(MyTestSource)`, but it should actually be `nameof(AnotherClass.MyTestSource)`.

We should handle that the source is in another class, a nested class, inside another class ...

mikkelbu:
Fixes #227

### DIFF
--- a/src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
+++ b/src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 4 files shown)


src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
```diff
@@ -164,11 +165,26 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context)
                         syntaxNode.GetLocation().SourceSpan.Start,
                         symbol);
 
-                    if (attributeInfo.IsStringLiteral && sourceIsAccessible)
+                    if (attributeInfo.IsStringLiteral && sourceIsAccessible && stringConstant != null)
```
Dreamescaper, line 168: Surprised to see this null check only here. Shouldn't it be somewhere higher?
Not sure that usages on lines 160 and 188 are supposed to accept null.

mikkelbu: You are right. I got an error that I could not palce stringConstant into `Dictionary<string, string>`, so I added the check without thinking much about it. Looking more closely, I think I'll move it up to 149, so that the condition becomes `syntaxNode == null || stringConstant == null`.

PR #237, opened 2020-05-08
Add ContainsConstraintWrongActualTypeAnalyzer

### LINKED ISSUE #221: Add warning when ContainsConstraint is used against incompatible argument type
```csharp
var actual = new [] {"1", "2", "3"};
var expected = 2;
Assert.That(actual, Does.Contain(expected));   <-- warning expected
```

actual should be either string or IEnumerable.
If actual is string, expected should be string as well.
If expected is IEnumerable<T>, expected should have matching with T type (using rules that we have in EqualToIncompatibleTypesAnalyzer.CanBeAssertedForEquality)

Dreamescaper:
Resolves #221 .

Examples:
```csharp
// The ContainsConstraint cannot be used with 'int' actual argument.
Assert.That(123, Does.Contain("1"));

// The ContainsConstraint cannot be used with 'int[]' actual argument.
Assert.That(new[] {1, 2, 3}, Does.Contain("1"));
```

Originally I thought that it is required to support collections with arbitrary element type, but it turned out that ContainsConstraint is used only with strings - `Does.Contain(object)` returns `SomeItemsConstraint` instead.

### DIFF
--- a/documentation/NUnit2025.md
+++ b/documentation/NUnit2025.md
@@ -12,11 +12,11 @@
 
 ## Description
 
-The ContainsConstraint constraint requires actual argument to be either string, or collection of strings.
+The ContainsConstraint requires actual value to be either a string or a collection of strings.
 
 ## Motivation
 
-Using ContainsConstraint with actual argument, which is not a string and not a collection of strings, leads to assertion error.
+Using a ContainsConstraint with an actual argument, which is neither a string nor a collection of strings, leads to an assertion error.
 
 ## How to fix violations
 
--- a/documentation/NUnit2025.md
+++ b/documentation/NUnit2025.md
@@ -1,0 +1,51 @@
+# NUnit2025
+## Wrong actual type used with ContainsConstraint.
+
+| Topic    | Value
+| :--      | :--
+| Id       | NUnit2025
+| Severity | Warning
+| Enabled  | True
+| Category | Assertion
+| Code     | [ContainsConstraintWrongActualTypeAnalyzer](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/ContainsConstraintWrongActualType/ContainsConstraintWrongActualTypeAnalyzer.cs)
+
+
+## Description
+
+The ContainsConstraint constraint requires actual argument to be either string, or collection of strings.
+
+## Motivation
+
+Using ContainsConstraint with actual argument, which is not a string and not a collection of strings, leads to assertion error.
+
+## How to fix violations
+
+Fix the actual value or use appropriate constraint.
+
+<!-- start generated config severity -->
+## Configure severity
+
+### Via ruleset file.
+
+Configure the severity per project, for more info see [MSDN](https://msdn.microsoft.com/en-us/library/dd264949.aspx).
+
+### Via #pragma directive.
+```C#
+#pragma warning disable NUnit2025 // Wrong actual type used with ContainsConstraint.
+Code violating the rule here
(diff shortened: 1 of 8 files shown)


mikkelbu:
Thanks for the PRs @Dreamescaper. I'll probably have time to review them tomorrow.

Dreamescaper:
I think I'll update PR with both `Does.Contain` cases.

documentation/NUnit2025.md
```diff
@@ -0,0 +1,51 @@
...
+| Code     | [ContainsConstraintWrongActualTypeAnalyzer](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/ContainsConstraintWrongActualType/ContainsConstraintWrongActualTypeAnalyzer.cs)
+
+
+## Description
+
+The ContainsConstraint constraint requires actual argument to be either string, or collection of strings.
+
+## Motivation
+
+Using ContainsConstraint with actual argument, which is not a string and not a collection of strings, leads to assertion error.
```
mikkelbu, line 19:
```suggestion
Using a ContainsConstraint with an actual argument, which is neither a string nor a collection of strings, leads to an assertion error.
```

src/nunit.analyzers/Constants/ContainsConstraintWrongActualTypeConstants.cs
```diff
@@ -0,0 +1,9 @@
+namespace NUnit.Analyzers.Constants
+{
+    internal static class ContainsConstraintWrongActualTypeConstants
+    {
+        public const string Title = "Wrong actual type used with ContainsConstraint.";
+        public const string Message = "The ContainsConstraint cannot be used with '{0}' actual argument.";
+        public const string Description = "The ContainsConstraint constraint requires actual argument to be either string, or collection of strings.";
```
mikkelbu, line 7:
```suggestion
        public const string Title = "Wrong actual type used with ContainsConstraint.";
        public const string Message = "The ContainsConstraint cannot be used with an actual value of type '{0}'.";
        public const string Description = "The Contains constraint requires actual value to be either a string or a collection of strings.";
```

Dreamescaper: You suggest to leave it as `ContainsConstraint` in title and message, but change to `Contains constraint` in description. Why not consistent?

Dreamescaper:
Updated PR based on your comment with an exception of `ContainsConstraint` -> `Contains constraint` change (see comment above).

src/nunit.analyzers/Constants/ContainsConstraintWrongActualTypeConstants.cs
mikkelbu: I think I wanted to changed the wording as the "The ContainsConstraint constraint" sounded a bit strange (we don't have other places where we have the substring "...Constraint constraint..."), and I found some places where we write "SameAs constraint" or "String Constraint". But you are right that this is inconsistent.

Perhaps the better option would be (also more consistent with the other analyzers):
```
        public const string Description = "The ContainsConstraint requires actual value to be either a string or a collection of strings.";
```

Dreamescaper: Yep, already done

mikkelbu: Great. I had only time to look at the comment, so I did not see the code changes.

mikkelbu:
Thanks for the PR 👍

PR #239, opened 2020-05-08
feat: Handle more conversions from string

### LINKED ISSUE #209: NUnit1001 and NUnit1005 complains about string to integer conversion
NUnit1001 complains with `"The value of the argument at position '0' of type string cannot be assigned to the argument 'numberValue' of type int."`, but the test can actually be executed, as the string is convertible to an int.
```csharp
        [TestCase("1")]
        public void SampleTest(int numberValue)
        {
            Assert.That(numberValue, Is.EqualTo(1));
        }
```

Also the following can be executed by NUnit, but the analyzer NUnit1005 complains
```csharp
[TestCase(1, ExpectedResult = "1")]
public int NUnit1005SampleTest(int inputValue)
{
    return inputValue;
}
```

mikkelbu:
Fixes #209

### DIFF
--- a/src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
+++ b/src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
@@ -22,15 +22,34 @@ namespace NUnit.Analyzers.Extensions
             typeof(DateTime),
         };
 
-        // Intrinsic type converters for types that are not SpecialTypes (and supported in netstandard1.6)
+        // Intrinsic type converters for types
         // https://github.com/dotnet/runtime/blob/master/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/ReflectTypeDescriptionProvider.cs
-        private static readonly List<(Type type, Lazy<TypeConverter> typeConverter)> IntrinsicTypeConverters =
-            new List<(Type type, Lazy<TypeConverter> typeConverter)>
+        // For converters that exists in .NET Standard 1.6 we use the converter.
+        // Otherwise, we assume that that the converter exists and it can convert the value.
+        private static readonly List<(Type type, Lazy<TypeConverter>? typeConverter)> IntrinsicTypeConverters =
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
```diff
@@ -22,15 +22,34 @@ internal static class AttributeArgumentSyntaxExtensions
             typeof(DateTime),
         };
 
-        // Intrinsic type converters for types that are not SpecialTypes (and supported in netstandard1.6)
+        // Intrinsic type converters for types
         // https://github.com/dotnet/runtime/blob/master/src/libraries/System.ComponentModel.TypeConverter/src/System/ComponentModel/ReflectTypeDescriptionProvider.cs
-        private static readonly List<(Type type, Lazy<TypeConverter> typeConverter)> IntrinsicTypeConverters =
-            new List<(Type type, Lazy<TypeConverter> typeConverter)>
+        // For converters that exists in .NET Standard 1.6 we use the converter.
+        // Otherwise, we assume that that the converter exists and it can convert the value.
```
Dreamescaper, line 28:
```suggestion
        // For converters that exist in .NET Standard 1.6 we use the converter.
        // Otherwise, we assume that the converter exists and it can convert the value.
```

src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
```diff
@@ -165,11 +184,16 @@ private static bool HasBuiltInImplicitConversion(ITypeSymbol argumentType, IType
             var typeName = targetTypeSymbol.GetFullMetadataName();
             var targetType = IntrinsicTypeConverters.FirstOrDefault(t => t.type.FullName == typeName);
 
-            if (targetType.typeConverter == null)
+            if (targetType == default((Type type, Lazy<TypeConverter> typeConverter)))
```
Dreamescaper, line 187:
```suggestion
            if (targetType == default)
```


(if it was added for clarity, feel free to ignore)

PR #240, opened 2020-05-09
Add SomeItemsIncompatibleTypesAnalyzer

Dreamescaper:
Related to #221 .

Examples:
```csharp
// The SomeItemsConstraint cannot be used with 'string[]' actual and 'int' expected arguments.
Assert.That(new[] {"1", "2"}, Does.Contain(1));

// The SomeItemsConstraint cannot be used with 'string[]' actual and 'int' expected arguments.
Assert.That(new[] {"1", "2"}, Contains.Item(1));
```

In order to re-use logic for types comparing, I moved method `CanBeEqual` from `EqualToIncompatibleTypesAnalyzer` to `NUnitEqualityComparerHelper`, and added tests for it.

I'd like to ask for advice about naming and messages. Technically it is for `SomeItemsConstraint` combined with `EqualConstraint`, which you get when you use `Does.Contain` or `Contains.Item` expressions.
So mentioning it as `SomeItemsConstraint` is not strictly correct. But I'm not sure how to name it better.

### DIFF
--- a/src/nunit.analyzers/Constants/SomeItemsIncompatibleTypesConstants.cs
+++ b/src/nunit.analyzers/Constants/SomeItemsIncompatibleTypesConstants.cs
@@ -3,7 +3,7 @@ namespace NUnit.Analyzers.Constants
     internal static class SomeItemsIncompatibleTypesConstants
     {
         public const string Title = "Wrong actual type used with the SomeItemsConstraint with EqualConstraint.";
-        public const string Message = "'{0}' constraint cannot be used with actual argument of type '{1}' and  expected argument of type '{2}'.";
+        public const string Message = "'{0}' constraint cannot be used with actual argument of type '{1}' and expected argument of type '{2}'.";
         public const string Description = "The SomeItemsConstraint with EqualConstraint requires the actual argument to be a collection where the element type can match the type of the expected argument.";
     }
 }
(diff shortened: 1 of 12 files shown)


src/nunit.analyzers/Constants/SomeItemsIncompatibleTypesConstants.cs
```diff
@@ -0,0 +1,9 @@
+namespace NUnit.Analyzers.Constants
+{
+    internal static class SomeItemsIncompatibleTypesConstants
+    {
+        public const string Title = "Wrong actual type used with SomeItemsConstraint.";
+        public const string Message = "The SomeItemsConstraint cannot be used with '{0}' actual and '{1}' expected arguments.";
+        public const string Description = "The SomeItemsConstraint requires actual argument to be a collection with matching element type to the expected argument.";
```
mikkelbu, line 7:
```suggestion
        public const string Message = "The SomeItemsConstraint cannot be used with actual argument of type '{0}' and  expected argument of type '{1}'.";
        public const string Description = "The SomeItemsConstraint requires the actual argument to be a collection where the element type can match the type of the expected argument.";
```

Dreamescaper:
I think it would make sense to use `SomeItemsConstraint with EqualConstraint` in analyzer description and title, and use user's actual constraint expression in message (`Does.Contain`/`Does.Not.Contain`/`Contains.Item` etc), e.g. `'Does.Contain' constraint cannot be used with actual argument of type 'int[]' and  expected argument of type 'string'.`.

mikkelbu:
I think that makes sense. For the end-user thinks about the constraint in terms of the syntax they have written and not in terms of the underlying classes.

Dreamescaper:
Updated PR based on review.

mikkelbu:
Great tests @Dreamescaper, in particular the ones in NUnitEqualityComparerHelperTests.cs. I've corrected my own typo and the name of one of the tests, but the rest looks great 💯 .

PR #248, opened 2020-06-11
feat: Add codefix for missing property analyzer

### LINKED ISSUE #233: Code fix switching Has.Length to Has.Count for NUnit2022
```cs
using NUnit.Framework;

class C
{
    [Test]
    public void M()
    {
        var collection = (IReadOnlyCollection<int>)new[] { 42 };

        // ⚠ NUnit2022 Actual argument type has no property 'Length'.
        // 🛠 Use 'Has.Count'   ↓↓↓↓↓↓↓↓↓↓
        Assert.That(collection, Has.Length.EqualTo(1));
    }
}
```

mikkelbu:
Fixes #233

### DIFF
--- a/src/nunit.analyzers.tests/MissingProperty/MissingPropertyCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/MissingProperty/MissingPropertyCodeFixTests.cs
@@ -29,7 +29,7 @@ namespace NUnit.Analyzers.Tests.MissingProperty
                 additionalUsings: "using System.Collections.Generic;");
 
             AnalyzerAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode,
-                fixTitle: string.Format(CodeFixConstants.MissingPropertyDescriptionFormat, "Count"));
+                fixTitle: string.Format(CodeFixConstants.UsePropertyDescriptionFormat, "Count"));
         }
 
         [Test]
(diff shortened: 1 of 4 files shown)


src/nunit.analyzers.tests/MissingProperty/MissingPropertyCodeFixTests.cs
```diff
@@ -0,0 +1,94 @@
...
+                Assert.That(actual, ↓Has.Message.EqualTo(2);");
+
+            AnalyzerAssert.NoFix(analyzer, fix, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void NoCodeFixForRandomMissingProperty()
+        {
+            var testCode = TestUtility.WrapInTestMethod(@"
+                var actual = 12345;
```
Dreamescaper, line 88: suggestion: could you make the actual here to have `Count` property? To check that code fix is not available when property is present, but original constraint is not suitable?

src/nunit.analyzers/Constants/CodeFixConstants.cs
```diff
@@ -5,5 +5,6 @@ internal static class CodeFixConstants
         internal const string TransformToConstraintModelDescription = "Transform to constraint model";
         internal const string SwapArgumentsDescription = "Swap actual and expected arguments";
         internal const string UseConstraintDescriptionFormat = "Use {0} constraint";
+        internal const string MissingPropertyDescriptionFormat = "Use '{0}' property";
```
Dreamescaper, line 8: I'd suggest to rename it to `UsePropertyDescriptionFormat`

mikkelbu:
Thanks for the suggestions and for the review @Dreamescaper. I've incorporated both of them.

PR #250, opened 2020-06-12
feat: Only report Nunit1007 on TestAttribute if no ITestBuilderAttribute exists

### LINKED ISSUE #212: Nunit1007 - ignore [Test] attribute when [TestCaseSource] is present
Consider the following sample:
```csharp
public static IEnumerable MyTestSource
{
    get
    {
        yield return new TestCaseData(1).Returns(true);
        yield return new TestCaseData(0).Returns(false);
    }
}

[Test]
[TestCaseSource(typeof(MyClass), nameof(MyTestSource))]
public bool Value_ShouldEqual_One(int value)
{
    return value == 1;
}
```

The above code generates the following error:
`NUnit1007 "Method has non-void return type 'bool', but no result is expected in ExpectedResult"`

**Problem**
But this is a false positive in this scenario. The `ExpectedResult` is actually provided through the `TestCaseSource` but the analyzer only looks at the `[Test]` attribute.

**Possible solution**
The NUnit1007 analyzer could just ignore the `[Test]` attribute when a `[TestCaseSource]` is present.

An alternative would be to raise an error when you add both `[Test]` and `[TestCaseSource]` but that seems overly restrictive because there's nothing really wrong with it (as I see it).

mikkelbu:
Fixes #212

### DIFF
--- a/src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
@@ -432,5 +432,73 @@ namespace NUnit.Analyzers.Tests.TestCaseUsage
     }");
             AnalyzerAssert.Valid<TestMethodUsageAnalyzer>(testCode);
         }
+
+        [Test]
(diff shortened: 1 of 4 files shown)


mikkelbu:
Thanks for the review @Dreamescaper

PR #251, opened 2020-06-13
feat: Add analyzer for ValueSourceAttribute

### LINKED ISSUE #93: Analyzer to ensure that the "target" of ValueSourceAttribute exists and is static
The target of `ValueSourceAttribute` should refer to a static field, property or method.

Much of the logic can probably be shared with #92.

In a subsequent task we can extend this with more of the requirements from https://github.com/nunit/docs/wiki/ValueSource-Attribute

For example
> If sourceType is specified, it represents the class that provides the data. **It must have a default constructor.**
> ...
> For the "target"
> * It must return an IEnumerable or a type that implements IEnumerable.
> * The individual items returned from the enumerator must be compatible with the type of the parameter on which the attribute appears.

mikkelbu:
Fixes #93


src/nunit.analyzers/ValueSourceUsage/ValueSourceUsageAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+
+                        if (method.Parameters.Length != 0)
+                        {
+                            context.ReportDiagnostic(Diagnostic.Create(
+                                methodExpectParameters,
+                                syntaxNode.GetLocation(),
+                                method.Parameters.Length));
+                        }
+                        break;
+                }
```
Dreamescaper, line 153: (feel free to ignore)

```suggestion                
                var memberType = symbol switch
                {
                    IPropertySymbol property => property.Type,
                    IFieldSymbol field => field.Type,
                    IMethodSymbol method => method.ReturnType,
                    _ => null
                };

                if (memberType != null)
                {
                    ReportIfSymbolNotIEnumerable(context, syntaxNode, memberType);
                }

                if(symbol is IMethodSymbol methodSymbol && methodSymbol.Parameters.Length != 0)
                {
                    context.ReportDiagnostic(Diagnostic.Create(
                        methodExpectParameters,
                        syntaxNode.GetLocation(),
                        methodSymbol.Parameters.Length));
                }
```

src/nunit.analyzers/ValueSourceUsage/ValueSourceUsageAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+                        { SourceCommonConstants.PropertyKeyNameOfTarget, nameOfTarget }
+                    };
+
+                    context.ReportDiagnostic(Diagnostic.Create(
+                        considerNameOfDescriptor,
+                        syntaxNode.GetLocation(),
+                        properties.ToImmutableDictionary(),
+                        nameOfTarget,
+                        stringConstant));
+                }
```
Dreamescaper, line 124: Could you extract that to separate method? E.g. `ReportToUseNameOfIfApplicable`.

Quite a lot happens in this method, a bit hard to keep track

src/nunit.analyzers/ValueSourceUsage/ValueSourceUsageAnalyzer.cs
Dreamescaper: Additional suggestion - to move the `memberType` switch to `ReportIfSymbolNotIEnumerable` method - to make `AnalyzeAttribute` a bit more concise.

mikkelbu: I actually tried to something similar, but it did not occur to me that I could use switch expressions (don't think that I've ever use them). I like the change 👍

src/nunit.analyzers/ValueSourceUsage/ValueSourceUsageAnalyzer.cs
mikkelbu: Sounds like a good suggestion, then we can also use it for the `TestCaseSourceUsesStringAnalyzer`

PR #252, opened 2020-06-15
feat: Ensure test methods are public

### LINKED ISSUE #245: Ensure that Test Methods are Public
Extracted from #5. We should find all test methods (so looking for the relevant interfaces and not just the normal `Test`, `TestCase`, ... attributes).

```csharp
        // Fails with "Method is not public"
        [Test]
        internal void Test()
        { }
```

mikkelbu:
Fixes #245

### DIFF
--- a/src/nunit.analyzers/Extensions/AttributeSyntaxExtensions.cs
+++ b/src/nunit.analyzers/Extensions/AttributeSyntaxExtensions.cs
@@ -55,11 +55,11 @@ namespace NUnit.Analyzers.Extensions
         private static bool DerivesFromInterface(
             SemanticModel semanticModel,
             AttributeSyntax attributeSyntax,
-            string fullyQualifiedMetadataName)
+            string interfaceTypeFullName)
         {
-            var ITestBuilderType = semanticModel.Compilation.GetTypeByMetadataName(fullyQualifiedMetadataName);
+            var interfaceType = semanticModel.Compilation.GetTypeByMetadataName(interfaceTypeFullName);
 
-            if (ITestBuilderType == null)
+            if (interfaceType == null)
                 return false;
 
             var attributeType = semanticModel.GetTypeInfo(attributeSyntax).Type;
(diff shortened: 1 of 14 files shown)


src/nunit.analyzers/Extensions/AttributeSyntaxExtensions.cs
```diff
@@ -42,15 +42,27 @@ internal static (ImmutableArray<AttributeArgumentSyntax> positionalArguments, Im
...
-                NunitFrameworkConstants.FullNameOfTypeITestBuilder);
+            return DerivesFromInterface(semanticModel, @this, NunitFrameworkConstants.FullNameOfTypeITestBuilder);
+        }
+
+        private static bool DerivesFromInterface(
+            SemanticModel semanticModel,
+            AttributeSyntax attributeSyntax,
+            string fullyQualifiedMetadataName)
+        {
+            var ITestBuilderType = semanticModel.Compilation.GetTypeByMetadataName(fullyQualifiedMetadataName);
```
Dreamescaper, line 60: var name seems to be wrong

src/nunit.analyzers/TestMethodAccessibilityLevel/TestMethodAccessibilityLevelAnalyzer.cs
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        private static void AnalyzeMethod(SyntaxNodeAnalysisContext context)
+        {
+            var methodNode = context.Node as MethodDeclarationSyntax;
+
+            if (methodNode != null && !methodNode.ContainsDiagnostics)
+            {
+                var testCaseType = context.SemanticModel.Compilation.GetTypeByMetadataName(NunitFrameworkConstants.FullNameOfTypeTestCaseAttribute);
+                var testType = context.SemanticModel.Compilation.GetTypeByMetadataName(NunitFrameworkConstants.FullNameOfTypeTestAttribute);
```
Dreamescaper, line 40: why do you need those two variables? Seems like they are not used?

mikkelbu: Initially, I kept them to ensure that we are in a NUnit context. But `IsTestMethod` serves the same purpose, so you are right that these are not needed.

PR #254, opened 2020-06-29
feat: Add diagnostic for simple test method with parameters

### LINKED ISSUE #215: Expected a warning when no arguments are specified for a parameterized test method
There should be a warning that arguments are required.

The warning shouldn't be shown if an attribute is applied to the method that implements the same interface as `TestCaseAttribute` and `TestCaseSourceAttribute` or if an attribute is applied to each parameter that implements the same interface as `ValuesAttribute` and `ValueSourceAttribute`.

```cs
public class C
{
    [Test]
    public void M(int p)
    {
        Assert.That(p, Is.EqualTo(42));
    }
}
```

Do you think a fix could be nice? I had:

```cs
        [Test]
        public static void Examples_from_docs(string path, string encoded)
        {

        }
```

And maybe there could be a fix that does this:

```diff
-       [Test]
+       [TestCase("path", "encoded")]
        public static void Examples_from_docs(string path, string encoded)
        {

        }
```

mikkelbu:
Fixes #215

### DIFF
--- a/src/nunit.analyzers/Constants/TestMethodUsageAnalyzerConstants.cs
+++ b/src/nunit.analyzers/Constants/TestMethodUsageAnalyzerConstants.cs
@@ -26,5 +26,9 @@ namespace NUnit.Analyzers.Constants
         internal const string AsyncExpectedResultAndNonGenericTaskReturnTypeTitle = "The async test method must have a Task<T> return type when a result is expected.";
         internal const string AsyncExpectedResultAndNonGenericTaskReturnTypeMessage = "The async test method must have a Task<T> return type when a result is expected, but the return type was '{0}'.";
         internal const string AsyncExpectedResultAndNonGenericTaskReturnTypeDescription = "The async test method must have a Task<T> return type when a result is expected.";
+
+        internal const string SimpleTestMethodHasParametersTitle = "The test method has parameters, but no arguments are supplied by attributes.";
+        internal const string SimpleTestMethodHasParametersMessage = "The test method has '{0}' parameter(s), but only '{1}' argument(s) are supplied by attributes.";
+        internal const string SimpleTestMethodHasParametersDescription = "The test method has parameters, but no arguments are supplied by attributes.";
     }
 }
--- a/src/nunit.analyzers/TestMethodUsage/TestMethodUsageAnalyzer.cs
+++ b/src/nunit.analyzers/TestMethodUsage/TestMethodUsageAnalyzer.cs
@@ -62,10 +62,20 @@ namespace NUnit.Analyzers.TestCaseUsage
                 defaultSeverity: DiagnosticSeverity.Error,
                 description: TestMethodUsageAnalyzerConstants.AsyncExpectedResultAndNonGenericTaskReturnTypeDescription);
 
+        private static readonly DiagnosticDescriptor simpleTestHasParameters =
+            DiagnosticDescriptorCreator.Create(
+                id: AnalyzerIdentifiers.SimpleTestMethodHasParameters,
+                title: TestMethodUsageAnalyzerConstants.SimpleTestMethodHasParametersTitle,
+                messageFormat: TestMethodUsageAnalyzerConstants.SimpleTestMethodHasParametersMessage,
+                category: Categories.Structure,
+                defaultSeverity: DiagnosticSeverity.Error,
+                description: TestMethodUsageAnalyzerConstants.SimpleTestMethodHasParametersDescription);
+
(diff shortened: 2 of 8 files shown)


src/nunit.analyzers/TestMethodUsage/TestMethodUsageAnalyzer.cs
```diff
@@ -104,10 +113,41 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context)
...
+                    }
                 }
             }
         }
 
+        private static SyntaxNode? GetMethod(SyntaxNode node)
+        {
+            foreach (var ancestor in node.Ancestors())
+            {
+                if (ancestor.IsKind(SyntaxKind.ParameterList))
```
Dreamescaper, line 143: Could you leave comment why we stop iteration if parameter list found?

src/nunit.analyzers/TestMethodUsage/TestMethodUsageAnalyzer.cs
```diff
@@ -104,10 +113,41 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context)
...
+                    var testMethodParameters = methodNode.ParameterList.Parameters.Count;
+                    var hasISimpleTestBuilderAttribute = HasISimpleTestBuilderAttribute(context.SemanticModel, methodNode.AttributeLists);
+                    var hasITestBuilderAttribute = HasITestBuilderAttribute(context.SemanticModel, methodNode.AttributeLists);
+                    var parametersMarkedWithIParameterDataSourceAttribute = ParametersMarkedWithIParameterDataSourceAttribute(
+                        context.SemanticModel, methodNode);
+
+                    if (testMethodParameters > 0 &&
+                        hasISimpleTestBuilderAttribute &&
+                        !hasITestBuilderAttribute &&
+                        parametersMarkedWithIParameterDataSourceAttribute < testMethodParameters)
```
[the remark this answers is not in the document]
mikkelbu: `parametersMarkedWithIParameterDataSourceAttribute` can newer be larger than `testMethodParameters` as this is a filtering of the parameters. I'll refactor the code to make this a bit more explicit.

src/nunit.analyzers/Constants/TestMethodUsageAnalyzerConstants.cs
```diff
@@ -26,5 +26,9 @@ internal static class TestMethodUsageAnalyzerConstants
         internal const string AsyncExpectedResultAndNonGenericTaskReturnTypeTitle = "The async test method must have a Task<T> return type when a result is expected.";
         internal const string AsyncExpectedResultAndNonGenericTaskReturnTypeMessage = "The async test method must have a Task<T> return type when a result is expected, but the return type was '{0}'.";
         internal const string AsyncExpectedResultAndNonGenericTaskReturnTypeDescription = "The async test method must have a Task<T> return type when a result is expected.";
+
+        internal const string SimpleTestMethodHasParametersTitle = "The test method has parameters, but no arguments are supplied by attributes.";
+        internal const string SimpleTestMethodHasParametersMessage = "The test method has '{0}' parameter(s), but only '{1}' argument(s) are supplied by attributes.";
```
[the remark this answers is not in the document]
mikkelbu: This cannot happen - see comment below :)

src/nunit.analyzers/TestMethodUsage/TestMethodUsageAnalyzer.cs
mikkelbu: Actually, I'm not proud of this method, and I'm very open for improvements/suggestions. It is an attempt to only examine the attributes on the method and not the attributes on the parameters of the method.

The problem was that before this method we just took `context.Node.Ancestors().OfType<MethodDeclarationSyntax>().FirstOrDefault()` to find the method related to the attribute, but this both work for attributes on the method and attributes on the parameters of the method.
So for the following code we would mark both `[Test]` and `[Values(1, 2, 3)]` with the diagnostic even though I only intended to mark `[Test]`:
```csharp
[Test]
public void M(int p, [Values(1, 2, 3)] int q, int r)
{
    Assert.That(p, Is.EqualTo(q + r));
}
```

I'll also think of how to rephrase the code/this method.

mikkelbu: To begin with I've decided to replace the method - and the previous `context.Node.Ancestors().OfType<MethodDeclarationSyntax>` -  with `var declarationNode = context.Node.Parent?.Parent;` as this solves the problem and is less cryptic than the `GetMethod` method.

PR #259, opened 2020-07-12
ConstActualValueUsageAnalyzer: #256: Allow Assert on Enum.Member

### LINKED ISSUE #256: NUnit2007 should not fire if both sides are constant
We have NUnit tests in place to assure that the value of an enumeration member has a certain value.
This triggers NUnit2007 because the actual value is constant.

```
        public enum YesNo
        {
            No = 0,
            Yes = 1,
        }

        [Test]
        public void TestEnumValue()
        {
            Assert.AreEqual(1, YesNo.Yes);
            Assert.That(YesNo.Yes, Is.EqualTo(1));
        }
```

manfred-brands:
This fixes #256 
It will always fire if actual is a literal as that is the classic error case.
It will however not fire if actual is a constant field and expected is also a constant.
This allows validating values of named enumeration values and constants.

### DIFF
--- a/src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
@@ -97,7 +97,7 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
         public void AnalyzeWhenConstFieldArgumentIsProvidedForAreEqual()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-            public class TestFixure
+            public class TestFixture
             {
                 private const string actual = ""act"";
 
@@ -115,7 +115,7 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
         public void AnalyzeWhenConstFieldArgumentIsProvidedForAssertThat()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-            public class TestFixure
+            public class TestFixture
             {
                 private const string actual = ""act"";
 
@@ -159,7 +159,7 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
         public void ValidWhenNonConstValueIsProvidedAsActualArgumentForAreEqual()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-            public class TestFixure
+            public class TestFixture
             {
                 private const string expected = ""exp"";
 
@@ -177,7 +177,7 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
         public void ValidWhenConstValueIsProvidedAsActualAndExpectedArgumentForAreEqual()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-            public class TestFixure
+            public class TestFixture
             {
                 private const string expected = ""exp"";
 
@@ -195,7 +195,7 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
         public void ValidWhenNonConstValueIsProvidedAsActualNamedArgumentForAreEqual()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-            public class TestFixure
+            public class TestFixture
             {
                 private const string expected = ""exp"";
 
@@ -213,7 +213,7 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
         public void ValidWhenConstValueIsProvidedAsActualAndExpectedNamedArgumentForAreEqual()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-            public class TestFixure
+            public class TestFixture
             {
                 private const string expected = ""exp"";
 
@@ -231,7 +231,7 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
         public void ValidWhenNonConstValueIsProvidedAsActualArgumentForAssertThat()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-            public class TestFixure
+            public class TestFixture
             {
                 private const string expected = ""exp"";
 
@@ -249,7 +249,7 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
         public void ValidWhenConstValueIsProvidedAsActualAndExpectedArgumentForAssertThat()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-            public class TestFixure
+            public class TestFixture
             {
                 private const string expected = ""exp"";
 
@@ -267,7 +267,7 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
         public void ValidWhenCheckingEnumerationValueForAssertThat()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-            public class TestFixure
+            public class TestFixture
             {
                 private enum Answer { No = 0, Yes = 1 };
 
@@ -284,7 +284,7 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
         public void ValidWhenCheckingAgainstStringEmptyAssertThat()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-            public class TestFixure
+            public class TestFixture
             {
                 private const string actual = ""act"";
 
--- a/src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
@@ -130,6 +130,32 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
         }
 
         [Test]
+        public void AnalyzeWhenStringEmptyArgumentIsProvidedForAreEqual()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                public void Test()
+                {
+                    string actual = ""act"";
+                    Assert.AreEqual(actual, string.Empty);
+                }");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenStringEmptyArgumentIsProvidedForAssertThat()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                public void Test()
+                {
+                    string actual = ""act"";
+                    Assert.That(string.Empty, Is.EqualTo(actual));
+                }");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
         public void ValidWhenNonConstValueIsProvidedAsActualArgumentForAreEqual()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageAnalyzer.cs
```diff
@@ -24,16 +26,56 @@ public class ConstActualValueUsageAnalyzer : BaseAssertionAnalyzer
...
+                // Check for Assert.That
+                if (AssertHelper.TryGetActualAndConstraintExpressions(assertExpression, context.SemanticModel,
+                    out _, out var constraintExpression))
+                {
+                    expectedExpression = constraintExpression.ConstraintParts
+                        .Select(part => part.GetExpectedArgumentExpression())
+                        .Where(e => e != null)
+                        .FirstOrDefault();
+                }
+            }
```
Dreamescaper, line 76: nit: could you extract `GetExpectedExpression` to separate method?

manfred-brands:
Addressed review request.
Noticed that negative numbers were not considered to be constants. Added code to detect constant literal expression in general.
Added special exemption for '.Empty' in cases like: Assert.AreEqual(actual, string.Empty)

manfred-brands:
Added a fixup for the .Empty case to allow: Assert.That(SomeCollection.Empty, Is.Empty)

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageAnalyzer.cs
```diff
@@ -24,21 +26,91 @@ public class ConstActualValueUsageAnalyzer : BaseAssertionAnalyzer
...
+                var argumentSymbol = context.SemanticModel.GetSymbolInfo(expression).Symbol;
+
+                return (argumentSymbol is ILocalSymbol localSymbol && localSymbol.IsConst) ||
+                       (argumentSymbol is IFieldSymbol fieldSymbol && fieldSymbol.IsConst);
+            }
+
+            bool IsEmpty(ExpressionSyntax expression)
+            {
+                return expression is MemberAccessExpressionSyntax memberAccessExpression &&
+                    memberAccessExpression.Name.Identifier.Text == "Empty";
```
Dreamescaper, line 60: Personally I don't mind considering `string.Empty` as const for this analyzer, but I'm not sure it applies to any `Empty` property possible.
I would limit it by strings only.

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageAnalyzer.cs
```diff
@@ -24,21 +26,91 @@ public class ConstActualValueUsageAnalyzer : BaseAssertionAnalyzer
...
+                if (IsLiteralExpression(expression))
+                    return true;
+
+                var argumentSymbol = context.SemanticModel.GetSymbolInfo(expression).Symbol;
+
+                return (argumentSymbol is ILocalSymbol localSymbol && localSymbol.IsConst) ||
+                       (argumentSymbol is IFieldSymbol fieldSymbol && fieldSymbol.IsConst);
+            }
+
+            bool IsEmpty(ExpressionSyntax expression)
```
Dreamescaper, line 57:
```suggestion
            bool IsStringEmpty(ExpressionSyntax expression)
```

IsEmpty implies that `expression` itself is empty, so would be better to rename it somehow

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageAnalyzer.cs
manfred-brands: I can live with that.

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageAnalyzer.cs
manfred-brands: Done.

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageAnalyzer.cs
```diff
@@ -24,21 +26,91 @@ public class ConstActualValueUsageAnalyzer : BaseAssertionAnalyzer
...
+            if (AssertHelper.TryGetActualAndConstraintExpressions(assertExpression, semanticModel,
+                out _, out var constraintExpression))
+            {
+                return constraintExpression.ConstraintParts
+                        .Select(part => part.GetExpectedArgumentExpression())
+                        .Where(e => e != null)
+                        .FirstOrDefault();
+            }
+
+            return null;
```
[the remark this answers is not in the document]
manfred-brands: Done.

manfred-brands:
Added a fixup commit for the items mentioned.

src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
```diff
@@ -128,11 +129,37 @@ public void Test()
...
         }
 
+        [Test]
+        public void AnalyzeWhenStringEmptyArgumentIsProvidedForAreEqual()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                public void Test()
+                {
+                    string actual = ""act"";
+                    Assert.AreEqual(actual, string.Empty);
```
mikkelbu, line 139:
```suggestion
                    string expected = ""exp"";
                    Assert.AreEqual(expected, ↓string.Empty);
```

src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
```diff
@@ -128,11 +129,37 @@ public void Test()
...
+        }
+
+        [Test]
+        public void AnalyzeWhenStringEmptyArgumentIsProvidedForAssertThat()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                public void Test()
+                {
+                    string actual = ""act"";
+                    Assert.That(string.Empty, Is.EqualTo(actual));
```
mikkelbu, line 152:
```suggestion
                    string expected = ""exp"";
                    Assert.That(↓string.Empty, Is.EqualTo(expected));
```

src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
manfred-brands: No, string.Empty is the expected value. The code was supposed to test if an actual value is string.Empty. As the arguments are reversed, this should trigger an analyzer violation. String.Empty is treated as if it is a constant.

src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
manfred-brands: No, string.Empty is the expected value. The code was supposed to test if an actual value is string.Empty. As the arguments are reversed, this should trigger an analyzer violation. String.Empty is treated as if it is a constant.
One wouldn't write this code, but it could be the result of the CodeFix to convert Classic AreEqual into Constraint model.

manfred-brands:
What is the ↓ character you added? I've seen it in some tests, but don't know what it is for now why.
When I extract the string to actual code the C# compiler complains about that character.

src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
mikkelbu: Ok. Then I misunderstod the intent. The signature of `Assert.AreEqual` is `Assert.AreEqual(object expected, object actual)`, so the first argument is the expected value and the second argument is the actual value - one reason to prefer to use `Assert.That(actual, Is.EqualTo(expected))` syntax.

src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
mikkelbu: Similar comment as above. The signature of `Assert.That` is `Assert.That<TActual>(TActual actual, IResolveConstraint expression)`, so the first argument is the actual value and the last is the expected value expressed as a constraint.

mikkelbu:
The `↓` character is a special character for this test framework that specifies where the diagnosis will occur. If you don't specify it the test framework will just check that the specified diagnosis will fire, but not where in the syntax it will fire.

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageAnalyzer.cs
```diff
@@ -24,21 +27,90 @@ public class ConstActualValueUsageAnalyzer : BaseAssertionAnalyzer
         protected override void AnalyzeAssertInvocation(SyntaxNodeAnalysisContext context,
             InvocationExpressionSyntax assertExpression, IMethodSymbol methodSymbol)
         {
+            bool IsLiteralExpression(ExpressionSyntax expression)
```
mikkelbu, line 30: I think we should have at least one test per case in this method - so that we don't introduce regressions later on.

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageAnalyzer.cs
```diff
@@ -24,21 +27,90 @@ public class ConstActualValueUsageAnalyzer : BaseAssertionAnalyzer
...
-                || (argumentSymbol is ILocalSymbol localSymbol && localSymbol.IsConst)
-                || (argumentSymbol is IFieldSymbol fieldSymbol && fieldSymbol.IsConst))
+            // The actual expression is a constant field, check if expected is also constant
+            var expectedExpression = this.GetExpectedExpression(assertExpression, methodSymbol, context.SemanticModel);
+
+            if (expectedExpression != null && !IsConstant(expectedExpression) && !IsStringEmpty(expectedExpression))
             {
                 context.ReportDiagnostic(Diagnostic.Create(
                     descriptor,
                     actualExpression.GetLocation()));
```
mikkelbu, line 94:
```suggestion
                Report(actualExpression);
```

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageAnalyzer.cs
```diff
@@ -24,21 +27,90 @@ public class ConstActualValueUsageAnalyzer : BaseAssertionAnalyzer
...
+
             var actualExpression = assertExpression.GetArgumentExpression(methodSymbol, NunitFrameworkConstants.NameOfActualParameter);
 
             if (actualExpression == null)
                 return;
 
-            var argumentSymbol = context.SemanticModel.GetSymbolInfo(actualExpression).Symbol;
+            if (IsLiteralExpression(actualExpression))
+            {
+                // Classic error case
```
mikkelbu, line 79: I think this comment can be removed as the code is self explanatory
```suggestion
```

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageAnalyzer.cs
```diff
@@ -24,21 +27,90 @@ public class ConstActualValueUsageAnalyzer : BaseAssertionAnalyzer
...
+                Report(actualExpression);
+                return;
+            }
+
+            if (!IsConstant(actualExpression) && !IsStringEmpty(actualExpression))
+                return; // Standard use case
```
mikkelbu, line 85: Only a suggestion
```suggestion
                return;
```

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageAnalyzer.cs
manfred-brands: There are use cases in `AnalyzeWhenLiteralArgumentIsProvidedForAssertThat`: `-1` triggers `PrefixUnaryExpressionSyntax` and the `(2 + 3) * 1024` triggers all the others:  `ParenthesizedExpressionSyntax`, `BinaryExpressionSyntax`, `LiteralExpressionSyntax` Would you prefer these to be in separate test methods?

mikkelbu: My fault, I overlooked those (so much for reviewing the tests one day and the prod code the next). I think it is fine as it is.

PR #262, opened 2020-07-13
#258: Deal with untyped Tasks

### LINKED ISSUE #258: NUnit2023 fires on variables of type Task
The following code results in an 'The type of the actual argument - 'void' - can never be null.

```
Task task = Task.Run(Do.Nothing);
Assert.That(task, Is.Not.Null);
```

The source code does an Unwrap of an awaitable type regardless if it is actually awaited:
```
            if (actualType.IsAwaitable(out var awaitReturnType))
                actualType = awaitReturnType;
```

Roslyn should already do that for us when called like:
```
                Task task = Task.Run(Do.Nothing);
                Assert.That(await task, Is.Not.Null);
```

manfred-brands:
Fixes #258 like in:
```
var Task = Task.CompletedTask;
Assert.That(task, Is.Not.Null);
Assert.That(task, Is.SameAs(Task.CompletedTask))
```

### DIFF
--- a/src/nunit.analyzers.tests/NullConstraintUsage/NullConstraintUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/NullConstraintUsage/NullConstraintUsageAnalyzerTests.cs
@@ -84,5 +84,16 @@ namespace NUnit.Analyzers.Tests.NullConstraintUsage
 
             AnalyzerAssert.Valid(analyzer, testCode);
         }
+
+        [TestCase("Is.Null")]
+        [TestCase("Is.Not.Null")]
+        public void ValidWhenActualIsTask(string constraint)
+        {
+            var testCode = TestUtility.WrapInTestMethod($@"
+                var task = Task.CompletedTask;
+                Assert.That(task, {constraint});");
+
+            AnalyzerAssert.Valid(analyzer, testCode);
+        }
     }
 }
--- a/src/nunit.analyzers.tests/NullConstraintUsage/NullConstraintUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/NullConstraintUsage/NullConstraintUsageAnalyzerTests.cs
@@ -35,6 +35,16 @@ namespace NUnit.Analyzers.Tests.NullConstraintUsage
 
         [TestCase("Is.Null")]
         [TestCase("Is.Not.Null")]
+        public void AnalyzeWhenActualIsTaskDelegate(string constraint)
+        {
+            var testCode = TestUtility.WrapInTestMethod($@"
+                Assert.That(() => Task.FromResult(3), ↓{constraint});");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [TestCase("Is.Null")]
+        [TestCase("Is.Not.Null")]
         public void ValidWhenActualIsNullableValueType(string constraint)
         {
             var testCode = TestUtility.WrapInTestMethod($@"
--- a/src/nunit.analyzers.tests/SameAsIncompatibleTypes/SameAsIncompatibleTypesAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/SameAsIncompatibleTypes/SameAsIncompatibleTypesAnalyzerTests.cs
@@ -177,6 +177,17 @@ namespace NUnit.Analyzers.Tests.SameAsIncompatibleTypes
         }
 
         [Test]
+        public void AnalyzeWhenActualAndExpectedTypesAreSame_WithTaskActualValue()
+        {
+            var testCode = TestUtility.WrapInTestMethod(@"
+                var actual = Task.FromResult("""");
+                var expected = """";
+                Assert.That(actual, Is.SameAs(↓expected));");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
         public void NoDiagnosticWhenActualAndExpectedTypesAreSame()
         {
             var testCode = TestUtility.WrapInTestMethod(@"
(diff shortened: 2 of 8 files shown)


src/nunit.analyzers/Helpers/AssertHelper.cs
```diff
@@ -46,7 +46,7 @@ public static ITypeSymbol UnwrapActualType(ITypeSymbol actualType)
             if (actualType is INamedTypeSymbol namedType && namedType.DelegateInvokeMethod != null)
                 actualType = namedType.DelegateInvokeMethod.ReturnType;
 
-            if (actualType.IsAwaitable(out var awaitReturnType))
+            if (actualType.IsAwaitable(out var awaitReturnType) && awaitReturnType.SpecialType != SpecialType.System_Void)
                 actualType = awaitReturnType;
 
             return actualType;
```
Dreamescaper, line 52:
```suggestion
            if (actualType is INamedTypeSymbol namedType && namedType.DelegateInvokeMethod != null)
            {
                actualType = namedType.DelegateInvokeMethod.ReturnType;

	            if (actualType.IsAwaitable(out var awaitReturnType))
	                actualType = awaitReturnType;
            }

            return actualType;
```


Seems like it should unwrap tasks only for delegates.

manfred-brands:
I agree with your changes, but this fails several NUnit tests.
But these tests are the ones that deal with the wrong assumption of Tasks being evaluated, e.g.: in StringConstraintWrongActualTypeAnalyzerTests.ValidWhenStringTaskValueProvided
```
Task<string> actual = Task.FromResult("1234");
Assert.That(actual, Does.StartWith("12"));
```

Instead of deleting the tests, I have reversed then from 'Valid' to 'Analyze' and assert they do fire.

src/nunit.analyzers.tests/NullConstraintUsage/NullConstraintUsageAnalyzerTests.cs
```diff
@@ -33,6 +33,16 @@ public void AnalyzeWhenActualIsNonNullableValueTypeDelegate(string constraint)
             AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
         }
 
+        [TestCase("Is.Null")]
+        [TestCase("Is.Not.Null")]
+        public void ValidWhenActualIsTaskDelegate(string constraint)
```
mikkelbu, line 38:
```suggestion
        public void AnalyzWhenActualIsTaskDelegate(string constraint)
```

src/nunit.analyzers.tests/SameAsIncompatibleTypes/SameAsIncompatibleTypesAnalyzerTests.cs
```diff
@@ -176,6 +176,17 @@ public void AnalyzeWhenIncompatibleTypesProvided_WithNegatedAssert()
             AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
         }
 
+        [Test]
+        public void AnalyzeWhenActualAndExpectedTypesAreSame_WithTaskActualValue()
+        {
+            var testCode = TestUtility.WrapInTestMethod(@"
+                var actual = Task.FromResult("""");
+                var expected = """";
+                Assert.That(actual, Is.SameAs(expected));");
```
mikkelbu, line 185:
```suggestion
                Assert.That(actual, Is.SameAs(↓expected));");
```

src/nunit.analyzers.tests/SomeItemsIncompatibleTypes/SomeItemsIncompatibleTypesAnalyzerTests.cs
```diff
@@ -44,6 +44,17 @@ public void AnalyzeWhenInvalidCollectionActualArgumentProvided()
                 testCode);
         }
 
+        [Test]
+        public void AnalyzeWhenActualIsCollectionTask()
+        {
+            var testCode = TestUtility.WrapInTestMethod(
+                $"Assert.That(Task.FromResult(new[] {{1,2,3}}), {this.constraint}(1));");
```
mikkelbu, line 51:
```suggestion
                $"Assert.That(Task.FromResult(new[] {{1,2,3}}), ↓{this.constraint}(1));");
```

src/nunit.analyzers.tests/StringConstraintWrongActualType/StringConstraintWrongActualTypeAnalyzerTests.cs
```diff
@@ -38,20 +38,20 @@ public void AnalyzeWhenNonStringValueProvided([ValueSource(nameof(StringConstrai
         }
 
         [Test]
-        public void ValidWhenStringValueProvided([ValueSource(nameof(StringConstraints))] string stringConstraint)
+        public void AnalyzeWhenStringTaskValueProvided([ValueSource(nameof(StringConstraints))] string stringConstraint)
         {
             var testCode = TestUtility.WrapInTestMethod($@"
-                var actual = ""1234"";
+                var actual = Task.FromResult(""1234"");
                 Assert.That(actual, {stringConstraint});");
```
mikkelbu, line 45:
```suggestion
                Assert.That(actual, ↓{stringConstraint});");
```

Dreamescaper:
> Instead of deleting the tests, I have reversed then from 'Valid' to 'Analyze' and assert they do fire.

I would suggest to change to delegate assert to test the actual use case.

```
Task<string> actual = Task.FromResult("1234");
Assert.That(() => actual, Does.StartWith("12"));
```

src/nunit.analyzers.tests/NullConstraintUsage/NullConstraintUsageAnalyzerTests.cs
manfred-brands: Renamed to AnalyzeWhenActualIsTaskDelegate (added e to your suggestion)

manfred-brands:
@Dreamescaper Instead of replacing the test, I added a new one, so we cover both normal delegates and delegates returning Task<T>.

manfred-brands:
@mikkelbu I have rewritten the branch into two commits, my original one, dealing with untyped Task and one that corrects the presumed behaviour that NUnit evaluates tasks, moving those tests from Valid to Analyze status.
You can merge it when you are ready.

mikkelbu:
@manfred-brands Thanks for the contribution. I'll be merging it now.

Issue #191, opened 2020-03-23
Make it possible to run tests both in .NET framework and .NET Core


mikkelbu:
To capture issues like #186 we need to run tests both in a .NET framework setting and in a .NET Core setting.

Dreamescaper:
Simply adding target framework didn't help to reproduce that issue - I've tried :) .
I suppose that's because we create our roslyn project not with design-time assemblies, but with forwarded ones, therefore analyzer is able to find Uri class.

manfred-brands:
We can dual target the nunit.analyzers.tests project by using plural TargetFrameworks:
`<TargetFrameworks>net461;netcoreapp2.1</TargetFrameworks>`

This causes 2 test failures on `AnalyzeWhenTestMethodHasValueTaskReturnTypeAndExpectedResult` and `AnalyzeWhenTestMethodHasValueTaskReturnTypeAndExpectedResultIsIncorrect`.
We can add ValueTask support to .net461 by adding:
`<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" Condition="'$(TargetFramework)' == 'net461'" />`

However the tests still fail with roslyn saying the `methodSymbol.ReturnType` is an `ErrorType`.

Unless one of you knows how to add support to the analyzer itself for ValueTask, I suggest to make these test conditional using `#if !NET461` as ValueTasks were not part of .net framework 4.6 anyhow.

I added a PR (#263) for this.

mikkelbu:
Closed by #263

PR #263, opened 2020-07-22
Target both .NET Framework (4.6.1) and .NET Core (2.1) for testing.

manfred-brands:
Addresses issue #191


mikkelbu (approved):
LGTM even though it will not help us with the original problem - #186 (see e.g. https://github.com/nunit/nunit.analyzers/issues/191#issuecomment-602853589), but I'll merge it now to have a clean slate for the next release.

One consequence of this is that we double the test time, but for now I don't think that this is a major issue as testing takes approx 30 per target on the CI.

PR #264, opened 2020-07-22
Update .editorconfig to match overall code-style used.

### LINKED ISSUE #261: Braces for simple 'if' needs updating in .editorconfig
We need to add the following to .editorconfig to make it match the used style for braces:

```
# IDE0011: Add braces
csharp_prefer_braces = when_multiline:warning
```

manfred-brands:
Fix two multi-line 'if' statements by adding braces.
Fix a few places that were not conforming to the overall code-style.

Fixes #261

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -17,7 +17,7 @@ charset = utf-8
 indent_size = 2
 
 [*]
-# https://github.com/nunit/docs/wiki/Coding-Standards#namespace-class-structure-interface-enumeration-and-method-definitions
+# https://docs.nunit.org/articles/developer-info/Coding-Standards.html#namespace-class-structure-interface-enumeration-and-method-definitions
 csharp_new_line_before_open_brace = all
 csharp_new_line_before_else = true
 csharp_new_line_before_catch = true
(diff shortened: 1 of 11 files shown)


.editorconfig
```diff
@@ -50,27 +50,44 @@ csharp_space_between_method_declaration_parameter_list_parentheses = false
...
+# Also 'apparent' does not mean the same for everyone.
+# Disable diagnostics since no firm policy is in place yet and the messages hide other violations.
+csharp_style_var_for_built_in_types = true:none
+csharp_style_var_when_type_is_apparent = true:none
+csharp_style_var_elsewhere = false:none
+
+# Disabled rule as convention is not mentioned in https://docs.nunit.org/articles/developer-info/Coding-Standards.html
+# Mentioned is 'if braces are not used on a control float statements with a single line, a blank line should follow'
+# This is not something we can configure.
+# The document doesn't say that it should have braces for multi-line statements, but maybe that is implied.
```
mikkelbu, line 82: The editorconfig in this project is partly extracted from the existing code we inherited - by a MS tool AFAICR - and partly inspired by the editorconfig from https://raw.githubusercontent.com/nunit/nunit/master/.editorconfig - which tries to match the coding standard.
I don't mind that we differ a little from the NUnit standard, but I think the comments would confuse readers in the future, so I propose that we just keep it simple.

```suggestion
# This convention is not mentioned in https://docs.nunit.org/articles/developer-info/Coding-Standards.html
```

manfred-brands: I am all for consistency, but coding guidelines that are vague and cannot be expressed in .editorconfig are not helping.
In my company, in the main shared component repository all rules are set to error in order to fail the build so consistency is maintained.

mikkelbu: I think one reason for this is that the NUnit coding standards existed long before .editorconfig (and probably also StyleCop and other products). Furthermore, there has been some discussion about both how precise the guidelines should be and also to which degree this should be enforced by tooling.

As I agree with enforcing the style more strictly, I'll create a new issue for the next milestone about increasing the severity of the rules from suggestion to warning/error. Then we can see if - and how - this will effect the contributors.

Issue #147, opened 2019-08-06
Analyzer for non-test methods to be non-public


Dreamescaper:
In our team we have a convention that only test methods can be public in test classes. If any helper methods are present for some reason, they should be private. Would be great to have an analyzer for that.

Probably should be disabled by default though.

mikkelbu:
Sounds reasonable.

We should probably include information of whether an analyzer is enabled by default or not both in the documentation of the analyzer and in the index https://github.com/nunit/nunit.analyzers/blob/master/documentation/index.md.

manfred-brands:
As we have the same rule, I will add a rule for this.

manfred-brands:
Created PR #265 

I do allow for `protected` methods as we have base classes defining `protected virtual` methods.
CS6028 already fires for adding a `protected` method in a `sealed` class so I didn't test for that here.

mikkelbu:
Resolved by #265

PR #265, opened 2020-07-22
NonTestMethodAccessibilityLevelAnalyzer: #147: Should be private

manfred-brands:
Only allow private and protected methods.

### DIFF
--- a/documentation/NUnit1026.md
+++ b/documentation/NUnit1026.md
@@ -1,6 +1,6 @@
 # NUnit1026
 
-## The test method is not public.
+## The test or setup/teardown method is not public.
 
 | Topic    | Value
 | :--      | :--
@@ -12,7 +12,7 @@
 
 ## Description
 
-The test method is not public.
+The test or setup/teardown method is not public.
 
 ## Motivation
 
@@ -23,10 +23,18 @@ To prevent tests that will fail at runtime, as NUnit only runs public test metho
 ### Example Violation
 
 ```csharp
+private int Value;
+
+[SetUp]
+void NUnit1026SetUp()
+{
+    Value = 42;
+}
+
 [Test]
 void NUnit1026SampleTest()
 {
-    Assert.Pass();
+    Assert.That(Value, Is.GreaterThan(0));
 }
 
 [TestCase(1)]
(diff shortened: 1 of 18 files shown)


Dreamescaper:
Methods marked with SetUp, Teardown, OneTimeSetUp, OneTimeTeardown attributes should be allowed as well.

manfred-brands:
@Dreamescaper Good catch. The reverse hold for the TestMethodAccessibilityLevelAnalyzer (NUnit1026).
I updated the both TestMethodAccessibilityLevelAnalyzer to regoznize methods with (OneTime)SetUp/TearDown attribute as being 'TestMethod'. The pre-existing rule will now fire if these methods are non-public.
@mikkelbu I have rebased the branch on top of master to prevent nunit failures on the documentation.

mikkelbu:
Good catch @Dreamescaper. Regarding TestMethodAccessibilityLevelAnalyzer then Setup/TearDown methods should either be public or protected (don't let the current wrong error message from the framework confuse you 😄 - this is corrected in https://github.com/nunit/nunit/commit/50babd1065afb63b58668f61449967d522bd9de4#diff-2db1dd498edafd51b9c83f1e99f5e0d5), so we need to handle Setup/TearDown a bit differently than the test methods.

manfred-brands:
@mikkelbu Code updated. It now matches the `methodInfo.IsPublic || methodInfo.IsFamily` from NUnit.
With all the SemanticModel lookups for attributes I was thinking of converting the rule to a SymbolAnalyzer as mentioned in #270, but without any benchmark comparison I wouldn't know if it is any better. I verified that `GetTypeInfo` is cached in the roslyn code so it is might not be so bad.

The documentation says that one can configure the rule or suppress it, but that doesn't make sense as any violation means those tests will not run (properly). However the NUnit Documentation tests will fail if that section is changed even adding `This rule should not be disabled and any violations must be fixed as otherwise the tests do not get run at all.` fails the validation.

If you are happy, it can be merged.

mikkelbu:
Thanks for the PR @manfred-brands. Currently, I don't mind having the sections about how to disable the diagnostic, but perhaps later we can improve on this. All diagnostics that are Errors will make tests not run - or fail in other manners.
Ps. I'll think about benchmarks for the analyzers.

PR #269, opened 2020-07-27
Assume that custom type converters can convert from a string value

### LINKED ISSUE #197: Handle custom type converters
We should be able to handle custom type converters. 

Suggestions from Gitter: https://gitter.im/GuOrg/Gu.Roslyn.Asserts?at=5e7935aa2e334e1b7aefffa0
> One heuristic you could use is to assume that the presence of a type converter for a type means that the type converter can convert from string. That is not a guarantee but it's such a mainline scenario.
...
So that would be a search in the custom attributes on the `targetType` ITypeSymbol
For a custom attribute with whatever the full type name of TypeConverterAttribute is
And that's when `argumentType` is System.String

jnm2:
Fixes #197. I do recommend reviewing by commit.

I also assumed that `(string)null` would be able to be converted, but I don't have any real-world cases in mind at the moment. I'm open to skipping that part.


src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -239,6 +239,85 @@ class AnalyzeArgumentHasImplicitConversion
...
+
+    class AnalyzeArgumentIsStringConvertedToEnum
+    {
+        [TestCase(""A"")]
+        public void Test(CustomType p) { }
+    }
+
+    [TypeConverter(typeof(CustomTypeConverter))]
+    struct CustomType { }
+    class CustomTypeConverter : TypeConverter { }");
```
jnm2, line 256: I saw the other tests had this indentation. Is this what you would still do even though there's a using statement? Is indenting something I should consider doing elsewhere?

src/nunit.analyzers/Extensions/AttributeArgumentSyntaxExtensions.cs
```diff
@@ -179,33 +178,41 @@ private static bool HasBuiltInImplicitConversion(ITypeSymbol argumentType, IType
...
-            if (targetType == default)
+            if (targetType != default)
             {
-                return false;
-            }
+                if (argumentValue is object)
+                {
+                    if (targetType.typeConverter == null)
+                    {
+                        return argumentValue.GetType() == typeof(string);
```
jnm2, line 192: (Unchanged code) Since `string` is a sealed type, this is the same as `argumentValue is string`.

src/nunit.analyzers/Extensions/INamespaceOrTypeSymbolExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+using Microsoft.CodeAnalysis;
+
+namespace NUnit.Analyzers.Extensions
+{
+    internal static class INamespaceOrTypeSymbolExtensions
+    {
+        /// <summary>
+        /// Checks the full name of a type or namespace without incurring the expense of building a full name string.
+        /// </summary>
+        public static bool HasFullName(this INamespaceOrTypeSymbol symbol, params string[] segments)
```
Dreamescaper, line 12: Is possible to make it to accept string, like `"System.ComponentModel.TypeConverterAttribute"`? It would be easier to re-use it across solution, I think.

jnm2: I compared the Name properties rather than building a full name string because that's how Roslyn does it. If we start using Substring or Split, those allocations are just as bad and I'd rather just shoot for using the simplest code than implement that (maybe even Compilation.GetTypeByMetadataName).

Dreamescaper: Sure, but I think it should possible to make it without Substring/Split and allocation-free (via Spans or direct char comparisons).

(I thought about using `Compilation.GetTypeByMetadataName` in such cases, but wasn't sure whether it is better, looked how it is implemented in roslyn-analyzers [here](https://github.com/dotnet/roslyn-analyzers/blob/master/src/Utilities/Compiler/WellKnownTypeProvider.cs), and, well, it's not trivial there.)

jnm2: GetTypeByMetadataName is a little different. It doesn't recognize attributes that are declared internally inside assemblies. But this is really not a concern with TypeConverterAttribute since the BCL has provided this since the beginning.

Allocation-free is possible if the analyzer references System.Memory which runs slow span on .NET Framework (and VS will run the analyzer on .NET Framework). But it's also a lot of work. What's the motivation?

> (I thought about using Compilation.GetTypeByMetadataName in such cases, but wasn't sure whether it is better, looked how it is implemented in roslyn-analyzers here, and, well, it's not trivial there.)

This doesn't look like how it is used in Roslyn diagnostics. It's pretty much three ops: GetTypeByMetadataName, bail if null, otherwise use SymbolEqualityComparer.Default.Equals to compare the returned symbol to each attribute class.

jnm2: (If we start implementing alloc-free splitting, we should probably add tests. I'm not opposed but also I want to understand the gain versus replacing the whole thing with GetTypeByMetadataName.)

Dreamescaper: > System.Memory which runs slow span on .NET Framework

Didn't know that, thanks! 
Still possible with direct chars comparison, but yeah, won't be pretty.

> What's the motivation?

We have lots of full name checks implemented via `GetFullMetadataName`. I'd like to be able to replace those with your method where possible.

src/nunit.analyzers/Extensions/INamespaceOrTypeSymbolExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            if (segments.Length < 1)
+                throw new ArgumentException("At least one name segment must be specified.", nameof(segments));
+
+            if (segments.Any(string.IsNullOrEmpty))
```
Dreamescaper, line 17: This check does allocate, right? (because of enumerator)

jnm2: Yes, and the `params` in the method signature allocates. I was actively avoiding string building more than allocations in general, and even more specifically thinking of this method as a helper to automate doing the pattern I see in Roslyn:

```cs
var isTypeConverterAttribute = type is
{
    Name: "TypeConverterAttribute",
    ContainingSymbol: INamespaceSymbol
    {
        Name: "ComponentModel",
        ContainingNamespace:
        {
            Name: "System",
            ContainingNamespace: { IsGlobalNamespace = true },
        }
    },
};
```

src/nunit.analyzers/Extensions/INamespaceOrTypeSymbolExtensions.cs
Dreamescaper: > If we start implementing alloc-free splitting

I was thinking more about "inline" comparison, not splitting, smth like
```csharp
            int index = fullName.Length - 1;
            if (!SegmentEquals(fullName, symbol.Name, ref index)) return false;
            
            ... (other segments checks)

            static bool SegmentEquals(string fullName, string segmentName, ref int index)
            {
                if (segmentName.Length > index + 1) return false;

                for (var i = segmentName.Length - 1; i >= 0; i--)
                    if (fullName[index--] != segmentName[i]) return false;

                return index < 0 || fullName[index--] == '.';
            }
```
But I agree that it's ugly.

>  the gain versus replacing the whole thing with GetTypeByMetadataName.

The only advantage I can think of is that we don't need `Compilation` for that to work, but that's probably not worth it. Less ugly code the better :)

jnm2: Should I go ahead and refactor this to use GetTypeByMetadata then?

Dreamescaper: I think that would be great, thanks!

jnm2: Updated.

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
mikkelbu: I think it is fine as it is.

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -239,6 +239,85 @@ class AnalyzeArgumentHasImplicitConversion
             AnalyzerAssert.Valid<TestCaseUsageAnalyzer>(testCode);
         }
 
+        [Test]
+        public void AnalyzeArgumentIsStringConvertedToTypeWithCustomConverter()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    using System.ComponentModel;
+
+    class AnalyzeArgumentIsStringConvertedToEnum
```
mikkelbu, line 248:
```suggestion
    class AnalyzeArgumentIsStringConvertedToTypeWithCustomConverter
```

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -239,6 +239,85 @@ class AnalyzeArgumentHasImplicitConversion
...
+            AnalyzerAssert.Valid<TestCaseUsageAnalyzer>(testCode);
+        }
+
+        [Test]
+        public void AnalyzeArgumentIsNullConvertedToTypeWithCustomConverter()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    using System.ComponentModel;
+
+    class AnalyzeArgumentIsStringConvertedToEnum
```
mikkelbu, line 267:
```suggestion
    class AnalyzeArgumentIsNullConvertedToTypeWithCustomConverter
```

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -239,6 +239,85 @@ class AnalyzeArgumentHasImplicitConversion
...
+            AnalyzerAssert.Valid<TestCaseUsageAnalyzer>(testCode);
+        }
+
+        [Test]
+        public void AnalyzeArgumentIsStringConvertedToTypeWithCustomConverterOnBaseType()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    using System.ComponentModel;
+
+    class AnalyzeArgumentIsStringConvertedToEnum
```
mikkelbu, line 286:
```suggestion
    class AnalyzeArgumentIsStringConvertedToTypeWithCustomConverterOnBaseType
```

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -239,6 +239,85 @@ class AnalyzeArgumentHasImplicitConversion
...
+            AnalyzerAssert.Valid<TestCaseUsageAnalyzer>(testCode);
+        }
+
+        [Test]
+        public void AnalyzeArgumentIsIntConvertedToTypeWithCustomConverter()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    using System.ComponentModel;
+
+    class AnalyzeArgumentIsStringConvertedToEnum
```
mikkelbu, line 306:
```suggestion
    class AnalyzeArgumentIsIntConvertedToTypeWithCustomConverter
```

jnm2: Oh, thanks. Is it Okay if I call it `C` like I usually do?

mikkelbu: Sure. When reviewing this I was also consider what we gained from this (except for additional work 😄).

jnm2:
Sure, it was fun!

PR #278, opened 2020-08-08
Analyzer for SameAs when used on Value Types.

### LINKED ISSUE #276: SameAs should warn if used with value types
Motivation: https://github.com/nunit/nunit/issues/3614#issuecomment-669962745

If either the expected value or actual value or both are a value-typed expression, `Assert.AreSame` or `Is.SameAs` are guaranteed to fail because C# will implicitly convert the value type to a new boxed instance. It doesn't make sense to use AreSame/SameAs unless the types of the passed expressions are both reference types.

It would make sense to offer a fix to change SameAs to EqualTo since that was probably the intent.

Should show an error or warning on `SameAs`:

```cs
class C
{
    void M()
    {
        var x = 3;
        Assert.That(x, Is.SameAs(3));

        // Also this
        Assert.That((object)x, Is.SameAs(3));

        // Also this
        Assert.That(x, Is.SameAs((object)3));
    }
}
```

Should not show an error or warning because the expressions are both reference-typed:

```cs
class C
{
    void M()
    {
        var x = 3;
        Assert.That((object)x, Is.SameAs((object)3));
    }
}
```

manfred-brands:
CodeFix to replace SameAs with IsEqualTo
Fixes #276

### DIFF
--- a/documentation/NUnit2040.md
+++ b/documentation/NUnit2040.md
@@ -1,0 +1,67 @@
+# NUnit2040
+
+## Non-reference types for SameAs constraint.
+
+| Topic    | Value
+| :--      | :--
+| Id       | NUnit2040
+| Severity | Error
+| Enabled  | True
+| Category | Assertion
+| Code     | [SameAsOnValueTypesAnalyzer](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/SameAsOnValueTypes/SameAsOnValueTypesAnalyzer.cs)
+
+## Description
+
+The SameAs constraint always fails on value types as the actual and the expected value cannot be the same reference.
+
+## Motivation
+
+```csharp
+var expected = Guid.Empty;
+var actual = expected;
+
+Assert.That(actual, Is.SameAs(expected));
+```
+
+As `Guid` is a `struct`, actual will be a copy of expected but not have the same reference.
+
+## How to fix violations
+
+Replace `Is.SameAs` with `Is.EqualTo`.
+
+```csharp
+var expected = Guid.Empty;
+var actual = expected;
+
+Assert.That(actual, Is.EqualTo(expected));
+```
+
+<!-- start generated config severity -->
+## Configure severity
+
+### Via ruleset file.
+
+Configure the severity per project, for more info see [MSDN](https://msdn.microsoft.com/en-us/library/dd264949.aspx).
+
+### Via #pragma directive.
+
+```csharp
+#pragma warning disable NUnit2040 // Non-reference types for SameAs constraint.
+Code violating the rule here
+#pragma warning restore NUnit2040 // Non-reference types for SameAs constraint.
+```
+
+Or put this at the top of the file to disable all instances.
+
+```csharp
+#pragma warning disable NUnit2040 // Non-reference types for SameAs constraint.
+```
+
+### Via attribute `[SuppressMessage]`.
+
+```csharp
+[System.Diagnostics.CodeAnalysis.SuppressMessage("Assertion",
+    "NUnit2040:Non-reference types for SameAs constraint.",
+    Justification = "Reason...")]
+```
+<!-- end generated config severity -->
--- a/src/nunit.analyzers.tests/SameAsOnValueTypes/SameAsOnValueTypesAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/SameAsOnValueTypes/SameAsOnValueTypesAnalyzerTests.cs
@@ -93,5 +93,53 @@ namespace NUnit.Analyzers.Tests.SameAsOnValueTypes
 
             AnalyzerAssert.Valid(analyzer, testCode);
         }
+
+        [Test]
+        public void AnalyzeClassicWhenLiteralTypesProvided()
+        {
+            var testCode = TestUtility.WrapInTestMethod(
+                "↓Assert.AreSame(1, 1);");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeClassicWhenStructTypesProvided()
+        {
+            var testCode = TestUtility.WrapInTestMethod(@"
+                var expected = Guid.Empty;
+                var actual = expected;
+
+                ↓Assert.AreSame(expected, actual);");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeClassicWhenActualIsReferenceTypeAndExpectedIsValueType()
+        {
+            var testCode = TestUtility.WrapInTestMethod(
+                @"↓Assert.AreNotSame(3, ""3"");");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeClassicWhenActualIsValueTypeAndExpectedIsReferenceType()
+        {
+            var testCode = TestUtility.WrapInTestMethod(
+                @"↓Assert.AreNotSame(""3"", 3);");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeClassicWhenBothArgumentsAreReferenceType()
+        {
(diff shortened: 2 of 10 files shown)


src/nunit.analyzers.tests/SameAsOnValueTypes/SameAsOnValueTypesAnalyzerTests.cs
```diff
@@ -0,0 +1,106 @@
...
+        public void AnalyzeWhenOtherArgumentIsCastToObject()
+        {
+            var testCode = TestUtility.WrapInTestMethod(
+                "↓Assert.That(Guid.Empty, Is.Not.SameAs((object)Guid.Empty));");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenBothArgumentsAreCastToObject()
```
jnm2, line 77: (Specifically the salient point is that both arguments are reference-typed, not that there is a cast in the expression.)

documentation/NUnit2040.md
```diff
@@ -0,0 +1,67 @@
...
+
+The SameAs constraint always fails on value types as the actual and the expected value cannot be the same reference.
+
+## Motivation
+
+```csharp
+var expected = Guid.Empty;
+var actual = expected;
+
+↓Assert.That(actual, Is.SameAs(expected));
```
mikkelbu, line 23:
```suggestion
Assert.That(actual, Is.SameAs(expected));
```

documentation/NUnit2040.md
```diff
@@ -0,0 +1,67 @@
...
+
+## How to fix violations
+
+Replace `Is.SameAs` with `Is.EqualTo`.
+
+```csharp
+var expected = Guid.Empty;
+var actual = expected;
+
+↓Assert.That(actual, Is.EqualTo(expected));
```
mikkelbu, line 36:
```suggestion
Assert.That(actual, Is.EqualTo(expected));
```

src/nunit.analyzers/SameAsOnValueTypes/SameAsOnValueTypesCodeFix.cs
```diff
@@ -0,0 +1,60 @@
...
+            if (invocationNode == null)
+                return;
+
+            // Find the 'SameAs' member access expression
+            var constraintArgument = invocationNode.ArgumentList.Arguments[1];
+            var isSameExpression = constraintArgument.DescendantNodesAndSelf()
+                                                           .Where(s => s.IsKind(SyntaxKind.SimpleMemberAccessExpression))
+                                                           .OfType<MemberAccessExpressionSyntax>()
+                                                           .Where(m => m.Name.ToString() == NunitFrameworkConstants.NameOfIsSameAs)
+                                                           .SingleOrDefault();
```
mikkelbu, line 45:
```suggestion
            var isSameExpression = constraintArgument
                .DescendantNodesAndSelf()
                .Where(s => s.IsKind(SyntaxKind.SimpleMemberAccessExpression))
                .OfType<MemberAccessExpressionSyntax>()
                .SingleOrDefault(m => m.Name.ToString() == NunitFrameworkConstants.NameOfIsSameAs);
```

src/nunit.analyzers/SameAsOnValueTypes/SameAsOnValueTypesCodeFix.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            var isEqualToExpression = isSameExpression.WithName(SyntaxFactory.IdentifierName(NunitFrameworkConstants.NameOfIsEqualTo));
```
mikkelbu, line 47: Perhaps we should have a return in case `isSameExpression` is null

src/nunit.analyzers/SameAsOnValueTypes/SameAsOnValueTypesAnalyzer.cs
```diff
@@ -0,0 +1,92 @@
...
+            // Currently only 'Not' suffix supported, as all other suffixes change actual type for constraint
+            // (e.g. All, Some, Property, Count, etc.)
+
+            return constraintPartExpression.GetPrefixesNames().Any(s => s != NunitFrameworkConstants.NameOfIsNot);
+        }
+
+        private static bool HasCustomEqualityComparer(ConstraintPartExpression constraintPartExpression)
+        {
+            return constraintPartExpression.GetSuffixesNames().Any(s => s == NunitFrameworkConstants.NameOfUsing);
+        }
```
mikkelbu, line 90: Not used
```suggestion
```

src/nunit.analyzers/SameAsOnValueTypes/SameAsOnValueTypesCodeFix.cs
manfred-brands: Done. 
However should this fix not only be called for the supported diagnostic, which would mean there should always be an `SameAs`?  Maybe it is better to be more conservative than throw.

manfred-brands:
Thanks @mikkelbu for the review. Review changes applied and squashed. Ready for merge.

jnm2:
What do you think about handling `Assert.AreSame`? Should the only diagnostic there be recommending that you switch to using the constraint model, and only then you see the problem that you're using value types?

manfred-brands:
@jnm2 I did think of that and assumed people would convert first. But what if they disable the Classic To Constraint rules then they would get no warning. 
The same holds for `SameAsIncompatibleTypesAnalyzerTests` which also only works on the Constraint Model.
But `ConstActualValueUsageAnalyzer` works on both.

manfred-brands:
@mikkelbu I changed the Identifier from NUnit2040 into NUnit1029. The 2k range is used for ClassicAssert conversions.

manfred-brands:
Updated the branch to support the Classic Assert.AreSame and Assert.AreNotSame methods including matching code fix to convert to AreEqual and AreNotEqual respectively.

Dreamescaper:
> @mikkelbu I changed the Identifier from NUnit2040 into NUnit1029. The 2k range is used for ClassicAssert conversions.

As far as I remember 1k range is used for Structure analyzers, and 2k - for Assertion analyzers.
I think that 2k is more applicable here.

manfred-brands:
I see now that the SameAsIncompatibleTypesAnalyzer uses 2020.  Reverted the commit.

mikkelbu:
@Dreamescaper Is right (I plan to make this more clear in the TOC by having two separate tables.

src/nunit.analyzers/SameAsOnValueTypes/SameAsOnValueTypesCodeFix.cs
mikkelbu: I agree that this should never fail, but I plan to enable more strict checks of nullable reference types, and then we will need to handle the case, as the compiler cannot tell that this cannot be null.

manfred-brands: Will you upgrade to a later version of Microsoft.CodeAnalysis which has nullable attributes? When I tried, it requires me to target netstandard2.0 and after that I got various nullability compile errors (CS) and analyzer violations (RS).

manfred-brands:
@mikkelbu Branch updates. Ready for merge again.

src/nunit.analyzers/SameAsOnValueTypes/SameAsOnValueTypesCodeFix.cs
mikkelbu: I'm considering trying to multi-target, and see if I can make it work, but currently I don't have enough sparetime to do much besides reviewing and screening issues in the various NUnit projects.

So something similar to https://www.meziantou.net/how-to-use-nullable-reference-types-in-dotnet-standard-2-0-and-dotnet-.htm

mikkelbu:
Thanks for the work @manfred-brands, I'll merge it now.

src/nunit.analyzers/SameAsOnValueTypes/SameAsOnValueTypesCodeFix.cs
manfred-brands: We added multi-targeting to our work code. Create an issue with what targets you want and assign it to me.

Issue #280, opened 2020-08-11
Consider overload resolution when transforming to comparison constraints


chm-tm:
Consider the struct `MyFloat`:

```C#
struct MyFloat
{
    readonly float _value;

    public MyFloat(float value) => _value = value;

    public static implicit operator float(MyFloat value) => value._value;
    public static implicit operator MyFloat(float value) => new MyFloat(value);
}
```
Now consider the following test code:
```C#
float x = 1;
MyFloat y = 2;
Assert.Less(x, y);
Assert.That(x, Is.LessThan(y)); // code fix of NUnit2029 for above statement
```
The first Assert succeeds, but the second Assert fails with the following message:
> System.ArgumentException : Object must be of type Single.

While this is a good indication for `MyFloat` to add some functionality, like to implement `IComparable`, the second Assert is semantically different to the first one.

The expected transform is this one:
```C#
Assert.That(x, Is.LessThan((float)y));
```
That is, the code fixes need to take the classic model overload into account.

manfred-brands:
On the other hand you could say that the Is.LessThan and Is.EqualTo, etc. could utilise the implicit cast.
There is a ticket https://github.com/nunit/nunit/issues/2891 for types that implement IConvertible but it was closed without a fix.

But a code fix should not cause tests to fail.
I will see if the conversion can find out the actual overload of the classic method to account for implicit cast and turning it into an explicit one.

chm-tm:
I imagine this could be quite a big amount of work to get this right everywhere. It could also be rendered superfluous if the framework implementation changes at a later time.
An alternative approach would be to add a note about possible semantic differences to the _How to fix violations_ section of each of the respective analyzers.

manfred-brands:
If you use `Assert.That(x, Is.EqualTo(y));` then NUnit2021 fires complaining the types are not compatible. I'm actually working on #279 which would then cause this to fire for Is.Less etc as well.

manfred-brands:
OK, I managed to create a generic solution.

PR #283, opened 2020-08-11
ClassicModelAssertUsageCodeFix: #280: Cast type to implicit type

manfred-brands:
When a specific classic assert overload is called.

### DIFF
--- a/src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
+++ b/src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
@@ -4,6 +4,7 @@ using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using NUnit.Analyzers.Constants;
+using NUnit.Analyzers.Extensions;
 using static NUnit.Analyzers.Constants.NunitFrameworkConstants;
 
 namespace NUnit.Analyzers.ClassicModelAssertUsage
@@ -203,7 +204,7 @@ namespace NUnit.Analyzers.ClassicModelAssertUsage
             defaultSeverity: DiagnosticSeverity.Info,
             description: ClassicModelUsageAnalyzerConstants.IsNotInstanceOfDescription);
 
-        private static readonly ImmutableDictionary<string, DiagnosticDescriptor> name =
+        private static readonly ImmutableDictionary<string, DiagnosticDescriptor> NameToDescriptor =
           new Dictionary<string, DiagnosticDescriptor>
           {
               { NameOfAssertIsTrue, isTrueDescriptor },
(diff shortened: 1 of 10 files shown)


src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
```diff
@@ -232,21 +233,58 @@ public sealed class ClassicModelAssertUsageAnalyzer : BaseAssertionAnalyzer
...
+                    if (implicitType == "System.Collections.ICollection")
+                    {
+                        return string.Empty;
+                    }
+
+                    return implicitType;
+                }
+
+                var arguments = assertExpression.ArgumentList.Arguments;
+                string firstArgumentImplicitTypeConversion = GetImplicitTypeConversion(arguments[0].Expression);
```
Dreamescaper, line 276: I think it should go to CodeFix itself, analyzer doesn't care about this information.

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
```diff
@@ -232,21 +233,58 @@ public sealed class ClassicModelAssertUsageAnalyzer : BaseAssertionAnalyzer
...
+                    var typeInfo = context.SemanticModel.GetTypeInfo(expression, context.CancellationToken);
+
+                    var convertedType = typeInfo.ConvertedType;
+
+                    if (convertedType == typeInfo.Type)
+                    {
+                        return string.Empty;
+                    }
+
+                    if (convertedType.SpecialType == SpecialType.System_Object ||
```
Dreamescaper, line 259: What is the logic here? Why object, IEnumerable and ICollection?

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
manfred-brands: Agreed. However the CodeFix has no access to the semantic model to access the necessary type information. Or do you know another way?

Dreamescaper: I suppose you can use `context.Document.GetSemanticModelAsync()`

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
manfred-brands: object is the parameter type for Assert.That constraints and we don't need to cast to that.

ICollection is the parameter type for Assert.Contains.
IEnumerable is the fallback parameter type for Assert.Is(Not)Empty.

These are to prevent that `Assert.IsEmpty(myCollection)` becomes `Assert.That((IEnumerable)myCollection, Is.Empty)`.
and `Assert.Contains(myRecord, myCollection)` becomes `Assert.That((ICollection)myCollection, Does.Contain((object)myRecord)`

I didn't want to restrict it to only do the conversion logic for the comparable set of asserts: `Less`, `LessOrEqual`, `Greater`, `GreaterOrEqual` as I noticed that `IsTrue` and `IsFalse`  also need an explicit cast in the constraint model. Similar for `IsEmpty` and `IsNotEmpty` when the `string` overload is called.

I added more test cases for the 4 other asserts needing casts.

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
manfred-brands: Thanks. Code moved from Analyzer to CoeFix.

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageAnalyzer.cs
Dreamescaper: I would suggest to get `compilation.ClassifyConversion(typeInfo.Type, convertedType)`, and checking for `IsUserDefined` property.
Would it work here?

manfred-brands: Thanks @Dreamescaper 
I tried `semanticModel.ClassifyConversion(expression, convertedType)` and it works! 
Allows me to remove these akward exclusions.

Dreamescaper:
Btw, use Fixes keyword to close the bug when PR is merged
https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords#closing-an-issue-in-the-same-repository

manfred-brands:
@mikkelbu Squashed and with the right keyword in the commit message it should close the issue when you merge.

mikkelbu:
Thanks @manfred-brands. I'll merge it now.

mikkelbu:
Ps. I think it is better to put the "Fixes ..." in the first message in the PR, then it will be more visible in the UI of the PR  
image
But I can also just add this information manually from the UI.

This will also effect the last message on the issue: `mikkelbu closed this in 1dd5089 2 minutes ago` (when it is a commit that closes the issue) vs. `mikkelbu closed this in #278 5 days ago` (when it is the PR).

Issue #218, opened 2020-05-01
Rework ConstraintExpression to better support operators precendence.


Dreamescaper:

mikkelbu:
Some more information. I found the following comment in nunit/nunit#2524
> The key difference is that Property is one of those syntactic elements that actually changes what the following constraint acts on- i.e. to the value of the property. But that only applies to the immediately following constraint. As soon as Or is seen, we go back to using the original actual.

Dreamescaper:
That's probably true for Property constraint, but not for all of them (e.g. All operator has higher precence than Or or And operators).

To resolve this completely we probably need to store precedence values for each analyzer, and resolve applied operators for each constraint part based on precedence.
But that doesn't sound easy.

mikkelbu:
I very much agree that it does not sound easy and I also think/guess that most people don't use very complicated constraints that often.

Personally, it is very rare that I use anything more fancy than `Is.Not.Null.And.Not.Empty`, as I tend to prefer multiple (individual) constraints that are easier to read and understand.

Dreamescaper:
Yeah, agree.
Still we should not allow false positives.
Currently we handle such cases on each analyzer separately, maybe makes sense to make it centralized (e.g. in some base class).

PR #284, opened 2020-08-14
Add CompatibleType checking to Comparable constraints

### LINKED ISSUE #279: Augment SameAsIncompatibleTypes and EqualToIncompatibleTypes to apply for more cases
The `Assert.That` constraints work on type `object` and the compiler doesn't detect comparing non-matching types.
The rules `SameAsIncompatibleTypes` and `EqualToIncompatibleTypes` made a good start on this but there are many other asserts.

This ticket is to augment these rules to:
* Work on more asserts, such as `Is.GreaterThanOrEquals` etc.
* Work on classic asserts, like `Assert.AreEqual`

The compiler is also happy to allow:

    object instance1 = new object();
    object instance2 = new object();
    Assert.That(instance1, Is.LessThan(instance2));

but it does not allow:

    Assert.Less(instance1, instance2);

Augment the incompatible type check to verify that the types implement `IComparable` where needed or add a new Comparable analyzer.

manfred-brands:
Fixes #279

### DIFF
--- a/src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
@@ -26,6 +26,15 @@ namespace NUnit.Analyzers.Tests.EqualToIncompatibleTypes
         }
 
         [Test]
+        public void AnalyzeClassicWhenIncompatibleTypesProvided()
+        {
+            var testCode = TestUtility.WrapInTestMethod(
+                "Assert.AreEqual(↓\"1\", 1);");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
         public void AnalyzeWhenCollectionsWithIncompatibleValueTypesProvided()
         {
             var testCode = TestUtility.WrapInTestMethod(@"
@@ -36,7 +45,6 @@ namespace NUnit.Analyzers.Tests.EqualToIncompatibleTypes
 
             AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
         }
-
 
         [Test]
         public void AnalyzeWhenDictionariesWithIncompatibleValueTypesProvided()
@@ -215,6 +223,27 @@ namespace NUnit.Analyzers.Tests.EqualToIncompatibleTypes
             var actual = new A();
             var expected = new B();
             Assert.That(actual, Is.Not.EqualTo(↓expected));
+        }
+    }");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeClassicWhenIncompatibleTypesProvided_WithNegatedAssert()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    class A { }
+    class B { }
+
+    public class Tests
+    {
+        [Test]
+        public void AnalyzeWhenIncompatibleTypesProvided_WithNegatedAssert()
+        {
+            var actual = new A();
+            var expected = new B();
+            Assert.AreNotEqual(↓expected, actual);
         }
     }");
 
@@ -840,6 +869,15 @@ namespace NUnit.Analyzers.Tests.EqualToIncompatibleTypes
                 Assert.That(task1, Is.SameAs(task2));");
 
             AnalyzerAssert.Valid(analyzer, testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenMoreThanOneConstraintExpressionIsUsed()
+        {
+            var testCode = TestUtility.WrapInTestMethod($@"
+                Assert.That(5, Is.Not.EqualTo(4) & Is.Not.EqualTo(↓""6""));");
+
+            AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
         }
     }
 }
--- a/src/nunit.analyzers/ComparableTypes/ComparableTypesAnalyzer.cs
+++ b/src/nunit.analyzers/ComparableTypes/ComparableTypesAnalyzer.cs
@@ -1,0 +1,139 @@
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Operations;
+using NUnit.Analyzers.Constants;
+using NUnit.Analyzers.Extensions;
+using NUnit.Analyzers.Helpers;
+using NUnit.Analyzers.Operations;
+
+namespace NUnit.Analyzers.ComparableTypes
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class ComparableTypesAnalyzer : BaseAssertionAnalyzer
+    {
+        private static readonly DiagnosticDescriptor descriptor = DiagnosticDescriptorCreator.Create(
+            id: AnalyzerIdentifiers.ComparableTypes,
+            title: ComparableConstants.Title,
+            messageFormat: ComparableConstants.Message,
+            category: Categories.Assertion,
+            defaultSeverity: DiagnosticSeverity.Warning,
+            description: ComparableConstants.Description);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(descriptor);
+
+        private static readonly ImmutableHashSet<string> SupportedConstraints = ImmutableHashSet.Create(
+            NunitFrameworkConstants.NameOfIsLessThan,
+            NunitFrameworkConstants.NameOfIsLessThanOrEqualTo,
+            NunitFrameworkConstants.NameOfIsGreaterThan,
+            NunitFrameworkConstants.NameOfIsGreaterThanOrEqualTo);
+
+        protected override void AnalyzeAssertInvocation(OperationAnalysisContext context, IInvocationOperation assertOperation)
+        {
+            if (!AssertHelper.TryGetActualAndConstraintOperations(assertOperation,
+                out var actualOperation, out var constraintExpression))
+            {
+                return;
+            }
+
+            if (actualOperation == null)
+                return;
+
+            var actualType = AssertHelper.GetUnwrappedActualType(actualOperation);
+
+            if (actualType == null)
+                return;
+
+            foreach (var constraintPartExpression in constraintExpression.ConstraintParts)
+            {
+                if (HasIncompatiblePrefixes(constraintPartExpression)
+                    || HasCustomComparer(constraintPartExpression)
+                    || constraintPartExpression.HasUnknownExpressions())
+                {
+                    continue;
+                }
+
+                var constraintMethod = constraintPartExpression.GetConstraintMethod();
+                if (constraintMethod == null)
+                    continue;
+
+                if (!SupportedConstraints.Contains(constraintMethod.Name))
+                    continue;
+
+                var expectedOperation = constraintPartExpression.GetExpectedArgument();
+                if (expectedOperation == null)
+                    continue;
+
+                var expectedType = expectedOperation.Type;
+                if (expectedType == null)
+                    continue;
+
+                if (!CanCompare(actualType, expectedType, context.Compilation))
+                {
+                    context.ReportDiagnostic(Diagnostic.Create(
+                        descriptor,
+                        expectedOperation.Syntax.GetLocation(),
+                        constraintMethod.Name));
+                }
+            }
+        }
+
+        private static bool CanCompare(ITypeSymbol actualType, ITypeSymbol expectedType, Compilation compilation)
+        {
+            if (actualType.OriginalDefinition.SpecialType == SpecialType.System_Nullable_T)
+                actualType = ((INamedTypeSymbol)actualType).TypeArguments[0];
+
+            if (expectedType.OriginalDefinition.SpecialType == SpecialType.System_Nullable_T)
+                expectedType = ((INamedTypeSymbol)expectedType).TypeArguments[0];
+
+            var conversion = compilation.ClassifyConversion(actualType, expectedType);
+            if (conversion.IsNumeric)
+                return true;
+
+            if (IsIComparable(actualType, expectedType) || IsIComparable(expectedType, actualType))
+                return true;
+
+            // NUnit doesn't demand that IComparable is for the same type.
+            // But MS does: https://docs.microsoft.com/en-us/dotnet/api/system.icomparable.compareto?view=netcore-3.1
+            if (actualType == expectedType && IsIComparable(actualType))
+                return true;
+
+            return false;
+        }
+
+        private static bool IsIComparable(ITypeSymbol typeSymbol, ITypeSymbol comparableTypeArguments)
+        {
+            if (typeSymbol.AllInterfaces.Any(i => i.TypeArguments.Length == 1
+                && i.TypeArguments[0].Equals(comparableTypeArguments)
+                && i.GetFullMetadataName() == "System.IComparable`1"))
+            {
+                return true;
+            }
+
+            // NUnit allows for an CompareTo method, even if not implementing IComparable.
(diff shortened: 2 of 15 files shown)


Dreamescaper:
What are those changes in EqualTo and SameAs analyzers?

manfred-brands:
Add support for classic model.
Currently there is no warning when you have: `Assert.AreEqual(new A(), new B());`
Only when converting this to the constraint model you get an error about the incompatible types.

manfred-brands:
Rebased and adjust to work with @Dreamescaper optimizations.

src/nunit.analyzers/ComparableTypes/ComparableTypesAnalyzer.cs
```diff
@@ -0,0 +1,125 @@
...
+            return false;
+        }
+
+        private static bool IsIComparable(ITypeSymbol typeSymbol, ITypeSymbol comparableTypeArguments)
+        {
+            return typeSymbol.AllInterfaces.Any(i => i.TypeArguments.Length == 1
+                && i.TypeArguments[0].Equals(comparableTypeArguments)
+                && i.GetFullMetadataName() == "System.IComparable`1");
+        }
+
```
Dreamescaper, line 110: Remove empty line

src/nunit.analyzers/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzer.cs
```diff
@@ -25,43 +25,59 @@ public class EqualToIncompatibleTypesAnalyzer : BaseAssertionAnalyzer
...
+
+                    var constraintMethod = constraintPartExpression.GetConstraintMethod();
+
+                    if (constraintMethod?.Name != NunitFrameworkConstants.NameOfIsEqualTo
+                        || constraintMethod.ReturnType?.GetFullMetadataName() != NunitFrameworkConstants.FullNameOfEqualToConstraint)
+                    {
+                        continue;
+                    }
+
+                    expectedOperation = constraintPartExpression.GetExpectedArgument();
```
Dreamescaper, line 62: same concerns as above - won't handle combined constraints

src/nunit.analyzers/SameAsIncompatibleTypes/SameAsIncompatibleTypesAnalyzer.cs
```diff
@@ -26,44 +27,73 @@ public class SameAsIncompatibleTypesAnalyzer : BaseAssertionAnalyzer
...
+                        continue;
+                    }
 
-                if (expectedType != null && !CanBeSameType(actualType, expectedType, context.Compilation))
-                {
-                    context.ReportDiagnostic(Diagnostic.Create(
-                        descriptor,
-                        expectedArgumentOperation!.Syntax.GetLocation()));
+                    expectedOperation = constraintPartExpression.GetExpectedArgument();
+                    break;
```
Dreamescaper, line 64: same

src/nunit.analyzers/ComparableTypes/ComparableTypesAnalyzer.cs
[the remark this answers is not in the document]
manfred-brands: If we enable StyleCop on this repository, it would have told me.

src/nunit.analyzers/ComparableTypes/ComparableTypesAnalyzer.cs
```diff
@@ -0,0 +1,125 @@
...
+                constraintMethod = constraintPartExpression.GetConstraintMethod();
+
+                if (constraintMethod == null)
+                    continue;
+
+                if (!SupportedConstraints.Contains(constraintMethod.Name))
+                    continue;
+
+                expectedOperation = constraintPartExpression.GetExpectedArgument();
+                break;
```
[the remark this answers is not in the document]
manfred-brands: Because I didn't know your example was possible, nor have I seen it before. This will also affect `SameAsOnValueTypesAnalyzer`. I will have to find a different way to deal with these.

manfred-brands:
> Two cases, that NUnit currently supports.
> 
> 1. Non-generic `IComparable`.
>    This assert passes.
> 
> ```cs
>         class A : IComparable
>         {
>             public int CompareTo(object obj) => 1;
>         }
> 
>         class B { }
> 
> 
>         Assert.That(new A(), Is.GreaterThan(new B()));
> ```
>
Should this pass? According to [IComparable](https://docs.microsoft.com/en-us/dotnet/api/system.icomparable.compareto?view=netcore-3.1)
> Compares the current instance with another object of the **same** type
Although nothing stops the implementation of  `IComparable`  to work on multiple types, it is more likely that it only supports the current type.
 
> 1. `CompareTo` method without any interface.
>    This assert passes.
> 
> ```cs
>         class A
>         {
>             public int CompareTo(B b) => 1;
>         }
> 
>         class B { }
> 
> 
>         Assert.That(new A(), Is.GreaterThan(new B()));
> ```

Should this pass? Why would someone create a method `CompareTo` but not claim `IComparable<B>`? Does NUnit only look for an `CompareTo` method and not the interface?

src/nunit.analyzers/ComparableTypes/ComparableTypesAnalyzer.cs
```diff
@@ -0,0 +1,125 @@
...
+                expectedType = ((INamedTypeSymbol)expectedType).TypeArguments[0];
+
+            var conversion = compilation.ClassifyConversion(actualType, expectedType);
+            if (conversion.IsNumeric)
+                return true;
+
+            if (IsIComparable(actualType, expectedType) || IsIComparable(expectedType, actualType))
+                return true;
+
+            if (actualType == expectedType && actualType.BaseType.SpecialType == SpecialType.System_Enum)
```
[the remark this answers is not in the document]
manfred-brands: Added support for IComparable when both actual and expected type are the same.

src/nunit.analyzers/ComparableTypes/ComparableTypesAnalyzer.cs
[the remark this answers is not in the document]
manfred-brands: Implemented now.

src/nunit.analyzers/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzer.cs
manfred-brands: Implemented now.

src/nunit.analyzers/SameAsIncompatibleTypes/SameAsIncompatibleTypesAnalyzer.cs
manfred-brands: Implemented now.

src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
```diff
@@ -841,5 +870,14 @@ public void NoDiagnosticWhenComparingTask2()
 
             AnalyzerAssert.Valid(analyzer, testCode);
         }
+
+        [Test]
+        public void AnalyzeWhenMoreThanOnConstraintExpressionIsUsed()
+        {
+            var testCode = TestUtility.WrapInTestMethod($@"
+                Assert.That(5, Is.Not.EqualTo(4) & Is.Not.EqualTo(""6""));");
```
Dreamescaper, line 878:
```suggestion
                Assert.That(5, Is.Not.EqualTo(4) & Is.Not.EqualTo(↓""6""));");
```

src/nunit.analyzers/ComparableTypes/ComparableTypesAnalyzer.cs
```diff
@@ -0,0 +1,153 @@
...
+                expectedType = ((INamedTypeSymbol)expectedType).TypeArguments[0];
+
+            var conversion = compilation.ClassifyConversion(actualType, expectedType);
+            if (conversion.IsNumeric)
+                return true;
+
+            if (IsIComparable(actualType, expectedType) || IsIComparable(expectedType, actualType))
+                return true;
+
+        /*
```
Dreamescaper, line 98: Remove commented code?

src/nunit.analyzers/ComparableTypes/ComparableTypesAnalyzer.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        private static bool IsIComparable(ITypeSymbol typeSymbol)
+        {
+            if (typeSymbol.AllInterfaces.Any(i => i.TypeArguments.Length == 0
+                && i.GetFullMetadataName() == "System.IComparable"))
+            {
+                return true;
+            }
+
+            // NUnit allows for an CompareTo method, even if not implementing IComparable.
```
Dreamescaper, line 135: Nope, this is not allowed. 
See https://github.com/nunit/nunit/blob/master/src/NUnitFramework/framework/Constraints/NUnitComparer.cs

Dreamescaper:
>Should this pass? Why would someone create a method CompareTo but not claim IComparable? Does NUnit only look for an CompareTo method and not the interface?

Maybe that's just a NUnitComparer's implementation detail, and it makes no sense to account for that.
@mikkelbu wdyt?

src/nunit.analyzers/ComparableTypes/ComparableTypesAnalyzer.cs
manfred-brands: My bad, I extrapolated your example to `CompareTo(object?)`

manfred-brands: Deleted and allows me to delete a unit tests which tested that.

src/nunit.analyzers/ComparableTypes/ComparableTypesAnalyzer.cs
manfred-brands: Done.

mikkelbu:
I'll try to answer the question about CompareTo first and then review the code later.

### Conclusion

My initial thought was that the analyzers should mimic NUnit as closely as possible and only show a warning if we are sure that the comparison would fail, i.e. if there is no `CompareTo` method (or only a `CompareTo(Type)` method where `Type` does not match the other argument). However, I think it make sense to have a more strict requirement - that the types should match for `IComparable`. If this becomes a problem for the users - I don't think so - then we can change the code to match NUnit more closely.

### Longer version

I did some digging trying to find why the logic is as it is now. In NUnit 3 we have https://github.com/nunit/nunit/blob/cefc0c3a86c065daa4130aa255b1384db51b2b08/src/NUnitFramework/framework/Constraints/NUnitComparer.cs#L61-L76 that first checks if either argument is `IComparable` and if so then `IComparable.CompareTo` is called with the other argument as argument. If neither argument is `IComparable` then we reflect after a method named "CompareTo" that accepts an argument of the other type, and if the method exists then we call it.

See https://github.com/nunit/nunit/blob/cefc0c3a86c065daa4130aa255b1384db51b2b08/src/NUnitFramework/framework/Constraints/NUnitComparer.cs#L61-L76 for implementation.

As NUnit 3 is based on NUnit 2 I looked into its history. And 11 years ago this [change](https://github.com/nunit-legacy/nunitv2/commit/301571e7502938486b9329cfeee7fccc8176164b#diff-52784dd90dfc0d7401829f750761e602) was made to *"Allow comparing objects that implement `IComparable<T>` but not `IComparable`"*. So it seems that the code - just reflecting for a method with the right name - was added to handle `IComparable<T>` and not a case where the method exists but no interface is implemented.

src/nunit.analyzers/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzer.cs
```diff
@@ -25,46 +25,68 @@ public class EqualToIncompatibleTypesAnalyzer : BaseAssertionAnalyzer
...
+            var expectedType = expectedOperation.Type;
+
+            if (actualType == null || expectedType == null)
+                return;
+
+            if (!NUnitEqualityComparerHelper.CanBeEqual(actualType, expectedType, context.Compilation))
+            {
+                context.ReportDiagnostic(Diagnostic.Create(
+                    descriptor,
+                        expectedOperation.Syntax.GetLocation()));
```
mikkelbu, line 86:
```suggestion
                    expectedOperation.Syntax.GetLocation()));
```

src/nunit.analyzers/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzer.cs
```diff
@@ -25,46 +25,68 @@ public class EqualToIncompatibleTypesAnalyzer : BaseAssertionAnalyzer
...
-                    || constraintMethod.ReturnType?.GetFullMetadataName() != NunitFrameworkConstants.FullNameOfEqualToConstraint)
+                foreach (var constraintPartExpression in constraintExpression.ConstraintParts)
                 {
-                    continue;
-                }
+                    if (HasIncompatiblePrefixes(constraintPartExpression)
+                        || HasCustomEqualityComparer(constraintPartExpression)
+                        || constraintPartExpression.HasUnknownExpressions())
+                    {
+                        return;
```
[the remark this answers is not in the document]
manfred-brands: With what I know now, I would say yes but this was explicitly changed from a `continue` into a `return` by @Dreamescaper 4 months ago (commit 1a004f7454250bbd01b0433d0605b9a1d9999623)
Changing this fails the following test:
```
var actual = new[] { 1, 2, 2, 1 };
Assert.That(actual, Has.All.EqualTo(1).Or.EqualTo(2));
```
This results in a `Has.AllEqualTo(1)` part which would be ignored and a `EqualTo(2)` part which is flagged as an error.
It seems that NUnit groups this to 'Has.All(.EqualTo(1).Or.EqualTo(2))

I tried to add a test that would miss detecting an error like in:
`Assert.That(5, Is.EqualTo("5").Using(new AsInt32Comparer()) & Is.EqualTo("5"));`
But I'm having trouble getting the NUnit to use a comparer:
```
            IEqualityComparer comparer = new AsInt32Comparer();
            Assert.That(comparer.Equals("5", 5), Is.True, "Comparer(string, int)");
            Assert.That(comparer.Equals(5, "5"), Is.True, "Comparer(int, string)");

            Assert.That("5", Is.EqualTo(5).Using(comparer), "NUnit using Comparer(string, int)");
            Assert.That(5, Is.EqualTo("5").Using(comparer), "NUnit using Comparer(int, string)");

            Assert.That(5, Is.EqualTo("5").Using(comparer) & Is.EqualTo(5), "NUnit Combined");
```
The first two asserts pass, proofing that my comparer works. The NUnit versions fail. This looks like the almost 4 year old bug: https://github.com/nunit/nunit/issues/1897
```
        public virtual bool CanCompare(object x, object y)
        {
            if (x is string && y is string)
                return true;

            if (x is IEnumerable || y is IEnumerable)
                return false;

            return true;
        }
```
Because only one of the arguments is `string` and `string` implements `IEnumerable` the method always returns `false`.

src/nunit.analyzers/SameAsOnValueTypes/SameAsOnValueTypesCodeFix.cs
```diff
@@ -54,20 +59,10 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
                     original = assertExpression;
                     replacement = assertExpression.WithName(SyntaxFactory.IdentifierName(NunitFrameworkConstants.NameOfAssertAreNotEqual));
                     break;
-                case NunitFrameworkConstants.NameOfAssertThat:
+                case NunitFrameworkConstants.NameOfIsSameAs:
                     // Find the 'SameAs' member access expression
```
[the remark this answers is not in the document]
manfred-brands: Comment removed.

manfred-brands (commented):
There are several loops over `constraintExpression.ConstraintParts` that bail out early. These expression are way more complex then just looping can handle without interpreting more. Not sure how we want to handle this.

src/nunit.analyzers/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzer.cs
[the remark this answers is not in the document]
Dreamescaper: Yeah, it is due to https://github.com/nunit/nunit.analyzers/issues/218.
Constraint expression is split by .And property, but it doesn't account for .All property operator in first part, which affects both parts.

mikkelbu:
Thanks for the contribution @manfred-brands. I'm merging it now.

PR #286, opened 2020-08-21
Add StyleCop to project

### LINKED ISSUE #20: Add StyleCop.Analyzers & fix warnings?
Hi,
I find https://www.nuget.org/packages/StyleCop.Analyzers nice for open source projects.
I can add it and fix the warnings in a PR if you want it.

Dreamescaper:
Fixes #20

Since we have lots of nitpick comments in PRs to fix spacing and formatting, I decided to open PR with suggestion to add StyleCop analyzer to project.
I used Unstable package, as current stable release has issues with nullabl annotations.

Based on current project style, I disabled the following rules:
- SA0001: Xml comment analysis disabled
- SA1101: Prefix local calls with this (disabled, but it is already checked by IDE analyzer)
- SA1116: Split parameters should start on line after declaration
- SA1117: Parameters should be on same line or separate lines
- SA1122: Use string.Empty for empty strings
- SA1124: Do not use regions
- SA1118: Parameter should not span multiple lines
- SA1200: Using directives should be placed correctly
- SA1201: Elements should appear in the correct order
- SA1202: Elements should be ordered by access
- SA1204: Static elements should appear before instance elements
- SA1311: Static readonly fields should begin with upper-case letter
- SA1316: Tuple element names should use correct casing
- SA1401: Fields should be private
- SA1413: Use trailing comma in multi-line initializers
- SA1503: Braces should not be omitted
- SA1512: Single-line comments should not be followed by blank line
- SA1516: Elements should be separated by blank line
- SA1600: Elements should be documented
- SA1623: Property summary documentation should match accessors
- SA1615: Element return value should be documented
- SA1611: Element parameters should be documented
- SA1612: Element parameter documentation should match element parameters
- SA1633: File should have header

If you think that any of those should be enabled (or the contrary - if any additional rules should be disabled), let me know.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -153,9 +153,6 @@ dotnet_diagnostic.SA1311.severity = none
 # SA1316: Tuple element names should use correct casing
 dotnet_diagnostic.SA1316.severity = none
 
-# SA1401: Fields should be private
-dotnet_diagnostic.SA1401.severity = none
-
 # SA1413: Use trailing comma in multi-line initializers
 dotnet_diagnostic.SA1413.severity = none
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -57,6 +57,10 @@ csharp_indent_case_contents = true
 csharp_indent_switch_labels = true
 csharp_indent_labels = flush_left
 
+# https://docs.nunit.org/articles/developer-info/Coding-Standards.html#file-organization
+dotnet_separate_import_directive_groups = false
+dotnet_sort_system_directives_first = true
+
 # https://docs.nunit.org/articles/developer-info/Coding-Standards.html#naming
 # Always use "this." and "Me." when applicable; let StyleCop Analyzers provide the warning and fix
 dotnet_style_qualification_for_field = true:warning
@@ -139,9 +143,6 @@ dotnet_diagnostic.SA1124.severity = none
 
 # SA1118: Parameter should not span multiple lines
 dotnet_diagnostic.SA1118.severity = none
-
-# SA1200: Using directives should be placed correctly
-dotnet_diagnostic.SA1200.severity = none
 
 # SA1201: Elements should appear in the correct order
 dotnet_diagnostic.SA1201.severity = none
--- a/.editorconfig
+++ b/.editorconfig
@@ -168,9 +168,6 @@ dotnet_diagnostic.SA1516.severity = none
 # SA1600: Elements should be documented
 dotnet_diagnostic.SA1600.severity = none
 
-# SA1623: Property summary documentation should match accessors
-dotnet_diagnostic.SA1623.severity = none
-
 # SA1615: Element return value should be documented
 dotnet_diagnostic.SA1615.severity = none
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -144,9 +144,6 @@ dotnet_diagnostic.SA1124.severity = none
 # SA1118: Parameter should not span multiple lines
 dotnet_diagnostic.SA1118.severity = none
 
-# SA1201: Elements should appear in the correct order
-dotnet_diagnostic.SA1201.severity = none
-
 # SA1202: Elements should be ordered by access
 dotnet_diagnostic.SA1202.severity = none
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -144,9 +144,6 @@ dotnet_diagnostic.SA1124.severity = none
 # SA1118: Parameter should not span multiple lines
 dotnet_diagnostic.SA1118.severity = none
 
-# SA1202: Elements should be ordered by access
-dotnet_diagnostic.SA1202.severity = none
-
 # SA1204: Static elements should appear before instance elements
 dotnet_diagnostic.SA1204.severity = none
 
--- a/.editorconfig
+++ b/.editorconfig
@@ -60,6 +60,7 @@ csharp_indent_labels = flush_left
 # https://docs.nunit.org/articles/developer-info/Coding-Standards.html#file-organization
 dotnet_separate_import_directive_groups = false
 dotnet_sort_system_directives_first = true
+csharp_using_directive_placement = outside_namespace:warning
 
 # https://docs.nunit.org/articles/developer-info/Coding-Standards.html#naming
 # Always use "this." and "Me." when applicable; let StyleCop Analyzers provide the warning and fix
--- a/.editorconfig
+++ b/.editorconfig
@@ -118,3 +118,75 @@ dotnet_naming_rule.public_fields.style = pascal_case
 dotnet_naming_rule.parameters_and_locals.severity = suggestion
 dotnet_naming_rule.parameters_and_locals.symbols = parameters_and_locals
 dotnet_naming_rule.parameters_and_locals.style = camel_case
+
+# SA0001: Xml comment analysis disabled
+dotnet_diagnostic.SA0001.severity = none
+
+# SA1101: Prefix local calls with this
+dotnet_diagnostic.SA1101.severity = none
+
+# SA1116: Split parameters should start on line after declaration
+dotnet_diagnostic.SA1116.severity = none
+
+# SA1117: Parameters should be on same line or separate lines
+dotnet_diagnostic.SA1117.severity = none
+
+# SA1122: Use string.Empty for empty strings
+dotnet_diagnostic.SA1122.severity = none
+
+# SA1124: Do not use regions
+dotnet_diagnostic.SA1124.severity = none
+
+# SA1118: Parameter should not span multiple lines
+dotnet_diagnostic.SA1118.severity = none
+
+# SA1200: Using directives should be placed correctly
+dotnet_diagnostic.SA1200.severity = none
+
+# SA1201: Elements should appear in the correct order
+dotnet_diagnostic.SA1201.severity = none
+
+# SA1202: Elements should be ordered by access
+dotnet_diagnostic.SA1202.severity = none
+
+# SA1204: Static elements should appear before instance elements
+dotnet_diagnostic.SA1204.severity = none
+
+# SA1311: Static readonly fields should begin with upper-case letter
+dotnet_diagnostic.SA1311.severity = none
+
+# SA1316: Tuple element names should use correct casing
+dotnet_diagnostic.SA1316.severity = none
+
+# SA1401: Fields should be private
+dotnet_diagnostic.SA1401.severity = none
+
+# SA1413: Use trailing comma in multi-line initializers
+dotnet_diagnostic.SA1413.severity = none
+
+# SA1503: Braces should not be omitted
+dotnet_diagnostic.SA1503.severity = none
+
+# SA1512: Single-line comments should not be followed by blank line
+dotnet_diagnostic.SA1512.severity = none
+
+# SA1516: Elements should be separated by blank line
+dotnet_diagnostic.SA1516.severity = none
+
+# SA1600: Elements should be documented
+dotnet_diagnostic.SA1600.severity = none
+
+# SA1623: Property summary documentation should match accessors
+dotnet_diagnostic.SA1623.severity = none
+
+# SA1615: Element return value should be documented
+dotnet_diagnostic.SA1615.severity = none
+
+# SA1611: Element parameters should be documented
+dotnet_diagnostic.SA1611.severity = none
+
+# SA1612: Element parameter documentation should match element parameters
+dotnet_diagnostic.SA1612.severity = none
+
+# SA1633: File should have header
+dotnet_diagnostic.SA1633.severity = none
--- a/.editorconfig
+++ b/.editorconfig
@@ -98,9 +98,6 @@ dotnet_style_coalesce_expression = true:suggestion
 dotnet_style_null_propagation = true:suggestion
 dotnet_style_explicit_tuple_names = true:suggestion
 
-# https://docs.nunit.org/articles/developer-info/Coding-Standards.html#file-organization
-dotnet_sort_system_directives_first = true
-
 # The first matching rule wins, more specific rules at the top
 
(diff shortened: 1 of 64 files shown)


manfred-brands:
> > SA1200: Using directives should be placed correctly. Enable. Add stylecop.json to specify placement outside namespace.
> 
> Yes, but I'd better use built-in diagnostic for that (frankly I thought it is already enabled), as it also affects code generation - `csharp_using_directive_placement = outside_namespace`. And it doesn't require `stylecop.json` file.
> 
Yes, we do need that directive. However we still might want the StyleCop part.
The difference is that the built-in diagnostics results in an IDE error which is not fired on a CI build (or dotnet build) See https://github.com/dotnet/roslyn/issues/6195 supposedly by design, It likely doesn't even warn those users using vscode or those using a different editor.
Unfortunately StyleCop doesn't read the `csharp_using_directive_placement` property. I'll see if I can raise that there.

> > SA1623: Property summary documentation should match accessors. I'm of the opinion that if something is documented, it should comply.
> 
> Even Roslyn itself does not follow those rule strictly, but I don't mind.
> 
Yes, I have seen several FxCop warnings in the past that are blatantly violated in the .NET runtime.
It seems to be _do as I say, not as I do_.

> > I also moved the StyleCop references from the individual projects into a Directory.Build.props file.
> > That way we only have to update one file when changing version or adding more analyzers (like FxCop)
> 
> Personally I'm not a fan of such approach (I prefer to be able to see dependencies explicitly), but it's not really required here - we can simple define it in `nunit.analyzers` project only, as tests project references that one anyway.
> 
Here you are right and I'm happy to go with that. 
In my work solution with 500 projects it is a completely different scenario.

> > I added two more commits to the branch enabling FxCop Analyzers and fixing violations.
> 
> Can you submit separate PR with FxCop analyzers? I tried too keep this PR smaller to make discussion easier.

Will do once StyleCop is in.

mikkelbu (commented):
Thanks for doing this @Dreamescaper. This is actually closing an old issue, #20, that I had forgot about.

Regarding SA1200, then I think @manfred-brands's suggestion makes sense, although it would be nice if we did not need this redundant specification.

Regarding SA1623, then I don't fell that strongly about it and I could go either way. On one hand, I don't find that the matching summary documentation provides much information, but on the other hand the fewer "special cases" (configuration of "dotnet_diagnostic" rules in this case) the better.

src/nunit.analyzers/ConstraintUsage/EqualConstraintUsageCodeFix.cs
```diff
@@ -46,6 +46,7 @@ private static (ExpressionSyntax? actual, ExpressionSyntax? expected) GetActualE
                     {
                         return (memberAccess.Expression, arguments[0].Expression);
                     }
+
```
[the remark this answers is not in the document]
Dreamescaper: I simply removed else, which is redundant in this case.

Dreamescaper:
Updated PR based on comments and cherry-picked discussed changes from @manfred-brands PR.

src/nunit.analyzers/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -10,8 +10,8 @@ internal static class IMethodSymbolExtensions
         /// </summary>
         /// <param name="this">The <see cref="IMethodSymbol"/> reference to get parameters from.</param>
         /// <returns>
-        /// The first count is the required parameters, the second is the optional count, 
-        /// and the last is the <code>params</code> count.
+        /// The first count is the required parameters, the second is the optional count,
+        /// and the last is the. <code>params</code> count.
```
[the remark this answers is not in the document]
manfred-brands: Technically it should be <see langword="params" />. That way if someone builds documentation for VB.Net it will be translated to `ParamArray`.

manfred-brands:
> Regarding SA1200, then I think @manfred-brands's suggestion makes sense, although it would be nice if we did not need this redundant specification.
@mikkelbu I got no reactions at [SetUp StyleCop using .editorconfig](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3045#issuecomment-67946181). I don't mind picking this up, but I would be interested in the internal GetOption methods from roslyn analyzer.

.editorconfig
```diff
@@ -57,6 +57,10 @@ csharp_indent_case_contents = true
 csharp_indent_switch_labels = true
 csharp_indent_labels = flush_left
 
+# https://docs.nunit.org/articles/developer-info/Coding-Standards.html#file-organization
+dotnet_separate_import_directive_groups = false
+dotnet_sort_system_directives_first = true
+
```
manfred-brands, line 63:
```suggestion
csharp_using_directive_placement = outside_namespace:warning
```

src/nunit.analyzers.tests/nunit.analyzers.tests.csproj
```diff
@@ -14,14 +14,22 @@
...
 
   <ItemGroup>
     <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
   </ItemGroup>
-  
+
+  <ItemGroup>
+    <AdditionalFiles Include="..\..\stylecop.json" Link="stylecop.json" />
+  </ItemGroup>
+
```
manfred-brands, line 34: Not really needed as you have added it as a solution item. But I don't mind if you also want it here.

src/nunit.analyzers/Constants/EqualToIncompatibleTypesConstants.cs
```diff
@@ -2,8 +2,8 @@ namespace NUnit.Analyzers.Constants
 {
     internal static class EqualToIncompatibleTypesConstants
     {
-        internal static string Title = "Incompatible types for EqualTo constraint.";
-        internal static string Message = "The EqualTo constraint always fails as the actual and the expected value cannot be equal.";
-        internal static string Description = "The EqualTo constraint always fails as the actual and the expected value cannot be equal.";
+        internal static readonly string Title = "Incompatible types for EqualTo constraint.";
```
manfred-brands, line 5:
```suggestion
        internal static const string Title = "Incompatible types for EqualTo constraint.";
```

src/nunit.analyzers/Constants/EqualToIncompatibleTypesConstants.cs
```diff
@@ -2,8 +2,8 @@ namespace NUnit.Analyzers.Constants
...
+        internal static readonly string Message = "The EqualTo constraint always fails as the actual and the expected value cannot be equal.";
```
manfred-brands, line 6:
```suggestion
        internal static const string Message = "The EqualTo constraint always fails as the actual and the expected value cannot be equal.";
```

src/nunit.analyzers/Constants/EqualToIncompatibleTypesConstants.cs
```diff
@@ -2,8 +2,8 @@ namespace NUnit.Analyzers.Constants
...
+        internal static readonly string Description = "The EqualTo constraint always fails as the actual and the expected value cannot be equal.";
```
manfred-brands, line 7:
```suggestion
        internal static const string Description = "The EqualTo constraint always fails as the actual and the expected value cannot be equal.";
```

src/nunit.analyzers/Constants/SameAsIncompatibleTypesConstants.cs
```diff
@@ -2,8 +2,8 @@ namespace NUnit.Analyzers
 {
     internal class SameAsIncompatibleTypesConstants
     {
-        internal static string Title = "Incompatible types for SameAs constraint.";
-        internal static string Message = "The SameAs constraint always fails because the actual and expected values have mutually exclusive types.";
-        internal static string Description = "The SameAs constraint always fails because the actual and expected values have mutually exclusive types.";
+        internal static readonly string Title = "Incompatible types for SameAs constraint.";
+        internal static readonly string Message = "The SameAs constraint always fails because the actual and expected values have mutually exclusive types.";
+        internal static readonly string Description = "The SameAs constraint always fails because the actual and expected values have mutually exclusive types.";
     }
```
manfred-brands, line 8: Same as for `EqualToIncompatibleTypesConstants`. Change to `const`.

src/nunit.analyzers/Constants/SameAsOnValueTypesConstants.cs
```diff
@@ -2,8 +2,8 @@ namespace NUnit.Analyzers.Constants
 {
     internal static class SameAsOnValueTypesConstants
     {
-        internal static string Title = "Non-reference types for SameAs constraint.";
-        internal static string Message = "The SameAs constraint always fails on value types as the actual and the expected value cannot be the same reference.";
-        internal static string Description = "The SameAs constraint always fails on value types as the actual and the expected value cannot be the same reference.";
+        internal static readonly string Title = "Non-reference types for SameAs constraint.";
+        internal static readonly string Message = "The SameAs constraint always fails on value types as the actual and the expected value cannot be the same reference.";
+        internal static readonly string Description = "The SameAs constraint always fails on value types as the actual and the expected value cannot be the same reference.";
     }
```
manfred-brands, line 8: Same as for `EqualToIncompatibleTypesConstants`. Change to `const`.

src/nunit.analyzers/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -10,8 +10,8 @@ internal static class IMethodSymbolExtensions
         /// </summary>
         /// <param name="this">The <see cref="IMethodSymbol"/> reference to get parameters from.</param>
         /// <returns>
-        /// The first count is the required parameters, the second is the optional count, 
-        /// and the last is the <code>params</code> count.
+        /// The first count is the required parameters, the second is the optional count,
+        /// and the last is the <c>params</c> count.
```
manfred-brands, line 14:
```suggestion
        /// and the last is the <see langword="params" /> count.
```

src/nunit.analyzers/Operations/ConstraintExpression.cs
```diff
@@ -8,13 +8,18 @@
 namespace NUnit.Analyzers.Operations
 {
     /// <summary>
-    /// Represents assert constraint expression, e.g. 'Is.EqualTo(expected)', 'Is.Not.Null & Is.Not.Empty'
+    /// Represents assert constraint expression, e.g. 'Is.EqualTo(expected)', 'Is.Not.Null & Is.Not.Empty'.
```
manfred-brands, line 11:
```suggestion
    /// Represents assert constraint expression, e.g. 'Is.EqualTo(expected)', 'Is.Not.Null &amp; Is.Not.Empty'.
```

src/nunit.analyzers/Operations/ConstraintExpression.cs
```diff
@@ -30,11 +35,6 @@ public ConstraintExpressionPart[] ConstraintParts
             }
         }
 
-        public ConstraintExpression(IOperation expressionOperation)
-        {
-            this.expressionOperation = expressionOperation;
-        }
-
         /// <summary>
         /// Split constraints into parts by binary ('&' or '|')  or constraint expression operators.
```
manfred-brands, line 39:
```suggestion
        /// Split constraints into parts by binary ('&amp;' or '|')  or constraint expression operators.
```

src/nunit.analyzers/Operations/ConstraintExpression.cs
```diff
@@ -47,7 +47,7 @@ private static IEnumerable<List<IOperation>> SplitConstraintByOperators(IOperati
 
         /// <summary>
         /// If provided constraint expression is combined using &, | operators - return multiple split expressions.
```
manfred-brands, line 49:
```suggestion
        /// If provided constraint expression is combined using &amp;, | operators - return multiple split expressions.
```

src/nunit.analyzers/Operations/ConstraintExpressionPart.cs
```diff
@@ -9,7 +9,7 @@
 namespace NUnit.Analyzers.Operations
 {
     /// <summary>
-    /// Represents part of <see cref="ConstraintExpression"/>, which is combined with others with 
+    /// Represents part of <see cref="ConstraintExpression"/>, which is combined with others with
     /// binary operators ('&', '|')  or methods ('And', 'Or', 'With').
```
manfred-brands, line 13:
```suggestion
    /// binary operators ('&amp;', '|')  or methods ('And', 'Or', 'With').
```

src/nunit.analyzers.tests/nunit.analyzers.tests.csproj
manfred-brands: Scratch that. It is needed here for StyleCop to pick it up.

manfred-brands (commented):
A few tiny things as mentioned with the individual files.

The invalid XML is not picked up because we are not generating XML documentation. This is a known issue with the Roslyn analyzers. To enable XML checking for validity we need to add the following to nunit.analyzers.csproj
```
<GenerateDocumentationFile>true</GenerateDocumentationFile>
```

The changes from `readonly` to `const` for the `Title` etc fields is not StyleCop but is for consistency. It will also be suggested when I add FxCop later. But as we already changed these we might as well complete it as part of this PR.

src/nunit.analyzers.tests/nunit.analyzers.tests.csproj
Dreamescaper: Yep :(
I've added it as solution item, because I think it is convenient when stylecop.json and .editorconfig are located together, but it makes no functional effect.

manfred-brands (approved):
Nothing further from me. Looks good.

mikkelbu (approved):
I also think it looks good. Thanks for doing this @Dreamescaper. I'll be merging it now.

PR #289, opened 2020-08-29
Add FxCop Analyzer

### LINKED ISSUE #288: Add FxCop Roslyn analyzer
As a follow up on StyleCop

manfred-brands:
Fixes #288 

I fixed the rule violations in separate commits so it would be easier to see what each rule does.

I have disabled the following rules:
* CA1031 Do not catch general exception types.
  * Only one violation which could be alternatively suppressed in the source file.
* CA1034: Nested types should not be visible
  * Only violation is in the `DocumentationTests` but as NUnit requires public methods we could not pass a nested private class as a `TestCaseSource`.
  * As this potential could be an issue in future test, I disabled the rule.
* CA1056: URI-like properties should not be strings
  * Only violation is in the `DocumentationTests`. Uri concatentation doesn't work. 
* CA1062: Validate arguments of public methods
  * Our only caller is the roslyn framework, no need to check arguments.
* CA1707: Identifiers should not contain underscores
  * Used in some Unit tests.
  * I added an extra commit to remove the _ for those tests. We use PascalCasing every where else. However, I'm happy to drop that commit if you think the test names are more readable that way.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -190,6 +190,3 @@ dotnet_diagnostic.CA1056.severity = none
 # CA1062: Validate arguments of public methods
 # Our callers are roslyn and ourselves.
 dotnet_diagnostic.CA1062.severity = none
-
-# CA1707: Identifiers should not contain underscores
-dotnet_diagnostic.CA1707.severity = none
--- a/.editorconfig
+++ b/.editorconfig
@@ -177,3 +177,19 @@ dotnet_diagnostic.SA1612.severity = none
 
 # SA1633: File should have header
 dotnet_diagnostic.SA1633.severity = none
+
+##################################################################################
+# FxCop Analyzers
+
+# CA1034: Nested types should not be visible
+dotnet_diagnostic.CA1034.severity = none
+
+# CA1056: URI-like properties should not be strings
+dotnet_diagnostic.CA1056.severity = none
+
+# CA1062: Validate arguments of public methods
+# Our callers are roslyn and ourselves.
+dotnet_diagnostic.CA1062.severity = none
+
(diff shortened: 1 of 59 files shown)


mikkelbu (approved):
Looks good to me @manfred-brands, and I'll happily merge it when you are ready.

Regarding CA1031, then I think we should suppress it in the source file, so we don't add more instances of this by accident.

Regarding CA1707, then I think it is fine to drop _, as I think the tests are just as easy to read without the `_`.

manfred-brands:
> Looks good to me @manfred-brands, and I'll happily merge it when you are ready.
> 
> Regarding CA1031, then I think we should suppress it in the source file, so we don't add more instances of this by accident.
> 
Removed rule from .editorconfig and added suppression to source file.

Rebased on master. Ready for merging.

Dreamescaper:
Does it make any sense forcing `ConfigureAwait(false)` in tests?

manfred-brands:
> Does it make any sense forcing `ConfigureAwait(false)` in tests?

I'm of the opinion that test code should conform to the same standards as other code. It is also a goto place for new users of a library to see how to use it.
For larger (integration) tests the addition of `ConfigureAwait(false)` could improve throughput.

Dreamescaper:
Not sure it can improve anything in tests, especially considering that currently NUnit uses sync-over-async for async tests)

But have nothing against common standard)

mikkelbu:
I agree that the `ConfigureAwait(false)` look a bit verbose in the tests, but like you I prefer the consistency in the codebase. Thanks for the contribution @manfred-brands. I'll merge the PR now.

PR #291, opened 2020-09-02
Actual value delegate

### LINKED ISSUE #290: Not treating Func<T> and Action properly for type compatibility.
The `UnwrapActualType` results in the return type for every delegate.
However NUnit doesn't always evaluate its _actual_ parameter.

```charp
Func<bool> function = () => false;
Assert.That(function, Is.Not.Null);
Assert.That(() => function, Is.Not.Null);
Assert.That(() => function(), Is.Not.Null);
```

Only the last one actually calls the _function_, but we do also evaluate the first.

The same method is called for the other rules testing for incompatible types.

manfred-brands:
Fixes #290 

The branch contains two commits, the first addressing the actual issue.

The 2nd commit adds a warning to something that caught me out:

```csharp
Assert.That(() => function(), Is.Not.Null);
Assert.That(() => action(), Is.Not.Null);
```

The two statements look the same, but the first actually calls _function_, but the 2nd does not call _action_.
The last test passes because it compares the delegate with _null_, not the _void_ result of the _action_.

With the 2nd commit, the analyzer will complain about incompatible types.

### DIFF
--- a/src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
@@ -389,14 +389,14 @@ namespace NUnit.Analyzers.Tests.EqualToIncompatibleTypes
         }
 
         [Test]
-        public void NoDiagnosticWhenActualAndExpectedTypesAreSameWithFuncActualValue()
+        public void AnalyzeWhenActualAndExpectedTypesAreSameWithFuncActualValue()
         {
             var testCode = TestUtility.WrapInTestMethod(@"
                 Func<string> actual = () => """";
                 var expected = """";
-                Assert.That(actual, Is.EqualTo(expected));");
+                Assert.That(actual, Is.EqualTo(↓expected));");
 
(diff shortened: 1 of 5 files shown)


mikkelbu (approved):
Looks good to me. I'm curious why you have chosen to use `EXPECT_TEST_DELEGATE_TO_BE_CALLED` instead of just having a comment, but I'm happy to merge it as it is.

manfred-brands:
> Looks good to me. I'm curious why you have chosen to use `EXPECT_TEST_DELEGATE_TO_BE_CALLED` instead of just having a comment, but I'm happy to merge it as it is.

Because the test as stands will fail. I could have tested on `actionCalled` to be `false` instead. Bit this makes things clearer.
I have rebased the branch on the latest master.
The last commit adds the actual nunit test to show nunit behaviour for `Action` and `Func<>`. It is not something we normally do as it doesn't test nunit.analyzer functionality. I'm happy to drop that commit if requested.

mikkelbu:
I agree that the test of the NUnit behaviour is a little untraditional, but I think it is fine to keep it. I, at least, would not be able to answer exactly what the behaviour of `Action` and `Func<>` would be without actually trying it out.

I'll be merging the PR now and try to get a new version out next week. Thanks for the issue and the PR @manfred-brands

PR #292, opened 2020-09-08
feat: Improve index.md to contain more information

### LINKED ISSUE #210: index.md in the documentation should also contain information about Severity
In the overview we should also have information about the Severity of each analyzer and whether we have a codefix associated with the analyzer. As this could take up most of the horizontal space we could do using icons similar to the approach taken in the README.md in https://github.com/wachulski/nunit-migrator.

Also we should perhaps partition the analyzers in index.md into sections with headings. One section per Category

mikkelbu:
Fixes #210

### DIFF
--- a/documentation/index.md
+++ b/documentation/index.md
@@ -1,71 +1,97 @@
-| Id       | Title       | Enabled by Default |
-| :--      | :--         | :--:               |
-| [NUnit1001](https://github.com/nunit/nunit.analyzers/tree/master/documentation/NUnit1001.md)| The individual arguments provided by a TestCaseAttribute must match the type of the corresponding parameter of the method. | :white_check_mark: |
-| [NUnit1002](https://github.com/nunit/nunit.analyzers/tree/master/documentation/NUnit1002.md)| The TestCaseSource should use nameof operator to specify target. | :white_check_mark: |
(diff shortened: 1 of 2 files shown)


mikkelbu:
I'm a bit unsure whether icons for the severity make sense, as I find it difficult to remember which icon is which severity when scrolling down.

Currently, the index look like this
image

manfred-brands (approved):
Code changes look good.

manfred-brands:
> I'm a bit unsure whether icons for the severity make sense, as I find it difficult to remember which icon is which severity when scrolling down.

I'm bad in icons, the color to me is the indication of severity.
Somehow the exclamation mark is left aligned iso centered like the others.

We seem to be inconsistent in severity, especially the Classic analyzers (which I added) some are _warning_, some _info_.
My suggestions are:
* Any code that is correct, like the Classic analyzer, should be _info_ by default.
* Any code that has maintainability issues, such as use `nameof`, should be a _warning_
* Any code that is wrong, such as incompatible types, should be _error_.

mikkelbu:
Thanks for the review @manfred-brands. I think we'll start with the current icons. I've centered all the columns with icons (I overlooked the syntax in the markdown tables). Regarding the inconsistent in severity, then I'll create a new issue for this.

PR #303, opened 2020-10-14
Fix TestCaseUsage array handling

### LINKED ISSUE #302: False positive NUnit1004 for test functions that receive array parameters
The code below causes NUnit1004 to be fired incorrectly on v0.5.0, except with the case where array has only a single element:

```csharp
[TestCase(new byte[] { })]
[TestCase(new byte[] { 1 })] // doesn't trigger NUnit1004
[TestCase(new byte[] { 1, 2 })]
[TestCase(new byte[] { 1, 2, 3 })]
[TestCase(new byte[] { 1, 2, 3, 4 })]
public void Test(byte[] buffer)
{
  // doesn't matter
}
```

This doesn't happen on 0.4.0.

Dreamescaper:
Fixes #302


mikkelbu (approved):
LGTM. Thanks for doing this @Dreamescaper 👍

PR #305, opened 2020-10-27
Fix false positive when TestCaseSource type is in another assembly

### LINKED ISSUE #304: Unable to compile test project when classlib with TestCaseSources is referenced
Hi,

I am trying to use NUnit.Analyzers and also TestCaseSource attribute documented on 

- https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcasesource.html
- https://docs.nunit.org/articles/nunit-analyzers/NUnit1011.html

I have prepared public gist showing the behavior I get while using classlib in test project solution. Maybe I am doing something wrong but I rather tell you about it as it might be scenario not used often.

Steps to reproduce:
- https://gist.github.com/adamvlasak/ff1acb7da54516b8f7a7625b339b88a5

Error I get (even though without NUnit.Analyzers tests run with no problem)
- https://gist.github.com/adamvlasak/ff1acb7da54516b8f7a7625b339b88a5#file-gistfile1-txt-L86

Can you please confirm it is a bug or tell me what I am doing wrong? Thank you very much.

Dreamescaper:
Fixes #304.

Analyzer cannot get source information when referenced type is in separate project.
Interestingly enough, in works in VS (no errors), but fails during build. 
Not sure why.

### DIFF
--- a/src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
@@ -1,3 +1,4 @@
+using System.Linq;
 using Gu.Roslyn.Asserts;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CodeFixes;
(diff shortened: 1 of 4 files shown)


src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
```diff
@@ -417,6 +418,51 @@ static IEnumerable<string> TestStrings(bool generateLongTestCase)
...
+            public static IEnumerable TestCaseDataProvider()
+            {
+                yield return new object[] { 1, 2, 3 };
+                yield return new object[] { 2, 3, 5 };
+            }
+        }
+    }";
+
+            var testDataReference = MetadataReferences.CreateBinary(testDataCode);
+            var references = AnalyzerAssert.MetadataReferences.Concat(new[] { testDataReference });
```
manfred-brands, line 457:
```suggestion
            var references = AnalyzerAssert.MetadataReferences.Append(testDataReference);
```

manfred-brands (approved):
Slight optimization to use `Append(T)` instead of `Concat(new T[])`

src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
Dreamescaper: It's not available for net461, which is one of our test targets (I'm not sure why though)

mikkelbu (approved):
LGTM and thanks for correcting my wrong attempt at finding the symbol 👍 .

PR #309, opened 2020-11-13
Nunit2041 false positives

### LINKED ISSUE #300: NUnit2041 false positives
Given test body
```C#
float x = 1;
C y = 1.1f;
Assert.That(y, Is.GreaterThan(x));
Assert.That(x, Is.LessThan(y));
object ox = x;
object oy = y;
Assert.That(oy, Is.GreaterThan(ox));
```
with
```C#
struct C : IComparable, IComparable<C> {
    float V { get; }
    public C(float v) => V = v;
    public static implicit operator C(float v) => new C(v);
    public int CompareTo(object obj) => V.CompareTo((float)obj));
    public int CompareTo(C c) => c.V.CompareTo(V);
}
```
All constraint arguments have false positive NUnit2041 errors.

manfred-brands:
Fixes #300 

Adds new rule NUnit2042 with severity information when one of the variables is an object.
Adds code to recognize generic parameter contraints.

### DIFF
--- a/documentation/NUnit2042.md
+++ b/documentation/NUnit2042.md
@@ -1,0 +1,86 @@
+# NUnit2042
+
+## Comparison constraint on object.
+
+| Topic    | Value
+| :--      | :--
+| Id       | NUnit2042
+| Severity | Info
+| Enabled  | True
+| Category | Assertion
+| Code     | [ComparableTypesAnalyzer](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/ComparableTypes/ComparableTypesAnalyzer.cs)
+
+## Description
+
+The comparison constraint might fail as the actual and the expected value might not implement IComparable.
+
+## Motivation
+
+```
+public static void ShouldBeGreaterThan(this object actual, object expected, string? message = null)
+{
+    Assert.That(actual, Is.GreaterThan(expected), message);
+}
+```
+
+In the above function it is assumed that every instance of object is comparable.
+The function works if the types actually are comparable, but will result in NUnit runtime errors if they are not.
+
+## How to fix violations
+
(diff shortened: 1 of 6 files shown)


manfred-brands:
I added one more commit. When using this in my work project, it was still firing. The only difference with the test in the analyzer project was that my work project was nullable. When I eventually managed to get VS to stop I noticed that the generic type of T looked the same, but were different references. Changing the == into Equals solved the problem.

In my work analyzer project, which is based upon the VS roslyn analyzer template, I get warnings about using == to compare symbols. It has a reference to the "Microsoft.CodeAnalysis.Analyzers" nuget package which raised RS1024 for this.
I managed to add a reference to an older version(2.9.8) as the latest doesn't support netstandard1.6 and it indeed gave a warning for this code (and one in SourceHelpers.ReportToUseNameOfIfApplicable) It might be good to add this (separate issue/PR)

However, I didn't manage to use '#nullable enable' in the nunit test, it wasn't recognized
I looked if there was a new version of Gu.Roslyn.Asserts, which there was, but upgrading wasn't compatible and outside the scope for this ticket and I still don't know if it then supports nullable.

mikkelbu (approved):
LGTM, and I'm happy to merge this as it is. Thanks for yet another contribution @manfred-brands.

Regarding "Microsoft.CodeAnalysis.Analyzers" then I think it is fine to add this (as another issue/PR).

Regarding '#nullable enable' then I don't think that this is supported now, but from GuOrg/Gu.Roslyn.Asserts#166 and Gu.Roslyn.Asserts#162 it seems that it will be in the future.

manfred-brands:
@mikkelbu Go ahead and merge as is. I will create a separate PR for Analyzers.

mikkelbu:
Thanks for the work @manfred-brands 👍 .

PR #310, opened 2020-11-13
NonTestMethodAccessibilityLevelAnalyzer: Detect Test attributes from overridden methods.

### LINKED ISSUE #308: NUnit1028 false positives
The code fires on `override` method where the `[Test]` attribute is set on the base class' method definition.

### LINKED ISSUE #311: NUnit1028 vs. IDisposable
Hi,

If you have a test fixture that implements ```IDisposable```, then the test runner will call ```Dispose()``` after running the tests.

If you have a fixure like that and run the nunit analyzers, you get an NUnit1028 warning on the Dispose() function:

image

I was just wondering if it would be considered reasonable to special case the Dispose() function and not warn about it, given that the test infrastructure does recognize and call it?

manfred-brands:
Fixes #308 
Fixes #311 

If a method is an override, check the attributes on the overridden method.

### DIFF
--- a/src/nunit.analyzers.tests/NonTestMethodAccessibilityLevel/NonTestMethodAccessibilityLevelAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/NonTestMethodAccessibilityLevel/NonTestMethodAccessibilityLevelAnalyzerTests.cs
@@ -81,5 +81,27 @@ namespace NUnit.Analyzers.Tests.NonTestMethodAccessibilityLevel
         {modifiers} void ↓AssertMethod() {{ }}");
             AnalyzerAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
         }
+
+        [TestCaseSource(nameof(TestMethodRelatedAttributes))]
+        public void AnalyzeWhenTestMethodIsPublicAndOverridden(string attribute)
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing($@"
+        public class BaseTestClass
+        {{
+            [{attribute}]
+            public virtual void TestMethod() {{ }}
+        }}
+
+        public sealed class DerivedTestClass : BaseTestClass
+        {{
+            public override void TestMethod() {{ }}
+
+            [Test]
+            public void OtherTestMethod() {{ }}
+        }}
+        ");
+
+            AnalyzerAssert.Valid<NonTestMethodAccessibilityLevelAnalyzer>(testCode);
+        }
     }
 }
--- a/src/nunit.analyzers.tests/NonTestMethodAccessibilityLevel/NonTestMethodAccessibilityLevelAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/NonTestMethodAccessibilityLevel/NonTestMethodAccessibilityLevelAnalyzerTests.cs
@@ -103,5 +103,74 @@ namespace NUnit.Analyzers.Tests.NonTestMethodAccessibilityLevel
 
             AnalyzerAssert.Valid<NonTestMethodAccessibilityLevelAnalyzer>(testCode);
         }
+
+        [Test]
+        public void AnalyzeConstructor()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+        public sealed class TestClass
+        {
+            public TestClass() {}
+
+            [Test]
+            public void TestMethod() { }
+        }
+        ");
+
+            AnalyzerAssert.Valid<NonTestMethodAccessibilityLevelAnalyzer>(testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenMethodIsExplicitDispose()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+        public sealed class TestClass : IDisposable
+        {
+            [Test]
+            public void TestMethod() { }
+
+            void IDisposable.Dispose() { }
+        }
+        ");
+
+            AnalyzerAssert.Valid<NonTestMethodAccessibilityLevelAnalyzer>(testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenMethodIsDispose()
+        {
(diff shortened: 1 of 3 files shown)


manfred-brands:
Added a commit to allow for public `Dispose` method for `IDisposable` as described in issues #311

src/nunit.analyzers/NonTestMethodAccessibilityLevel/NonTestMethodAccessibilityLevelAnalyzer.cs
```diff
@@ -79,5 +86,20 @@ private static bool IsPublicOrInternalMethod(IMethodSymbol method)
...
+        {
+            if (method.IsOverride)
+            {
+                return IsDisposeMethod(method.OverriddenMethod);
+            }
+
+            INamedTypeSymbol? type = method.ContainingType;
+            var disposeInterfaceMethods = type.AllInterfaces
+                    .Where(x => x.SpecialType == SpecialType.System_IDisposable)
+                    .SelectMany(i => i.GetMembers().OfType<IMethodSymbol>());
```
mikkelbu, line 100: Reading the code it seems a bit odd that `type` is typed as nullable and we then in the following line dereference it. Should this be `type?.AllInterfaces.Where(....).SelectMany(...) ?? Array.Empty<IMethodSymbol>()`?

src/nunit.analyzers/NonTestMethodAccessibilityLevel/NonTestMethodAccessibilityLevelAnalyzer.cs
```diff
@@ -79,5 +86,20 @@ private static bool IsPublicOrInternalMethod(IMethodSymbol method)
...
+
+            return disposeInterfaceMethods.Any(m => type.FindImplementationForInterfaceMember(m).Equals(method));
```
mikkelbu, line 102: I think FindImplementationForInterfaceMember sometimes can return null, but perhaps this is not the case here?

src/nunit.analyzers.tests/NonTestMethodAccessibilityLevel/NonTestMethodAccessibilityLevelAnalyzerTests.cs
```diff
@@ -81,5 +81,96 @@ public void AnalyzeWhenNonTestMethodIsPublicOrInternal(string modifiers)
...
+            [Test]
+            public void OtherTestMethod() {{ }}
+        }}
+        ");
+
+            AnalyzerAssert.Valid<NonTestMethodAccessibilityLevelAnalyzer>(testCode);
+        }
+
+        [Test]
+        public void AnalyzeConstructor()
```
mikkelbu, line 108: Is this just a testcase that we forgot? (I guess that this is handled by the existing check for `MethodKind.Ordinary`)?

mikkelbu (approved):
LGTM. I just have some questions and otherwise I happy to merge it in. Thanks both for the issue, the PR, and commenting on the other issues @manfred-brands.

src/nunit.analyzers/NonTestMethodAccessibilityLevel/NonTestMethodAccessibilityLevelAnalyzer.cs
manfred-brands: You are right. The problem is that we have nullable enabled but the Roslyn nuget package we reference has no nullability attributes, so I don't get the warnings in the compiler.  I will see if we can upgrade to a later package (in a separate PR). Are we constrained to netstandard1.6 in order to support VS2017?

Although I think in this case the `ContainingType` will never be null as the symbol comes from `type.GetMembers()`.
To be safe for future code changes, I will add a condition to bail out if type is unexpectedly null.

src/nunit.analyzers/NonTestMethodAccessibilityLevel/NonTestMethodAccessibilityLevelAnalyzer.cs
manfred-brands: Changed code to: `Equals(type.FindImplementationForInterfaceMember(m), method)`

src/nunit.analyzers.tests/NonTestMethodAccessibilityLevel/NonTestMethodAccessibilityLevelAnalyzerTests.cs
manfred-brands: Yes, I wanted to make sure that we don't trip up when the new nunit support for instance-per-test comes out where SetUp/TearDown is replaced with contructor/Dispose to better support nullable fields.

manfred-brands:
@mikkelbu All done, ready for merge.

manfred-brands:
@mikkelbu I found we had already a method: `IsInterfaceImplementation`. Simplified the call to `IsDisposeMethod` to use this.

src/nunit.analyzers.tests/NonTestMethodAccessibilityLevel/NonTestMethodAccessibilityLevelAnalyzerTests.cs
mikkelbu: God idea. I had not thought about this.

src/nunit.analyzers/NonTestMethodAccessibilityLevel/NonTestMethodAccessibilityLevelAnalyzer.cs
mikkelbu: Yes. The reason for netstandard1.6 is to still support VS 2017, but I'm considering if we should drop it. However, we still see some reports in the NUnit repo from people still on 2017 (and even 2015). Alternatives could be multitargeting or separate projects/repos, but both seem somewhat overkill. 

I'll give it some more thought.

mikkelbu:
Great find @manfred-brands. The simplified code is easier to read. I'll merge the PR now. Thanks for the contribution.

PR #313, opened 2020-11-15
Fix PropertyAnalyzer for combined asserts with All/Some operators

### LINKED ISSUE #306: NUnit2022 rule has false positive on hierarchies when using And
Hello,

it seems that the rule NUnit2022 has a false positive in case there are second level Prefixes (like another Property) applied. I say false positive, because with the analyzers not installed or disabled, the test runs as expected.

NUnit.Analyzers version: 0.5.0
NUnit: 3.12.0
Visual Studio 2019: 16.7.7

Given the following structure:

```
SomeClass
|- List<OtherClass>
   |- List<InnerListClass>
      |- LeafProperty
```

Given the assert for an instance of SomeClass:

```
Assert.That(
  someClass.OtherClasses,
  Has.Some.Property("InnerListClasses")
  .Some.TypeOf<SomeSpecificInnerListClass>()
  .And.Property("LeafProperty"));
```

*Eratta*: from before `TypeOf` a `Some` was missing in the original issue.

Expected: no error is raised by the analyzer, since `SomeClass` has a list of `OtherClass` instances, which in turn has a list of `InnerListClass` instances. The class `InnerListClass` has a property `LeafProperty`.
Actual: there is a compilation error raised because the analyzer says there is no `LeafProperty` on `IList<InnerListClass>`.

I also noted that in https://github.com/nunit/nunit.analyzers/blob/0.5.0/src/nunit.analyzers/MissingProperty/MissingPropertyAnalyzer.cs there is a comment on line 53. So the issue was considered when implementing, just not fully accommodated, I guess.

>  // Only first prefix supported (as preceding prefixes might change validated type)

Thank you!

Dreamescaper:
Fixes #306
(Or not, not sure).

Added checks for other operators, which might lead to false positive (e.g. All/Some)

### DIFF
--- a/src/nunit.analyzers.tests/MissingProperty/MissingPropertyAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/MissingProperty/MissingPropertyAnalyzerTests.cs
@@ -188,6 +188,23 @@ namespace NUnit.Analyzers.Tests.MissingProperty
         }
 
         [Test]
+        public void ValidWhenMultiplePartsCombinedWithCollectionOperatorPrefix()
+        {
+            var testCode = TestUtility.WrapInTestMethod(@"
+                var actual = new[]
+                {
+                    new
+                    {
+                        Foo = ""Fuzz"",
+                        Bar = ""Buzz""
+                    }
+                };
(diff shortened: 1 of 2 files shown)


mikkelbu (approved):
LGTM. Thanks for another contribution @Dreamescaper 👍 .

PR #315, opened 2020-11-16
Added Microsoft.CodeAnalysis.Analyzers and updated other packages.

### LINKED ISSUE #314: Add Microsoft.CodeAnalysis.Analyzers
This analyzer will detect possible issues in the nunit.analyzer code itself.

manfred-brands:
Fixes #314 

.NETStandard 1.6 cannot go beyond version 2.10.0 of Microsoft.CodeAnalysis.CSharp.Workspaces.
Unfortunately this still doesn't give use nullability attributes.

So I changed nunit.analyzer to be multi-target with .NetStandard 1.6 referencing version 2.10 and .NetStandard 2.0 referencing version 3.8.0

That way, when developing we get the nullability attributes which are in the 3.x versions.

* Fixed a few places where possible null reference exceptions could happen.
* Fixed ISymbol comparions the way the Roslyn team wants it. This required making a 'dummy' implementation of `SymbolEqualityComparer` for .NetStandard 1.6
* The Analyzer insists that Diagnostic Title's do not end in a period. This required updating the documentation.

There is one rule which I have disabled, which is the 'release tracking'.
This requires an AnalyzerReleases.Shipped.md and AnalyzerRelease.Unshipped.md


Dreamescaper:
@mikkelbu 
What do you think about dropping neststandard1.6? Do we still need that?

manfred-brands:
Rebased on latest master.

src/nunit.analyzers/Constants/ConstActualValueUsageAnalyzerConstants.cs
```diff
@@ -1,10 +1,12 @@
...
 namespace NUnit.Analyzers.Constants
 {
     internal static class ConstActualValueUsageAnalyzerConstants
     {
-        public const string Title = "The actual value should not be a constant.";
-        public const string Message = "The actual value should not be a constant - " +
-            "perhaps the actual value and the expected value have switched places.";
+        public const string Title = "The actual value should not be a constant";
+        public const string Message = "The actual value should not be a constant" +
+            "perhaps the actual value and the expected value have switched places";
```
mikkelbu, line 9: This will result in the word `constantperhaps`. I think we should either reintroduce the `-` or rephrase the message into two sentences, separated with e `.`.

src/nunit.analyzers/Constants/ValueSourceUsageConstants.cs
```diff
@@ -2,23 +2,23 @@ namespace NUnit.Analyzers.Constants
...
-        internal const string MethodExpectParametersMessage = "The ValueSource cannot supply parameters, but the target method expects '{0}' parameter(s).";
+        internal const string MethodExpectParametersTitle = "The target method expects parameters which cannot be supplied by the ValueSource";
+        internal const string MethodExpectParametersMessage = "The ValueSource cannot supply parameters, but the target method expects '{0}' parameter(s)";
         internal const string MethodExpectParametersDescription = "The target method expects parameters which cannot be supplied by the ValueSource.";
 
-        internal const string SourceDoesNotReturnIEnumerableTitle = "The source specified by the ValueSource does not return an IEnumerable or a type that implements IEnumerable.";
+        internal const string SourceDoesNotReturnIEnumerableTitle = "The source specified by the ValueSource does not return an IEnumerable or a type that implements IEnumerable";
         internal const string SourceDoesNotReturnIEnumerableMessage = "The ValueSource does not return an IEnumerable or a type that implements IEnumerable. Instead it returns a '{0}'.";
```
mikkelbu, line 22: I thought that this also should have stripped the trailing period, but looking at the analyzer this seems correct 

> RS1032: Define diagnostic message correctly
> The diagnostic message should not contain any line return character nor any leading or trailing whitespaces and should either be a single sentence without a trailing period or a multi-sentences with a trailing period

src/nunit.analyzers/Constants/TestMethodUsageAnalyzerConstants.cs
```diff
@@ -2,33 +2,33 @@ namespace NUnit.Analyzers.Constants
...
-        internal const string ExpectedResultTypeMismatchMessage = "The type of the value specified via ExpectedResult cannot match the method's return type '{0}'.";
+        internal const string ExpectedResultTypeMismatchTitle = "The type of the value specified via ExpectedResult must match the return type of the method";
+        internal const string ExpectedResultTypeMismatchMessage = "The type of the value specified via ExpectedResult cannot match the method's return type '{0}'";
         internal const string ExpectedResultTypeMismatchDescription = "The type of the value specified via ExpectedResult must match the return type of the method. " +
             "Otherwise, this will lead to an error at run-time.";
 
-        internal const string SpecifiedExpectedResultForVoidMethodTitle = "ExpectedResult must not be specified when the method returns void.";
+        internal const string SpecifiedExpectedResultForVoidMethodTitle = "ExpectedResult must not be specified when the method returns void";
         internal const string SpecifiedExpectedResultForVoidMethodMessage = "Don't specify ExpectedResult when the method returns void";
```
mikkelbu, line 11: So we were not 100% consistent before :)

src/nunit.analyzers/nunit.analyzers.csproj
```diff
@@ -2,25 +2,36 @@
...
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard1.6'" >
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.10.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="2.9.8" />
+  </ItemGroup>
+  
   <ItemGroup>
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.6.1" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.8.0" Condition="'$(TargetFramework)' == 'netstandard2.0'" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="2.10.0" Condition="'$(TargetFramework)' == 'netstandard1.6'" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.1" />
```
mikkelbu, line 25: This seem redundant given the conditional itemgroups above. Am I missing something?

mikkelbu (approved):
Wow for a contribution @manfred-brands 🥇. I've only found two minor issues/comments, otherwise I'm happy to merge it.

mikkelbu:
> What do you think about dropping netstandard1.6? Do we still need that?

My main concern is still to be able to support VS2017, as we still have many NUnit users that use this version of VS. Whether we solve this using multi-targeting or compiler constants I'm not so sure. The nice thing about multi-targeting is that it is inbuild in VS which makes it easy to change taget etc. But on the other hand it seems a bit odd to have to different tagets primarily to enable referencing different packages to help with NRT etc. 

I'm very open to ideas/suggestions :)

src/nunit.analyzers/nunit.analyzers.csproj
manfred-brands: You are not missing anything, it was me that forgot to delete these.

src/nunit.analyzers/Constants/ConstActualValueUsageAnalyzerConstants.cs
manfred-brands: Reintroduced the ` - `.

manfred-brands:
> But on the other hand it seems a bit odd to have to different targets primarily to enable referencing different packages to help with NRT etc.
We want to target the later packages to allow NRT and keep up with changes to the framework/language.
Multitargeting allows supporting older versions.

At work we multi-target .net5.0 for our libraries to take advantage of nullability annotations added to more runtime libraries such as System.Linq which are not available in earlier frameworks.

manfred-brands:
Code review items applied and branch rebased and squashed. Ready for merge.

mikkelbu (approved):
Thanks for the PR @manfred-brands. I'll merge it now.

PR #321, opened 2020-11-27
Detect TestCaseSource attributes defined in base class.

### LINKED ISSUE #320: Incorrect NUnit1011 when TestCaseSource references a base class member
Repros in 0.4.0–0.6.0, not present in 0.3.0

```cs
using NUnit.Framework;

class Base
{
    public static IEnumerable<string> TargetFrameworks { get; } = new[] { "net48", "netcoreapp3.1" };
}

class Derived : Base
{
    // ❌ NUnit1011 The TestCaseSource argument 'TargetFrameworks' does not specify an existing member
    //              ↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓↓
    [TestCaseSource(nameof(TargetFrameworks))]
    public void Test(string targetFramework)
    {
    }
}
```

manfred-brands:
Fixes #320

### DIFF
--- a/src/nunit.analyzers/SourceCommon/SourceHelpers.cs
+++ b/src/nunit.analyzers/SourceCommon/SourceHelpers.cs
@@ -79,11 +79,26 @@ namespace NUnit.Analyzers.SourceCommon
                 return null;
             }
 
-            var symbol = attributeInformation.SourceType
-                .GetMembers(attributeInformation.SourceName)
+            return attributeInformation.SourceType.GetMember(attributeInformation.SourceName);
+        }
+
+        public static ISymbol? GetMember(this INamedTypeSymbol typeSymbol, string name)
+        {
+            ISymbol? symbol = typeSymbol
+                .GetMembers(name)
                 .FirstOrDefault(m => m.Kind == SymbolKind.Field
                     || m.Kind == SymbolKind.Property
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers/SourceCommon/SourceHelpers.cs
```diff
@@ -79,12 +79,27 @@ internal static class SourceHelpers
...
+
+        public static ISymbol? GetMember(this INamedTypeSymbol typeSymbol, string name)
+        {
+            ISymbol? symbol = typeSymbol
+                .GetMembers(name)
                 .FirstOrDefault(m => m.Kind == SymbolKind.Field
                     || m.Kind == SymbolKind.Property
                     || m.Kind == SymbolKind.Method);
 
+            if (symbol is null && typeSymbol.BaseType != null)
```
jnm2, line 93: I think there needs to be a loop. A test that places the test case source in the base class of the base class would indicate this.

jnm2: Never mind, I missed the recursion.

mikkelbu (approved):
Looks good. Thanks for this improvement @manfred-brands.

PR #324, opened 2020-11-29
Move to latest GuOrg.Roslyn.Assert package

### LINKED ISSUE #323: Update to latest GuOrg.Roslyn.Asserts
We are behind and the current code doesn't fail if there are actually compiler errors in the test code strings.

manfred-brands:
Fixes #323 

The big downside with 'strings' as test is that they contained compile timer errors without us realizing it.
This new version of the package fails the test if there are compile time errors.
Fixed about 10 cases where our tests contained compile time errors.

The other downside is that we don't know if the test examples actually work on nunit.


mikkelbu (approved):
Looks good @manfred-brands. I hope that you have used a refactoring (or regular expression) for most of the changes (at least the PRs was very repetitive to review 😄)

mikkelbu:
Regarding 'strings' as test, then this project initially had the "test code" included in the project (I cannot remember if it was as regular code or just as content), but the setup was quite cumbersome, and it was quite annoying to add new tests.

I'm unsure if we can improve on "we don't know if the test examples actually work on nunit"? Perhaps we can dynamically compile the code in the valid code strings and and then run that using NUnitLite or something similar.

PR #325, opened 2020-11-29
Added DiagnosticsSuppressor for NullReference

### LINKED ISSUE #157: Suppress nullability warnings after Is.Not.Null constraint
As discussed here https://github.com/nunit/nunit/issues/3376#issuecomment-535879642, it would be great to suppress nullability warnings after Is.Not.Null asserts, e.g.
```csharp
string? str = GetNullable();
Assert.That(str, Is.Not.Null);
Assert.That(str.Length, Is.Not.Zero);  // <----- should be no warning here
```

Unfortunately, we cannot impact roslyn's nullability analysis, only suppress analyzers (see https://github.com/dotnet/roslyn/issues/32042#issuecomment-540524650), which means that at least very basic logic should be re-invented.

Cases that I think should be covered:
`Assert.That(obj?.PropA?.PropB, Is.Not.Null)`
`Assert.That(obj, Is.Not.Null & Is.Not.Empty)`
`Assert.That(obj, Is.Not.Null.And.Not.Empty)`

```csharp
Assert.Multiple(() =>
{
	string? str = GetNullable();
	Assert.That(str, Is.Not.Null);
	Assert.That(str.Length, Is.Not.Zero);  // <----- warning should remain here
});
```

manfred-brands:
Fixes #157 

Preliminary version:
* Detects: Assert.NotNull, Assert.IsNotNull and Assert.That(..., Is.Not.Null)
* Recognizes: Assert.Multiple
* Recognizes: Assignments

The current version is certainly not catching all cases.
I welcome suggestion for other test cases.

### DIFF
--- a/src/nunit.analyzers.tests/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressorTests.cs
+++ b/src/nunit.analyzers.tests/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressorTests.cs
@@ -1,0 +1,312 @@
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis.Diagnostics;
+using NUnit.Analyzers.DiagnosticSuppressors;
+using NUnit.Framework;
+
+namespace NUnit.Analyzers.Tests.DiagnosticSuppressors
+{
+    public class DereferencePossiblyNullReferenceSuppressorTests
+    {
+        private const string ABDefinition = @"
+            private static A? GetA(bool create) => create ? new A() : default(A);
+
+            private static B? GetB(bool create) => create ? new B() : default(B);
+
+            private static A? GetAB(string? text) => new A { B = new B { Text = text } };
+                
+            private class A
+            {
+                public B? B { get; set; }
+            }
+
+            private class B
+            {
+                public string? Text { get; set; }
+
+                public void Clear() => this.Text = null;
+
+                [System.Diagnostics.Contracts.Pure]
+                public string SafeGetText() => this.Text ?? string.Empty;
+            }
+        ";
+
+        private static readonly DiagnosticSuppressor suppressor = new DereferencePossiblyNullReferenceSuppressor();
+
+        [TestCase("")]
+        [TestCase("Assert.NotNull(string.Empty)")]
+        [TestCase("Assert.IsNull(s)")]
+        [TestCase("Assert.Null(s)")]
+        [TestCase("Assert.That(s, Is.Null)")]
+        public async Task NoValidAssert(string assert)
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@$"
+                #nullable enable
+                [TestCase("""")]
+                public void Test(string? s)
+                {{
+                    {assert};
+                    Assert.That(s.Length, Is.GreaterThan(0));
+                }}
+            ");
+
+            await DiagnosticsSuppressorAnalyzer.EnsureNotSuppressed(suppressor, testCode)
+                                               .ConfigureAwait(false);
+        }
+
+        [TestCase("Assert.NotNull(s)")]
+        [TestCase("Assert.IsNotNull(s)")]
+        [TestCase("Assert.That(s, Is.Not.Null)")]
+        public async Task WithLocalValidAssert(string assert)
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@$"
+                #nullable enable
+                [TestCase("""")]
+                public void Test(string? s)
+                {{
+                    {assert};
+                    Assert.That(s.Length, Is.GreaterThan(0));
+                }}
+            ");
+
+            await DiagnosticsSuppressorAnalyzer.EnsureSuppressed(suppressor,
+                DereferencePossiblyNullReferenceSuppressor.SuppressionDescriptors["CS8602"], testCode)
+                .ConfigureAwait(false);
+        }
+
+        [TestCase("Assert.NotNull(s)")]
+        [TestCase("Assert.IsNotNull(s)")]
+        [TestCase("Assert.That(s, Is.Not.Null)")]
+        public async Task WithFieldValidAssert(string assert)
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@$"
+                #nullable enable
+                private string? s;
+                [Test]
+                public void Test()
+                {{
+                    {assert};
+                    Assert.That(s.Length, Is.GreaterThan(0));
+                }}
+
+                public void SetS(string? v) => s = v;
+            ");
+
+            await DiagnosticsSuppressorAnalyzer.EnsureSuppressed(suppressor,
+                DereferencePossiblyNullReferenceSuppressor.SuppressionDescriptors["CS8602"], testCode)
+                .ConfigureAwait(false);
+        }
+
+        [Test]
+        public async Task ReturnValue()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                #nullable enable
+                [TestCase("""")]
+                public void Test(string? s)
+                {
+                    string result = DoSomething(s);
+                }
+
+                private static string DoSomething(string? s)
+                {
+                    Assert.NotNull(s);
+                    return s;
+                }
+            ");
+
+            await DiagnosticsSuppressorAnalyzer.EnsureSuppressed(suppressor,
+                DereferencePossiblyNullReferenceSuppressor.SuppressionDescriptors["CS8603"], testCode)
+                .ConfigureAwait(false);
+        }
+
+        [Test]
+        public async Task Parameter()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                #nullable enable
+                [TestCase("""")]
+                public void Test(string? s)
+                {
+                    Assert.NotNull(s);
+                    DoSomething(s);
+                }
+
+                private static void DoSomething(string s)
+                {
+                    _ = s;
+                }
+            ");
+
+            await DiagnosticsSuppressorAnalyzer.EnsureSuppressed(suppressor,
+                DereferencePossiblyNullReferenceSuppressor.SuppressionDescriptors["CS8604"], testCode)
+                .ConfigureAwait(false);
+        }
+
+        [Test]
+        public async Task NullableCast()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                #nullable enable
+                [Test]
+                public void Test()
+                {
+                    int? possibleNull = GetNext();
+                    Assert.NotNull(possibleNull);
+                    int i = (int)possibleNull;
+                    Assert.That(i, Is.EqualTo(1));
+                }
+
+                private static int? GetNext() => 1;
+            ");
+
+            await DiagnosticsSuppressorAnalyzer.EnsureSuppressed(suppressor,
+                DereferencePossiblyNullReferenceSuppressor.SuppressionDescriptors["CS8629"], testCode)
+                .ConfigureAwait(false);
+        }
+
+        [Test]
+        public async Task WithReassignedAfterAssert()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                #nullable enable
+                [TestCase("""")]
+                public void Test(string? s)
+                {
+                    Assert.NotNull(s);
+                    s = null;
+                    Assert.That(s.Length, Is.GreaterThan(0));
+                }
+            ");
+
+            await DiagnosticsSuppressorAnalyzer.EnsureNotSuppressed(suppressor, testCode)
+                                               .ConfigureAwait(false);
+        }
+
+        [Test]
+        public async Task WithReassignedFieldAfterAssert()
+        {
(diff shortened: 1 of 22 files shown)


manfred-brands:
Added recognition that `Assert.That(a?.B?.C, Is.Not.Null)` means that all three of `a`, `a.B` and `a.B.C` are not null.
Added recognition that assigning to `a` or `b` will invalidate this.

Cleaned up the code a bit.
Added more unit test cases.

manfred-brands:
Our .nuspec file was hardcoded to take the netstandard1.6 dll which does not support the DiagnosticSuppressor.
Roslyn analyzers cannot be multi-targeted. The [ticket](https://github.com/dotnet/roslyn-sdk/issues/107) has been open for almost 3 years. Roslyn itself has the 2.x range and the 3.x range. We could issue two sets of our package as well or drop support for VS2017.

For now I added an option to the build scripts to specify which dll to add to the nuget package: `-targetFramework netstandard2.0`
Depending the decision from @mikkelbu we will need to update the script to create two different nuget packages: 0.x range for VS2017 and 1.x range for VS2019 (or we could make the leap and use the same range as roslyn: 2.x and 3.x)

manfred-brands:
After using it in one of our NUnit projects which needed converting to Nullability, there were more CS errors that needs suppressing. Added a few more use cases.
Otherwise it seems to work.

manfred-brands:
The last commit adds a suppressor for nullable fields not initialized. The rule is suppressed if the field is initialized in the (OneTime)SetUp method. This helps for test setting class instances in the Setup of the test.

mikkelbu:
First and foremost, I very much appreciate this contribution, and I hope to have time to review it in detail the one of the following days (so far I've only skimmed the files).

I still think we should support VS2017, and I like your approach of being able to specify the targetFramework. I'm unsure of whether the best approach would be to have two separate feeds or one feed with two different versions numbers, but I'll give it some more thought.

src/nunit.analyzers/DiagnosticSuppressors/NonNullableFieldIsUninitializedSuppressor.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                return false;
+            }
+
+            foreach (var statement in method.Body.Statements)
+            {
+                if (statement is ExpressionStatementSyntax expressionStatement &&
+                    expressionStatement.Expression is AssignmentExpressionSyntax assignmentExpression)
+                {
+                    if (assignmentExpression.Left.ToString() == fieldName)
```
mikkelbu, line 69: Perhaps we should also handle when the left-side is prefixed with `this`, e.g. `this.someVariable`?

src/nunit.analyzers/DiagnosticSuppressors/NonNullableFieldIsUninitializedSuppressor.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                SyntaxNode? node = diagnostic.Location.SourceTree?.GetRoot(context.CancellationToken)
+                                                                  .FindNode(diagnostic.Location.SourceSpan);
+
+                if (node is null)
+                {
+                    continue;
+                }
+
+                string fieldName = node.ToString();
```
mikkelbu, line 36: Probably my lack of understanding of `DiagnosticSuppressor`, but can we do something more strongly typed than a string for the field? (same comment for the other suppressor)

mikkelbu (commented):
So far it looks good. I've tested the code a little and reviewed all but the tests (and `DiagnosticsSuppressorAnalyzer.cs`) - I'll see if I've time for the last part tomorrow night.

src/nunit.analyzers.tests/DiagnosticSuppressors/DiagnosticsSuppressorAnalyzer.cs
```diff
@@ -0,0 +1,109 @@
...
+                Assert.That(suppressionsHashSet, Does.Contain((suppressionDescriptor.Id, suppressionDescriptor.Justification)));
+            }
+
+            /*
+            if (error is DiagnosticWithProgrammaticSuppression diagnosticWithProgrammaticSuppression)
+            {
+                ImmutableHashSet<(string Id, LocalizableString Justification)> suppressionsHashSet =
+                diagnosticWithProgrammaticSuppression.ProgrammaticSuppressionInfo.Suppressions;
+            }
+            */
```
mikkelbu, line 106:
```suggestion
```

src/nunit.analyzers.tests/DiagnosticSuppressors/DiagnosticsSuppressorAnalyzer.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            // We cannot check for the correct suppression nicely.
+            // DiagnosticWithProgrammaticSuppression is a private class.
+            // ProgrammaticSuppressionInfo is an internal class.
+            // Resorting to reflection ...
+            Type errorType = error.GetType();
+            Assert.That(errorType.Name, Is.EqualTo("DiagnosticWithProgrammaticSuppression"));
+
+            var programmaticSuppressionInfoProperty = errorType.GetProperty("ProgrammaticSuppressionInfo",
+                                                                            BindingFlags.Instance | BindingFlags.NonPublic);
```
mikkelbu, line 86: I could not understand why we didn't need to check `programmaticSuppressionInfoProperty` for null, as my VS says that the type is `PropertyInfo?` (which is strange as the VS also states that `GetProperty` returns a `PropertyInfo`), and in the next line it says `'programmaticSuppressionInfoProperty' is not null here.`.

Is this because in ".NET Standard 2.1" the return type of `GetProperty` is `PropertyInfo`, so the compiler assumes that this cannot be null, but since I've also have ".NET 5.0" installed (which has `PropertyInfo?` as the return type), then VS shows me this information?

mikkelbu (approved):
Great work @manfred-brands 🥇 . I've only left a few comments and questions.

Regarding releasing this then I'm currently leaning towards your proposal of having two versions in the same feed (and use the same range as Roslyn - even though these version numbers are unknown for most people). But I think we should do this in a subsequent issue/PR to make the change more visible (with respect to release notes etc.).

I'm most happy to merge this when you are 👍

src/nunit.analyzers/DiagnosticSuppressors/NonNullableFieldIsUninitializedSuppressor.cs
manfred-brands: I added two test cases, but the one advantage of text based comparison is that the code already handles `this.member` as the left part of the assignment.
The downside it, it doesn't handle inconsistent use of the `this` prefix, e.g. using it in the assignment, but not in references. But that would already be flagged by the compiler (IDE0009).

src/nunit.analyzers/DiagnosticSuppressors/NonNullableFieldIsUninitializedSuppressor.cs
manfred-brands: CS8618 is raised for nulable fields, so we know that the SyntaxNode is a field. We don't actually care about the type.
But in this case, I do need to detect `this.field` as a separate assignment case: Code updated.

manfred-brands: For the actual null reference suppressor, we get the error reported on the expression. Again we don't care about the type. I check if the same expression that raises the null reference error is also used in the `Assert` call.
It fails when mixing expressions like in:
```
A? a = GetA(create);
Assert.That(a?.B?.Text, Is.Not.Null);
B b = a.B;
Assert.That(b.Text.Length, Is.GreaterThan(0));
```
Even though `b.Text` was tested as `a?.B?.Text` neither string comparison nor symbol analysis would help. Only if we could analyze the flow of assignments. The example is contrived, a more appropriate example would be:
```
A? a = GetA(create);
string? t = a?.B?.Text;
Assert.That(t, Is.Not.Null);
Assert.That(t.Length, Is.GreaterThan(0));
```

I am a bit disappointed that the following also fails:
```
A? a = GetA(create);
Assert.That(a?.B, Is.Not.Null);
B b = a.B;
string? t = b.Text;
Assert.That(t, Is.Not.Null);
Assert.That(t.Length, Is.GreaterThan(0));
```

The assignment `B b = a.B` where `b` is not nullable is suppressed correctly, but then the compiler still complains about `b` being possible `null` in `b.Text`. It looks like suppressors can only suppress an error, but not alter the compilers idea of what could be `null`.
It works when doing:
```
A? a = GetA(create);
B? b = a?.B;
Assert.That(b, Is.Not.Null);
string? t = b.Text;
Assert.That(t, Is.Not.Null);
Assert.That(t.Length, Is.GreaterThan(0));
```

src/nunit.analyzers.tests/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressorTests.cs
```diff
@@ -0,0 +1,271 @@
...
+
+        [Test]
+        public async Task ReturnValue()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                #nullable enable
+                [TestCase("""")]
+                public void Test(string? s)
+                {
+                    Assert.NotNull(s);
```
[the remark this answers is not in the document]
manfred-brands: Correct. I have removed it.

src/nunit.analyzers.tests/DiagnosticSuppressors/DiagnosticsSuppressorAnalyzer.cs
manfred-brands: Removed.

src/nunit.analyzers.tests/DiagnosticSuppressors/DiagnosticsSuppressorAnalyzer.cs
manfred-brands: All our target frameworks predate nullability. One way to get the correct annotations in is to use the excellent [TunnelVisionLabs.ReferenceAssemblyAnnotator](https://github.com/tunnelvisionlabs/ReferenceAssemblyAnnotator) to pull in .NET5.0 annotated references.
I did that and guess what, it complained about the `programmaticSuppressionInfoProperty` being null. So good catch by you. It also reported a few more.
I added two commits, one with updating the .csproj files and one with fixing the violations.

manfred-brands:
> I'm most happy to merge this when you are

The one thing I didn't add was documentation. As it suppresses compiler errors, it doesn't raise an issue by itself.
But maybe someone wants to disable the suppressors as it makes their code fail when analyzers are off. On our big project we have Live analyzers enabled (while editing in VS)so that code that gets touched, gets improved, but they are disabled during builds as older code doesn't (yet) comply. This means code won't compile even though it doesn't show errors in VS.
Disabling the suppression is a hassle until [roslyn#4972](https://github.com/dotnet/roslyn/issues/4972) is resolved.
It requires a separate .ruleset 

I added two documentation files and updated the index.md
This caught that I should have started the new range with NUnit3001 to be consistent with the others: Fixed.

The documentation is not yet tested in DocumentationTests.

manfred-brands:
> The documentation is not yet tested in DocumentationTests.

It is now.

mikkelbu (approved):
Again, great work. I've added some nitpick (mostly about the documentation), but besides this I think makes a great addition. As before I'm - even more - happy to merge this, when you are ready.

src/nunit.analyzers.tests/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressorTests.cs
```diff
@@ -0,0 +1,312 @@
...
+                #nullable enable
+                private string? s;
+                [Test]
+                public void Test()
+                {{
+                    {assert};
+                    Assert.That(s.Length, Is.GreaterThan(0));
+                }}
+
+                public void SetS(string? v) => s = v;
```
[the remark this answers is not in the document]
manfred-brands: This is needed to prevent a compile error (_warning CS0649: Field 'TestClass.s' is never assigned to, and will always have its default value null_) which causes the unit test to fail.

src/nunit.analyzers.tests/DocumentationTests.cs
```diff
@@ -351,9 +429,83 @@ private static string CreateStub(DiagnosticDescriptor descriptor)
...
+{EscapeTags(descriptor.Justification)}
+
+## Motivation
+
+ADD MOTIVATION HERE
+
+## How to fix violations
+
+ADD HOW TO FIX VIOLATIONS HERE
+
```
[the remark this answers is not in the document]
manfred-brands: Yes, added. And it wound an issue in the documentation using the wrong description in the .editorconfig section.

manfred-brands:
Thanks @mikkelbu for the review. Ready for merge.

src/nunit.analyzers.tests/DocumentationTests.cs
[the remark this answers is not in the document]
mikkelbu: Great 👍

src/nunit.analyzers/DiagnosticSuppressors/NonNullableFieldIsUninitializedSuppressor.cs
mikkelbu: Thanks for the explanation. It helped a lot on my understanding.

PR #327, opened 2020-12-01
Fix EqualTo analyzer when arguments have both IEnumerable and IEquatable

### LINKED ISSUE #322: NUnit2021 false positive for IEquatable<T>
NUnit2021 fires when different types `T1` and `T2` are compared, but `T1` implements `IEquatable<T2>`:

```csharp
using Microsoft.Extensions.Primitives;
using NUnit.Framework;

namespace UnitTests
{
    public class Tests
    {
        [Test]
        public void Test1()
        {
            var stringValues = new StringValues("abc");
            Assert.That(stringValues, Is.EqualTo("abc")); // <-- NUnit2021 fires here. Test passes ✅.
        }
    }
}
```

Dreamescaper:
Fixes #322.

We already support IEquatables, but interesting case here is that `string` implements `IEnumerable<char>`, and `StringValues` implements `IEnumerable<string>`.
So IEnumerable path took precedence before, we returned `false` based on that.

### DIFF
--- a/src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
@@ -649,6 +649,34 @@ using System.Collections.Generic;");
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


manfred-brands (approved):
Subtle change. Nice find.

mikkelbu (approved):
LGTM. Thanks for fixing this @Dreamescaper 👍 .

PR #334, opened 2021-01-12
Suppress "Possible Null" from a return value of Assert.Throws

### LINKED ISSUE #333: Add nullable suppression for Assert.Throws
`Assert.Throws` is declared to return `Exception?`. It will return `null` if no exception is thrown. This is only relevant when run inside an `Assert.Multiple` call.
To prevent developers having to check for `null` in the trivial cases, add a suppression rule if no `Assert.Multiple` is detected.

manfred-brands:
Fixes #333 

Suppress CS86xx for return values from Assert.Throw, except if run inside an Assert.Multiple


mikkelbu (approved):
LGTM and I like how only few change were needed to also handle the new cases 👍. I think we at some point should refactor `DereferencePossiblyNullReferenceSuppressor.ReportSuppressions`, as it is becoming somewhat long and nested, but I think this can wait for another day. 

I'm happy to merge this now if you are.

manfred-brands:
@mikkelbu I added a commit which refactors the code a little. Keeping it as two commits allow separating the refactoring with the addition of the new functionality. If you are happy, I'm ok to merge.

mikkelbu (approved):
@manfred-brands I think the refactoring is a good improvement. I'll be merging the PR now. Thanks for the contribution 👍 .

PR #337, opened 2021-01-14
Check for custom equality comparer in SomeItemsIncompatibleTypesAnalyzer

### LINKED ISSUE #332: NUnit2026 shouldn't fire when CustomEqualityComparer is used
The assertion below raises NUnit2026 `The 'Does.Contain' constraint cannot be used with actual argument of type 'string[]' and expected argument of type 'int'`, but there is no problem as we supply a custom comparer function. So if a custom comparer function is supplied to the constraint then we must not raise NUnit2026.

```csharp
[Test]
public void TestA()
{
    var actual = new[] { "1", "2", "3" };
    Func<string, int, bool> func = (a, b) => true;
    #pragma warning disable NUnit2026 // Wrong actual type used with the SomeItemsConstraint with EqualConstraint
    Assert.That(actual, Does.Contain(1).Using(func));
    #pragma warning restore NUnit2026 // Wrong actual type used with the SomeItemsConstraint with EqualConstraint
}
```

manfred-brands:
Fixes #332

### DIFF
--- a/src/nunit.analyzers.tests/Operations/ConstraintExpressionPartTests.cs
+++ b/src/nunit.analyzers.tests/Operations/ConstraintExpressionPartTests.cs
@@ -243,6 +243,30 @@ namespace NUnit.Analyzers.Tests.Operations
             Assert.That(constraintPart.HasUnknownExpressions(), Is.True);
         }
 
+        [Test]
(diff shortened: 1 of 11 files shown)


mikkelbu (approved):
This looks great. Thanks @manfred-brands

PR #338, opened 2021-01-15
Move constants

### LINKED ISSUE #335: Move Constant code closer to actual rule
We have now a directory called "Constants" which has files for all kinds of rules.
The files should be moved to the actual rule implementation directory so that all code for a rule is together.

manfred-brands:
Fixes #335


mikkelbu (approved):
Thanks for improving the codebase @manfred-brands. When looking through the files I was thinking that we also should change the namespaces in the moved files, so the namespace in the file match the folder. I can also do this in a subsequent PR?

manfred-brands:
> When looking through the files

I didn't even think about it, I just moved the files and compiled and ensure the tests all still passed, my bad.

> we also should change the namespaces in the moved files, so the namespace in the file match the folder

It should certainly be part of this PR: Update the code to do so.

mikkelbu (approved):
LGTM. Thanks for the contribution @manfred-brands 👍.

PR #354, opened 2021-02-07
Upgrade from FxCopAnalyzers to NetAnalyzers

### LINKED ISSUE #353: Replace deprecated FxCopAnalyzer with NetAnalyzer
The FxCopAnalyzers are marked as Legacy and no longer maintained as mentioned on:
https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers/
And will trigger CA9998 as mentioned on: https://github.com/dotnet/roslyn-analyzers/releases/tag/v3.3.2

manfred-brands:
Fixes #353 

Whilst in the process of upgrading dependencies.

### DIFF
--- a/src/nunit.analyzers/nunit.analyzers.csproj
+++ b/src/nunit.analyzers/nunit.analyzers.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <MinimumVisualStudioVersion>11.0</MinimumVisualStudioVersion>
     <RootNamespace>NUnit.Analyzers</RootNamespace>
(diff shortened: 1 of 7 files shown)


src/nunit.analyzers/nunit.analyzers.csproj
```diff
@@ -7,31 +7,35 @@
...
     </PackageReference>
-    <PackageReference Include="Microsoft.CodeAnalysis.FxCopAnalyzers" Version="3.3.1">
+    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.3">
       <PrivateAssets>all</PrivateAssets>
-      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.CodeStyle" Version="3.8.0">
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.3.2" />
```
mikkelbu, line 38: Before we used different versions of this for the `netstandard1.6`/`netstandard2.0` builds, but now we use the same. Is this safe?

manfred-brands: The Analyzers are a compile time dependency to the build process to watch out for pit-falls and not platform dependent.
We only have one code base for both builds. In a few places, even the previous analyzer suggested to use SymbolEqualityComparer where we made a pseudo netstandard 1.6 version to prevent #if/#else in our code.

mikkelbu: Thanks for the explanation @manfred-brands 👍

mikkelbu (approved):

PR #359, opened 2021-02-13
Add DelegateRequiredAnalyzer

### LINKED ISSUE #149: Warning when non-delegate actual value is used with DelayedConstraint
Case 1:
```csharp
int i = 0;
int actualFunc() => i++;
Assert.That(actualFunc(), Is.EqualTo(5).After(1000, 10));
```

DelayedConstraint is allowed to be used with non-delegate values, but that makes no sense, since it will never change, and most likely is not intended.
I suggest to add analyzer for this case, with Warning severity.

Case 2:
```csharp
Assert.That("123 ".Trim, Is.EqualTo("123"));
```
Opposite case. Delegate actual argument is allowed for any Constraint, but it will simply invoke it. 
It might be a bit confusing though, so I suggest adding analyzer for such cases (with Info severity).
Exception: async delegates.

### LINKED ISSUE #357: Warns about passing a non-lambda to Throws assertions.
People sometimes write `Assert.That(MyMethod(), Throws.InstanceOf<SomeException>());`  and wonder why it fails.

We need an analyzer rule that detects this and with code-fixes that convert the code to:
`Assert.That(() => MyMethod(), Throws.InstanceOf<SomeException>());`

manfred-brands:
Fixes #357 
Fixes #149 

There is no need to check classic method Assert.Throw etc as the argument definition already specifies a delegate.

### DIFF
--- a/documentation/NUnit2044.md
+++ b/documentation/NUnit2044.md
@@ -1,0 +1,65 @@
+# NUnit2044
+
+## Non-delegate actual parameter
+
+| Topic    | Value
+| :--      | :--
+| Id       | NUnit2044
+| Severity | Error
+| Enabled  | True
+| Category | Assertion
+| Code     | [DelegateRequiredAnalyzer](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/DelegateRequired/DelegateRequiredAnalyzer.cs)
+
+## Description
+
+The actual argument needs to be evaluated by the Assert to catch any exceptions.
+
+## Motivation
+
+In order for the `Assert.That` to catch an exception or a timeout, the code must be
+a delegate so it can be evaluated by the method. If the parameter is not a
+delegate, it will be evaluated before the call to `Assert.That` and stop further
+execution.
+
+## How to fix violations
+
+Convert the call into a delegate using a lambda expression or method group.
+
+<!-- start generated config severity -->
+## Configure severity
+
+### Via ruleset file
+
+Configure the severity per project, for more info see [MSDN](https://msdn.microsoft.com/en-us/library/dd264949.aspx).
+
+### Via .editorconfig file
+
+```ini
+# NUnit2044: Non-delegate actual parameter
+dotnet_diagnostic.NUnit2044.severity = chosenSeverity
+```
+
+where `chosenSeverity` can be one of `none`, `silent`, `suggestion`, `warning`, or `error`.
+
+### Via #pragma directive
+
+```csharp
+#pragma warning disable NUnit2044 // Non-delegate actual parameter
+Code violating the rule here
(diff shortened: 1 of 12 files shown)


manfred-brands:
Now also addresses case 1 of #149

documentation/NUnit2044.md
```diff
@@ -0,0 +1,65 @@
...
+## Motivation
+
+In order for the `Assert.That` to catch an exception or a timeout, the code must be
+a delegate so it can be evaluated by the method. If the parameter is not a
+delegate, it will be evaluated before the call to `Assert.That` and stop further
+execution.
+
+## How to fix violations
+
+Convert the call into a delegate using a lambda expressionor method group.
```
mikkelbu, line 26:
```suggestion
Convert the call into a delegate using a lambda expression or method group.
```

src/nunit.analyzers.tests/DelegateRequired/DelegateRequiredAnalyzerTests.cs
```diff
@@ -0,0 +1,109 @@
+using System;
+using System.Threading.Tasks;
+using Gu.Roslyn.Asserts;
+using Microsoft.CodeAnalysis.Diagnostics;
+using NUnit.Analyzers.Constants;
+using NUnit.Analyzers.DelegateRequired;
+using NUnit.Framework;
+
+namespace NUnit.Analyzers.Tests.ThrowsWithNonDelegate
```
mikkelbu, line 9:
```suggestion
namespace NUnit.Analyzers.Tests.DelegateRequired
```

src/nunit.analyzers.tests/DelegateRequired/DelegateRequiredCodeFixTests.cs
```diff
@@ -0,0 +1,94 @@
+using Gu.Roslyn.Asserts;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Diagnostics;
+using NUnit.Analyzers.Constants;
+using NUnit.Analyzers.DelegateRequired;
+using NUnit.Framework;
+
+namespace NUnit.Analyzers.Tests.ThrowsWithNonDelegate
```
mikkelbu, line 8:
```suggestion
namespace NUnit.Analyzers.Tests.DelegateRequired
```

mikkelbu (approved):
Looks great @manfred-brands. I some typos, but that is all. I'm happy to merge this when you are 👍.

manfred-brands:
Thanks @mikkelbu, I thought I had renamed everything when I extended the scope of the ticket, but forgot the test namespaces.
All done now.

mikkelbu (approved):

mikkelbu:
@manfred-brands Thanks for the contribution, I'll merge it now 👍

PR #362, opened 2021-02-14
Within usage analyzer + tests + codefix

### LINKED ISSUE #344: Add a rule informing that .Within is not valid for non-numeric types.
See https://github.com/nunit/nunit/issues/3736

The nunit syntax allows:
```
[Test]
public void RecordsEqualsMismatch()
{
    var a = new Data(1, 1.0);
    var b = new Data(1, 1.1);

    Assert.That(a, Is.EqualTo(b).Within(0.2), $"{a} != {b}");
}

private sealed record Data(int number, double Value);
```

But the `.Within` part is completely ignored. Only the NumericEquals and TupleEquals will use it.

Add a rule warning the user when the `.Within` constraint is not valid.
Add a CodeFix to remove the constraint.

Antash:
Closes #344


manfred-brands:
I added a [PullRequest](https://github.com/Antash/nunit.analyzers/pull/1) on top of your branch which you can merge into your branch.
It fixes the CodeFix for your single use case, so you can continue with more nunit tests and add the documentation.
Note that if you run all nunit tets you get a skeleton of the new documentation, rename the file and add the missing bits.

src/nunit.analyzers/WithinUsage/WithinUsageAnalyzer.cs
```diff
@@ -0,0 +1,118 @@
...
+            foreach (var constraintPart in constraintExpression.ConstraintParts)
+            {
+                if (!SupportedIsMethods.Contains(constraintPart.GetConstraintName()))
+                    continue;
+
+                // e.g. Is.EqualTo(expected).Within(0.1)
+                // Need to check type of expected
+
+                var withinSuffix = constraintPart.GetSuffix(NunitFrameworkConstants.NameOfEqualConstraintWithin) as IInvocationOperation;
+
```
manfred-brands, line 50: Note that you don't allow for Is.EquivalentTo

Antash: Noted!

Antash:
@manfred-brands many thanks! Will finish the work soon

Antash:
@manfred-brands I think I'm ready with the changes, could you please take a look?

src/nunit.analyzers/WithinUsage/WithinUsageAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+using NUnit.Analyzers.Helpers;
+
+namespace NUnit.Analyzers.WithinUsage
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public class WithinUsageAnalyzer : BaseAssertionAnalyzer
+    {
+        private static readonly string[] SupportedIsMethods = new[]
+        {
+            NunitFrameworkConstants.NameOfIsEqualTo
```
manfred-brands, line 20:
```suggestion
             NunitFrameworkConstants.NameOfIsEqualTo,
            NunitFrameworkConstants.NameOfIsLessThan,
            NunitFrameworkConstants.NameOfIsLessThanOrEqualTo,
            NunitFrameworkConstants.NameOfIsGreaterThan,
            NunitFrameworkConstants.NameOfIsGreaterThanOrEqualTo,
```

You can actually do: `Assert.That(Math.PI, Is.LessThan(Math.E).Within(0.5));`

src/nunit.analyzers/WithinUsage/WithinUsageAnalyzer.cs
manfred-brands: You will have to find a way to deal with the other use cases.

src/nunit.analyzers/WithinUsage/WithinUsageAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+            var fullName = namedType.GetFullMetadataName();
+
+            if (fullName.StartsWith("System.Tuple`", StringComparison.Ordinal))
+                return namedType.TypeArguments.All(t => IsTypeSupported(t, checkedTypes));
+
+            if (fullName.StartsWith("System.TimeSpan", StringComparison.Ordinal))
+                return true;
+
+            if (fullName.StartsWith("System.DateTime", StringComparison.Ordinal))
+                return true;
```
manfred-brands, line 117: Are you trying to match System.DateTimeOffset as System.DateTime is already checked earlier. Now you also match System.DateTimeKind which is not supported. If you want to match System.DateTimeOffset use Equals

src/nunit.analyzers/WithinUsage/WithinUsageAnalyzer.cs
```diff
@@ -0,0 +1,122 @@
...
+            // Allowed - tuples
+            if (namedType.IsTupleType)
+                return namedType.TupleElements.All(e => IsTypeSupported(e.Type, checkedTypes));
+
+            var fullName = namedType.GetFullMetadataName();
+
+            if (fullName.StartsWith("System.Tuple`", StringComparison.Ordinal))
+                return namedType.TypeArguments.All(t => IsTypeSupported(t, checkedTypes));
+
+            if (fullName.StartsWith("System.TimeSpan", StringComparison.Ordinal))
```
manfred-brands, line 113:
```suggestion
            if (fullName.Equals("System.TimeSpan", StringComparison.Ordinal))
```

manfred-brands (changes requested):
Thanks @Antash  The code looks good.
I found that WithIn can also be supplied to ComparisonConstrants (LessThan etc).
So please add support/some test for that.

Also have some test cases that use the 'Not' prefix, like: `Is.Not.EqualTo` and `Is.Not.LessThan` 
Further some small items.

src/nunit.analyzers/WithinUsage/WithinUsageAnalyzer.cs
Antash: Ah I see. Thank you for spotting this!

Antash:
@manfred-brands thanks for the instant review! I've addressed points you raised.

manfred-brands (approved):
Beside the spelling errors in the parameter name, nothing else from me.
It might be good to rebase and squash the branch into a single commit.

Antash:
@manfred-brands Thank you! Didn't know how to rebase without the conflict so 2 commits finally.

mikkelbu (commented):
Good work @Antash. I've left some comments, but most of them are minor issues.

Hopefully, I'll have time to try out the analyzer/codefix tomorrow (for some manual testing), so far I've just reviewed the code.

mikkelbu:
> @manfred-brands Thank you! Didn't know how to rebase without the conflict so 2 commits finally.

No problem. We can also do a "squash and merge" when merging this in, so we only get one commit.

src/nunit.analyzers/WithinUsage/WithinUsageAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+                return false;
+
+            // Allowed - tuples
+            if (namedType.IsTupleType)
+                return namedType.TupleElements.All(e => IsTypeSupported(e.Type, checkedTypes));
+
+            var fullName = namedType.GetFullMetadataName();
+
+            if (fullName.StartsWith("System.Tuple`", StringComparison.Ordinal))
+                return namedType.TypeArguments.All(t => IsTypeSupported(t, checkedTypes));
```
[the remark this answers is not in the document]
Antash: Ah, you are right here, my bad.

Antash:
@mikkelbu thank you for your comments, all addressed!

src/nunit.analyzers/WithinUsage/WithinUsageAnalyzer.cs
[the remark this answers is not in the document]
mikkelbu: NP 😄

mikkelbu:
Hi @Antash 

I did some manual testing and found some issues (I should have thought of these when I did the review and I'm sorry for the very long comment).

We have some special comparers in NUnit - see https://github.com/nunit/nunit/tree/master/src/NUnitFramework/framework/Constraints/Comparers - that override the normal equality for certain types (mostly collections, tuples, dictionaries etc.). Several of these make use of the `Tolerance` parameter (and we have already handled Tuples, DateTimeOffsets, numeric in this PR). As far as I can tell we need to handle:
* Arrays (or more generally IEnumerables) where we use the tolerance in the comparison of the individual items
* IDictionary (note we only use the tolerance in the comparison of the values)
* DictionaryEntry (note we only use the tolerance in the comparison of the values)
* KeyValuePair (note we only use the tolerance in the comparison of the values)
* IStructuralEquatable (e.g. ImmutableArray  as far as I can remember)

So for instance the following `Within` is currently wrongly marked by the analyzer
```csharp
        [Test]
        public void TestArray()
        {
            var value = 5.5;
            var value2 = 6;
            Assert.That(new[] { value }, Is.EqualTo(new[] { value2 }).Within(2));
        }
```
but the test pass, as the tolerance is used in the comparison.

Similarly, the following test pass (so `Within` is wronly marked), as we use the tolerance on the values in dictionaries, dictionaryentries, and keyvalue paris)
```csharp
        [Test]
        public void DictionaryEntryValue()
        {
            var a = new DictionaryEntry("key", 1);
            var b = new DictionaryEntry("key", 2);
            Assert.That(a, Is.EqualTo(b).Within(2));
        }
```

The mark of `Within` in the following is correct, as we don't use tolerance on strings
```csharp
        [Test]
        public void DictionaryEntryKey()
        {
            var a = new DictionaryEntry(1, "key");
            var b = new DictionaryEntry(2, "key");
            Assert.That(a, Is.EqualTo(b).Within(2));
        }
```

Also note that we pass down the tolerance recursively, so the following is valid (although a bit contrived example)
```csharp
        [Test]
        public void NestedCase()
        {
            var a = new KeyValuePair<string, KeyValuePair<string, KeyValuePair<string, int>>>(
                "key1", new KeyValuePair<string, KeyValuePair<string, int>>("Key2", new KeyValuePair<string, int>("Key3", 3)));
            var b = new KeyValuePair<string, KeyValuePair<string, KeyValuePair<string, int>>>(
                "key1", new KeyValuePair<string, KeyValuePair<string, int>>("Key2", new KeyValuePair<string, int>("Key3", 4)));
            Assert.That(a, Is.EqualTo(b).Within(2));
        }
```

So I think we need to examine the types recursively to figure out if the `Within` make sense (similarly to how the comparers work), so if we compare collections, we need to check if the collection elements are of a type that use tolerance (and probably handle `object` as a type that allows tolerance as well); if we check DictionaryEntries, we need to check if the value is of a type that use tolerance (and ignore the key) ...

Note that we have a special comparison for Chars (and this does not make use of the tolerance). But currently the analyzer does not mark Within in the following test
```csharp
        [Test]
        public void Char()
        {
            var a = 'a';
            var b = 'b';
            Assert.That(a, Is.EqualTo(b).Within(2));
        }
```
but the test fails with 
```
      Expected: 'b' +/- 2
      But was:  'a'
      Off by:   1.0d
```
So I think we should also mark Within on chars.

manfred-brands:
Replaced with #537

PR #369, opened 2021-03-03
Also detect NonNullableProperties being initialized in SetUp methods.

### LINKED ISSUE #368: Extend NonNullableFieldIsUninitialized to also work for properties.
The situation when a base class fixture sets a protected non-nullable property in its SetUp method.

manfred-brands:
Fixes #368


mikkelbu (approved):
Looks good. Thanks for the contribution @manfred-brands 👍

PR #386, opened 2021-08-21
Issue 385: Detect indirect assignment of non-nullable fields.

### LINKED ISSUE #385: Supress (even more) 'Non-nullable field must contain a non-null value when exiting constructor (CS8618)'
I started to migrate a larger codebase towards nullable reference types and the Analyzers are of great use. Many thanks for the effort so far!

The analyzer is already capable of suppressing simple cases where the variable is assigned within the `SetUp()` method explicitly (Case `someVariable`).

```c#
[TestFixture]
public class TestClass
{
    private string someVariable; // CS8618 very nicely suppressed!
    private string someVariableThatGetSetInACodeAnalysisDecoratedFunction; // CS8618 does not get (currently) suppressed

    [SetUp]
    public void SetUp()
    {
        someVariable = "value";
        SetValue();
    }

    [MemberNotNull("someVariableThatGetSetInACodeAnalysisDecoratedFunction")]
    private void SetValue()
    {
        someVariableThatGetSetInACodeAnalysisDecoratedFunction = "also-value";
    }
}
```

I was wondering if it is possible to detect some more complex cases (Case `someVariableThatGetSetInACodeAnalysisDecoratedFunction`)? Of course it may not be a good idea to do extensive travelling of the AST to find an assignment deep down in some function called by `SetUp()`,

Instead, I thought it might be possible to take advantage of the `System.Diagnostics.CodeAnalysis` attributes. In this case e.g. `MemberNotNull`. In such a case it would only the functions that are called directly called by `SetUp()` needed to be analyzed and checked for the `MemberNotNull` attribute. Then the attributes argument has to be matched against the member variables that are causing warnings to find one that could be suppressed.

What are your thoughts on this?

manfred-brands:
Fixes #385

### DIFF
--- a/documentation/NUnit3002.md
+++ b/documentation/NUnit3002.md
@@ -20,6 +20,8 @@ This rule check diagnostics reported by the CS8618 compiler error:
 If the violating field/property is set in the `SetUp` or `OneTimeSetUp` method. The rule suppresses the error.
 This allows for non-nullable fields/properties to be used in a `TestFixture`.
 
+The rule does detect indirect calls, when the field is set in a method called by the `SetUp` or `OneTimeSetUp` methods.
+
 ## Motivation
 
 ```csharp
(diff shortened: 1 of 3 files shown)


mikkelbu (approved):
Looks great @manfred-brands (and I like that the new solution is simpler).

I'll merge it when you are happy.

manfred-brands:
@mikkelbu Thanks. Commits Squashed. Ready for merge.

mikkelbu:
Merged. Thanks for the contribution @manfred-brands 👍

PR #391, opened 2021-10-11
Use assert.multiple

### LINKED ISSUE #389: Idea: Suggest Assert.Multiple in tests where Assert statements are called back to back
Just happened to be thinking -- a lot of folks don't know about [NUnit Multiple Asserts](https://docs.nunit.org/articles/nunit/writing-tests/assertions/multiple-asserts.html) and how it can be helpful in situations where you may want multiple distinct assertions that don't rise to the level of separate tests. 

In tests that contain multiple asserts -- or where asserts exist one after another -- I think it'd be helpful to suggest this.

manfred-brands:
Fixes #389 

The code only works on `Assert.That`, other code fixes will convert to that first.
An `Assert.That` is considered independent if the argument is not a suffix of an earlier argument:

```
Assert.That(instance.Property1, Is.EqualTo(0));
Assert.That(instance.Property2, Is.EqualTo(0));
Assert.That(instance.Property12, Is.EqualTo(0));
```

But not:
```
Assert.That(instance, Is.Not.Null);
Assert.That(instance, Has.Length.EqualTo(1));
Assert.That(instance.Length, Is.EqualTo(1));
Assert.That(instance[0], Is.EqualTo(' '));
```

More complex cases might fail.

The issue raised is of level Info.

### DIFF
--- a/documentation/NUnit2045.md
+++ b/documentation/NUnit2045.md
@@ -1,0 +1,78 @@
+# NUnit2045
+
+## Use Assert.Multiple
+
+| Topic    | Value
+| :--      | :--
+| Id       | NUnit2045
+| Severity | Info
+| Enabled  | True
+| Category | Assertion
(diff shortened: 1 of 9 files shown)


manfred-brands:
Squashed the commits, ready for release.

PR #395, opened 2021-10-24
Add UseCollectionConstraintAnalyzer

### LINKED ISSUE #330: Analyzer to warn against calling .Length or .Count() in the actual expression
Add an analyzer or set of analyzers to warn against calling `.Count()` or `.Length` in the actual expression of `Assert.That`
For example, `Assert.That(value.Count(), Is.EqualTo(5))` should be `Assert.That(value, Has.Count.EqualTo(5))`

manfred-brands:
Fixes #330

### DIFF
--- a/documentation/NUnit2046.md
+++ b/documentation/NUnit2046.md
@@ -1,0 +1,90 @@
+# NUnit2046
+
+## Use CollectionConstraint for better assertion messages in case of failure
+
+| Topic    | Value
+| :--      | :--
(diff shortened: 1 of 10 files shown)


manfred-brands:
@mikkelbu Split the large method up, separating the 'Is.Empty' and 'Is.Not.Empty' cases in a separate method.

You can merge if you're happy with it.

mikkelbu (approved):
Great work. I'll be merging it now

PR #400, opened 2021-11-04
Lang version9

### LINKED ISSUE #397: Allow using C# 9
@mikkelbu Would you be willing to accept a PR to change language version to C#9.
This does not affect runtime as most of it is syntax compiled to the same IL.

However it allows replacing: `!(instance is null)` with `instance is not null` which is far more readable.

We also can remove repeating long generic types with targeted new expression:
```
private static readonly ConcurrentDictionary<Type, CodeFile> cache =
    new ConcurrentDictionary<Type, CodeFile>();
```

Becomes
```
private static readonly ConcurrentDictionary<Type, CodeFile> cache = new();
 ```

### PR DESCRIPTION
**Pull request by manfred-brands**
Fixes #397

### DIFF
--- a/src/nunit.analyzers/UseCollectionConstraint/UseCollectionConstraintCodeFix.cs
+++ b/src/nunit.analyzers/UseCollectionConstraint/UseCollectionConstraintCodeFix.cs
@@ -37,10 +37,11 @@ namespace NUnit.Analyzers.UseCollectionConstraint
 
             var node = root.FindNode(context.Span);
 
-            if (!(node is ArgumentSyntax actualArgument) ||
-                !(actualArgument.Expression is MemberAccessExpressionSyntax actualExpression) ||
-                !(actualArgument.Parent is ArgumentListSyntax argumentList) ||
-                !(argumentList.Arguments.Count > 1 && argumentList.Arguments[1] is ArgumentSyntax constraintArgument))
+            if (node is not ArgumentSyntax actualArgument ||
+                actualArgument.Expression is not MemberAccessExpressionSyntax actualExpression ||
+                actualArgument.Parent is not ArgumentListSyntax argumentList ||
+                argumentList.Arguments.Count <= 1 ||
+                argumentList.Arguments[1] is not ArgumentSyntax constraintArgument)
             {
                 return;
             }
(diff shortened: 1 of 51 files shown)

### CODE REVIEW

File: src/nunit.analyzers/UseCollectionConstraint/UseCollectionConstraintCodeFix.cs [hunk 1]
```diff
@@ -37,18 +37,19 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
             var node = root.FindNode(context.Span);
 
-            if (!(node is ArgumentSyntax actualArgument) ||
-                !(actualArgument.Expression is MemberAccessExpressionSyntax actualExpression) ||
-                !(actualArgument.Parent is ArgumentListSyntax argumentList) ||
-                !(argumentList.Arguments.Count > 1 && argumentList.Arguments[1] is ArgumentSyntax constraintArgument))
+            if (node is not ArgumentSyntax actualArgument ||
+                actualArgument.Expression is not MemberAccessExpressionSyntax actualExpression ||
+                actualArgument.Parent is not ArgumentListSyntax argumentList ||
+                argumentList.Arguments.Count == 1 ||
```
mikkelbu, line 43: Shouldn't this be `argumentList.Arguments.Count <= 1 ||` ?

Before this change we would return if `argumentList.Arguments.Count == 0` now I think we would fail as we are accessing index 1

mikkelbu (approved):
I only have one comment. All the rest looks great 👍

src/nunit.analyzers/UseCollectionConstraint/UseCollectionConstraintCodeFix.cs
```diff
@@ -37,18 +37,19 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+                argumentList.Arguments[1] is not ArgumentSyntax constraintArgument)
```
mikkelbu, line 44: Shouldn't the first line be `argumentList.Arguments.Count <= 1 ||` ?

Before this change we would return if `argumentList.Arguments.Count == 0` now I think we would fail as we are accessing index 1 (in the second line)

manfred-brands: If it helps with readability, yes, I'm in favour.

Practically it isn't necessary.
The Diagnostic is not even raised if there are less than two arguments (see `AssertHelper.TryGetActualAndConstraintOperations`)
The line above ensures that the node we are dealing with is an `ArgumentSyntax` with the `argumentList` being its `Parent`. This implies that there is at least one argument.

I will update it.
I also have to fix the rebase issues.

manfred-brands:
Change made, code rebased, resolved conflicts.
Build is green, ready for merge when you are.

src/nunit.analyzers/UseCollectionConstraint/UseCollectionConstraintCodeFix.cs
mikkelbu: Thanks 👍

mikkelbu:
Looks great. I'll be merging it now.

PR #409, opened 2021-11-26
Chore/analyzer versions

manfred-brands:
Updates Analyzers etc as suggested by dependabot.

### DIFF
--- a/src/nunit.analyzers/nunit.analyzers.csproj
+++ b/src/nunit.analyzers/nunit.analyzers.csproj
@@ -29,7 +29,7 @@
     <PackageReference Include="StyleCop.Analyzers.Unstable" Version="1.2.0.354">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="5.0.3">
+    <PackageReference Include="Microsoft.CodeAnalysis.NetAnalyzers" Version="6.0.0">
       <PrivateAssets>all</PrivateAssets>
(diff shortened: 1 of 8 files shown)


manfred-brands:
@mikkelbu The build fails because the new Gu.Roslyn.Asserts defaults to C#10 triggering https://github.com/nunit/nunit/issues/3993
I have asked if we can set the language version on Gu.Roslyn (https://github.com/GuOrg/Gu.Roslyn.Asserts/issues/231)

src/nunit.analyzers/nunit.analyzers.csproj
```diff
@@ -15,27 +15,19 @@
   </PropertyGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'" >
-    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="3.8.0" />
+    <PackageReference Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.0.1" />
```
mikkelbu, line 18: Wouldn't this affect the version of Visual Studio that can use the analyzer? I thought that Roslyn 3.x mean VS 2019, so I would assume that version 4.x means VS 2022 ?

mikkelbu (commented):
I can see that Gu.Roslyn.Asserts have been released as 4.0.1 with a lot of breaking changes. If that is the case, then I think we should remove bumping Gu.Roslyn.Asserts from this PR, and then we can make that a separate PR for bumping Gu.Roslyn.Asserts.

PR #411, opened 2021-11-30
NUnit2044: Check type of operation instead.

### LINKED ISSUE #410: NUnit2044 false-positive on explicit delegate type
```csharp
TestDelegate action = resourceStreams.Commit;
Assert.That(action, Throws.InvalidOperationException.With.Message.EqualTo("The unit of work is read-only."));
```

This triggers **NUnit2044**:The type of the actual argument must be a delegate.

In this case, the type already _is_ a delegate, as `resourceStreams.Commit` is a method defined as `public void Commit()`.

manfred-brands:
Fixes #410 

This allows for local/parameters of type Delegate.


mikkelbu (approved):
Thanks for another simple solution @manfred-brands. I'll merge it now.

PR #427, opened 2022-02-24
Add README.md to NuGet package

### LINKED ISSUE #418: Add README.md to nuget file
Currently we have the following "warning" on NuGet
image

See https://docs.microsoft.com/en-us/nuget/reference/nuspec#readme for how to add the readme to the nuget file.

manfred-brands:
Fixes #418 

@mikkelbu Not sure if we should include this "README.md" as this one contains the build status and also references the documentation on github using a relative path.

I'll see what it does on the myget feed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Prerelease nuget packages can be found on [MyGet](https://www.myget.org/feed/nun
 
 ## Analyzers ##
 
-The full list of analyzers can be found in the [documentation](documentation/index.md).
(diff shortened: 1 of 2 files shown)


manfred-brands:
It does look quite good on [myget](https://www.myget.org/feed/nunit-analyzers/package/nuget/NUnit.Analyzers).
As expected the `documentation` link doesn't work. Changed link into absolute.

Issue #416, opened 2022-01-02
Option to make a single package version support multiple versions of Roslyn


jnm2:
StrongInject is gaining the ability to build for both Roslyn 3.8 and 4.0 in the same package: https://github.com/YairHalberstadt/stronginject/pull/181

A similar (but not exactly identical) approach could be used to build for both Roslyn 2.x and 3.x (and even 4.x) in one package if you were interested.

The .NET SDK starting with v6 (= VS2022) natively understands how to properly load from .nupkg structures such as `analyzers/dotnet/roslyn3.8/cs/nunit.analyzers.dll`, and a package .targets file can fill in the gap for earlier versions.

manfred-brands:
@jnm2 I have used this approach for my company analyzer and it works for switching between 3.x and 4.0 version.
How would you update the target to switch between 2.x, 3.x and 4.x? is the Roslyn version known as an msbuild variable?

jnm2:
I'd have to look through MSBuild Structured Log Viewer for candidates, but I'm pretty sure I've seen the SDK version itself as a properly and the major versions line up in parallel if all else fails.

manfred-brands:
@mikkelbu I will be working on this. Can you assign it to me (or give me assign limited rights in this repository)?

mikkelbu:
Thanks @manfred-brands. I've assigned the issue to you and I will look into giving you more rights in this repo as you have been the main driving force the last couple of years 👍

PR #432, opened 2022-03-19
Net6 sdk

### LINKED ISSUE #272: Enforce coding standard more strictly
Increase severity of "suggestion" and "none" to "warning". See https://github.com/nunit/nunit.analyzers/pull/264#discussion_r460432710 and the rest of the thread for context.

manfred-brands:
Fixes #272 

- Upgrade to .NET6 SDK
- Move common definitions from individual .csproj files to Directory.Build.props
- Fix nullability issues.
- Do what we preach, reference nunit.analyzers as an Analyzer
- Convert suggestions for code style to warnings. The code was already clean, but this way we keep it that way.

@mikkelbu this is in preparation of #416 where we need way more projects, one for every supported compiler, to prevent duplication, I moved to using the Directory.Build.props to host common definitions.


mikkelbu:
Thanks for doing this @manfred-brands. I think I'll have time to review it tonight or tomorrow.

src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
```diff
@@ -46,22 +46,28 @@ public void VerifySupportedDiagnostics()
...
-                    $"{diagnostic.Id} : {nameof(DiagnosticDescriptor.Title)}");
-                Assert.That(diagnostic.Category, Is.EqualTo(Categories.Structure),
-                    $"{diagnostic.Id} : {nameof(DiagnosticDescriptor.Category)}");
-                Assert.That(diagnostic.DefaultSeverity, Is.EqualTo(DiagnosticSeverity.Error),
-                    $"{diagnostic.Id} : {nameof(DiagnosticDescriptor.DefaultSeverity)}");
-            }
+                foreach (var diagnostic in diagnostics)
+                {
+                    Assert.That(diagnostic.Title.ToString(CultureInfo.InvariantCulture), Is.Not.Empty,
+  $"{diagnostic.Id} : {nameof(DiagnosticDescriptor.Title)}");
```
mikkelbu, line 54:
```suggestion
                        $"{diagnostic.Id} : {nameof(DiagnosticDescriptor.Title)}");
```

src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
```diff
@@ -46,22 +46,28 @@ public void VerifySupportedDiagnostics()
...
+                }
+            });
 
             var diagnosticMessage = diagnostics.Select(_ => _.MessageFormat.ToString(CultureInfo.InvariantCulture)).ToImmutableArray();
 
-            Assert.That(diagnosticMessage, Contains.Item(TestMethodUsageAnalyzerConstants.ExpectedResultTypeMismatchMessage),
+            Assert.Multiple(() =>
+            {
+                Assert.That(diagnosticMessage, Contains.Item(TestMethodUsageAnalyzerConstants.ExpectedResultTypeMismatchMessage),
                 $"{TestMethodUsageAnalyzerConstants.ExpectedResultTypeMismatchMessage} is missing.");
```
mikkelbu, line 67:
```suggestion
                    $"{TestMethodUsageAnalyzerConstants.ExpectedResultTypeMismatchMessage} is missing.");
```

src/nunit.analyzers.tests/Extensions/ITypeSymbolExtensionsTests.cs
```diff
@@ -208,9 +208,12 @@ private static async Task<ImmutableArray<ITypeSymbol>> GetTypeSymbolAsync(string
 
             foreach (var typeName in typeNames)
             {
-                types.Add(rootAndModel.Model.GetDeclaredSymbol(rootAndModel.Node
-                    .DescendantNodes().OfType<TypeDeclarationSyntax>()
-                    .Where(_ => _.Identifier.ValueText == typeName).Single()));
+                INamedTypeSymbol? item = rootAndModel.Model.GetDeclaredSymbol(
+                    rootAndModel.Node
+                                .DescendantNodes().OfType<TypeDeclarationSyntax>()
+                                .Where(_ => _.Identifier.ValueText == typeName).Single());
```
mikkelbu, line 214: I was wondering if we should move this into a variable for readability. So something like

```
TypeDeclarationSyntax typeDeclaration = rootAndModel.Node
    .DescendantNodes()
    .OfType<TypeDeclarationSyntax>()
    .Where(_ => _.Identifier.ValueText == typeName)
    .Single();
INamedTypeSymbol? item = rootAndModel.Model.GetDeclaredSymbol(typeDeclaration);
...
```

mikkelbu (approved):
Looks great @manfred-brands 👍 . I only have the smallest of comments, but I'm also happy to merge it as it is.

src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
manfred-brands: Done.

src/nunit.analyzers.tests/TestMethodUsage/TestMethodUsageAnalyzerTests.cs
manfred-brands: Done.

src/nunit.analyzers.tests/Extensions/ITypeSymbolExtensionsTests.cs
manfred-brands: This code was more or less in there from the beginning, but you are right; splitting is make it more readable.
Done.

manfred-brands:
Thanks @mikkelbu for your review. Changes applied, ready for merge.

mikkelbu:
Thanks for the PR @manfred-brands. I'm merging it now.

manfred-brands:
@mikkelbu This might have broken your versioning of the dlls which I detected when working on #416 
I suggest to delay any release until then.
There is only one bugfix in since the last version and they should give a reason when using `Assert.Ignore()` and `Assert.Fail()`.

mikkelbu:
@manfred-brands Thanks for noticing this. I should have given it some thought when I performed the review

manfred-brands:
@mikkelbu This was because I removed AssemblyInfo.cs files which were used in build.cake to add Generated.AssemblyInfo.cs with version information.
This is fixed with #439 which set the Version parameter as part of the DotNetBuild step.

manfred-brands:
We should also get #438 into that new release.

PR #433, opened 2022-03-19
Relax raising DelegateRequiredAnalyzer on non-value types.

### LINKED ISSUE #431: NUnit2044 false positive with `Has`
The following code works as intended, but is banned with NUnit2044.
``` C#
var list = new List<int>();
Task.Run(async () => {
    await Task.Delay(500);
    list.Add(0);
});
Assert.That(list, Has.Count.EqualTo(1).After(1000, 1));
```
Maybe there are other use cases where a delegate condition isn't necessary.

To stay with the analyzers, simply use
``` C#
Assert.That(() => list, Has.Count.EqualTo(1).After(1000, 1));
```
for now.

manfred-brands:
Fixes #431 

Added a test to not throw if the DelayedConstraint is on a class.

It will lessen the amount of issues raised to prevent many that do work without a delegate
With an added risk that it is not always raised when it was needed.

### DIFF
--- a/src/nunit.analyzers.tests/DelegateRequired/DelegateRequiredAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/DelegateRequired/DelegateRequiredAnalyzerTests.cs
@@ -131,5 +131,29 @@ namespace NUnit.Analyzers.Tests.DelegateRequired
 
             RoslynAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
         }
+
+        // https://github.com/nunit/nunit.analyzers/issues/431
+        [Test]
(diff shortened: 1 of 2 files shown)


chm-tm:
Thanks for your fast response.
This approach would perfectly guard all 100+ occurrences where we use `.After` in our code base.
But I suspect that there could be many occurrences with checks of the original expected value against `null` or a `string` which wouldn't longer be guarded.
Given that I've found just one single occurrence triggering the false positive, I prefer adding the `() =>` there and continue to have reference type values covered by `NUnit2044`.

manfred-brands:
Better more warnings than missing some.

PR #435, opened 2022-03-27
Chore/416 single version

manfred-brands:
Still in draft.
Code compiles, switching works between SDK 5 and 6.


manfred-brands:
Abandonded in favour of #609

PR #438, opened 2022-03-29
NullReferenceSuppressor: Keep looking outside Assert.Multiple

### LINKED ISSUE #436: Using Has.Count in `Assert.Multiple` re-raises nullable reference warning
I recently apply a quickfix to use `Has.Count.EqualTo(1)` instead of the classical `Count, Is.EqualTo(1)`.

What I noticed afterwards is that some (to me suspicious) CS8604 are raised for the collection variable that also has an explicit `It.Not.Null` check. It only happens if the `Has.Count` is used in an `Assert.Multiple` block.

I crafted an example where it failes and two examples which compile file without any warning.

```
[Test]
public void HasCount_CreatesSuspiciousNullRefWarning()
{
    List<int>? maybeNull = null;
    Assert.That(maybeNull, Is.Not.Null);
    Assert.Multiple(() =>
    {
        Assert.That(maybeNull, Has.Count.EqualTo(2));
        Assert.That(maybeNull.Select(s => s), Is.All.EqualTo(1)); // CS8604 here, but should not!
    });
}

[Test]
public void FineWithoutHasCount()
{
    List<int>? maybeNull = null;
    Assert.That(maybeNull, Is.Not.Null);
    Assert.Multiple(() =>
    {
        Assert.That(maybeNull.Select(s => s), Is.All.EqualTo(1)); // Ok
    });
}

[Test]
public void FineWithHasCountOutsideOfAssertMultiple()
{
    List<int>? maybeNull = null;
    Assert.That(maybeNull, Is.Not.Null);
    Assert.That(maybeNull, Has.Count.EqualTo(2));
    Assert.Multiple(() =>
    {
        Assert.That(maybeNull.Select(s => s), Is.All.EqualTo(1)); // Ok
    });
}
```

Building for net5 using .net6 sdk, TestSdk 17.1, NUnit 3.13.3 and Analyzers 3.3.0.

manfred-brands:
for other reason to suppress.

Fixes #436

The problem was not the Has.Count.
This was considered a suppressible line, but as it was inside the Assert.Multiple it was ignored.
It should have looked further to find the Is.Not.Null.

### DIFF
--- a/src/nunit.analyzers.tests/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressorTests.cs
+++ b/src/nunit.analyzers.tests/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressorTests.cs
@@ -616,5 +616,31 @@ namespace NUnit.Analyzers.Tests.DiagnosticSuppressors
             await DiagnosticsSuppressorAnalyzer.EnsureNotSuppressed(suppressor, testCode)
                                                .ConfigureAwait(false);
         }
+
+        [Test]
+        public async Task TestIssue436()
+        {
+            // Test is changed from actual issue by replacing the .Select with an [1].
+            // The original code would not give null reference issues on the .Select for the .NET Framework
+            // because System.Linq is not annotated and therefore the compiler doesn't know null is not allowed.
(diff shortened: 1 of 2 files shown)


mikkelbu (approved):
LGTM and I find the new code a bit easier to read than the original code, so great work @manfred-brands. I'll merge it now

PR #445, opened 2022-04-08
Do not generate 'Is.Empty' code fix for types that are not enumerable.

### LINKED ISSUE #444: Anaylzer conversion for class with Count property can be incorrect
Example code

```csharp
using NUnit.Framework;

namespace Tests
{
    internal sealed class UseHas
    {
        [Test]
        public void UseHasCount()
        {
            MyClass myClass = new();

            myClass.Count = 1;
            Assert.That(myClass.Count, Is.EqualTo(1));
            myClass.Count--;
            Assert.That(myClass.Count, Is.EqualTo(0));
        }

        private sealed class MyClass
        {
            public int Count { get; set; }
        }
    }
}
```
Gets converted to

```csharp
using NUnit.Framework;

namespace Tests
{
    internal sealed class UseHas
    {
        [Test]
        public void UseHasCount()
        {
            MyClass myClass = new();

            myClass.Count = 1;
            Assert.That(myClass, Has.Count.EqualTo(1));
            myClass.Count--;
            Assert.That(myClass, Is.Empty);
        }

        private sealed class MyClass
        {
            public int Count { get; set; }
        }
    }
}
```
Obviously the above then throws at run time, because 'MyClass' is not enumerable. Ideally the analyzer should not suggest changes where the type requires a specific implementation that it does not support.

manfred-brands:
Fixes #444

### DIFF
--- a/src/nunit.analyzers.tests/UseCollectionConstraint/UseCollectionConstraintCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/UseCollectionConstraint/UseCollectionConstraintCodeFixTests.cs
@@ -112,5 +112,48 @@ namespace NUnit.Analyzers.Tests.UseCollectionConstraint
         }", "using System.Collections.Generic;");
             RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode);
         }
+
+        [Test]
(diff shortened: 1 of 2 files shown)


mikkelbu (approved):
LGTM. Thanks for the PR @manfred-brands

PR #449, opened 2022-04-12
NonNullableFieldOrPropertyIsUninitializedSuppressor:

### LINKED ISSUE #448: Analyzer doesn't detect non-nullable field set in overridden SetUp method
Historically some of our classes have virtual `SetUp` methods that are overridden in derived classes.
The derived class doesn't have the `SetUp` attribute, but the base class definition has.

The rule should look up the chain to see if the method that it overrides has the attribute.

```csharp
public abstract class BaseClass
{
    [SetUp]
    public virtual void Initialize()
    {
    }
}

[TestFixture]
public sealed class TestClass : BaseClass
{
    private const string Default = "Default";

    private string _field;

    public override void Initialize()
    {
        _field = Default;
    }

    [Test]
    public void Test()
    {
        Assert.That(_field, Is.EqualTo(Default));
    }
}
```

manfred-brands:
Look at base class method definition for the SetUp attribute

Fixes #448


mikkelbu (approved):

PR #451, opened 2022-05-03
Mark build.sh as executable

eatdrinksleepcode:
To use the `build.sh` script on non-Windows systems it must be marked as executable.


mikkelbu (approved):
Looks good to me. Thanks for the contribution @eatdrinksleepcode 👍

PR #453, opened 2022-05-06
Support classic StringAssert in ConstActualValueUsage analyzer and code fix

eatdrinksleepcode:
This adds support for `StringAssert` to `ConstActualValueUsageAnalyzer` and `ConstActualValueUsageCodeFix`.

The analyzers already support "classic" assertions in many places, but only ones that are defined on `Assert`. I have an existing codebase that frequently uses `StringAssert` and the other classic Assert classes, and I recently found a case where a `StringAssert.Contains` call started failing incorrectly after a refactoring because the arguments were reversed. I have tested that this change correctly flags that case.

This change is narrowly focused on `StringAssert` and `ConstActualValueUsage`; but if the general approach is acceptable, I can add more classic Asserts to more analyzers. I am probably going to do it anyway to use on my own code.

Also, I noticed that the existing code fix only supports the classic assertions specifically related to equality; but I saw no reason not to support all of the asserts on `StringAssert`, as they all follow the same format.

### DIFF
--- a/src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
@@ -311,5 +311,162 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
 
             RoslynAssert.Valid(analyzer, testCode);
         }
+
+        [Test]
+        public void AnalyzeWhenLiteralArgumentIsProvidedForStringAssertContains()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                public void Test()
+                {
+                    string expected = ""exp"";
+                    StringAssert.Contains(expected, ↓""act"");
+                }");
+
+            RoslynAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenLiteralNamedArgumentIsProvidedForStringAssertContains()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                public void Test()
+                {
+                    string expected = ""exp"";
+                    StringAssert.Contains(actual: ↓""act"", expected: expected);
+                }");
+
+            RoslynAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenLocalConstArgumentIsProvidedForStringAssertContains()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                public void Test()
+                {
+                    const string actual = ""act"";
+                    string expected = ""exp"";
+                    StringAssert.Contains(expected, ↓actual);
+                }");
+
+            RoslynAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenLocalConstNamedArgumentIsProvidedForStringAssertContains()
+        {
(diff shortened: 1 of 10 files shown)


src/nunit.analyzers/BaseAssertionAnalyzer.cs
```diff
@@ -16,24 +23,19 @@ public override void Initialize(AnalysisContext context)
...
-                IsAssert(invocationOperation);
-        }
-
-        protected static bool IsAssert(IInvocationOperation invocationOperation)
+        protected bool IsAssert(IInvocationOperation invocationOperation)
         {
-            return invocationOperation.TargetMethod.ContainingType.IsAssert();
+            return this.includeClassic
+                ? invocationOperation.TargetMethod.ContainingType.IsAnyAssert()
+                : invocationOperation.TargetMethod.ContainingType.IsAssert();
```
manfred-brands, line 30: Instead of conditional code in the base class, I would prefer this to be a virtual method which is overridden only in the ConstActualValueUsageAnalyzer. It is a condition set in a constructor which doesn't "smell" right.
If more analyzers would need this, we can do this override in a ClassicBaseAssertionAnalyzer instead.

manfred-brands (commented):
Thanks @eatdrinksleepcode for your contribution.
Your additions are neat and clean.
I added one remark about the use of the 'includeClassic' field.

src/nunit.analyzers/BaseAssertionAnalyzer.cs
eatdrinksleepcode: Yeah, in this situation I always have the "carry an extra field in the instance vs the extra overhead of a virtual method call" debate with myself, and I rarely end up with a strong opinion about it. Will amend shortly.

src/nunit.analyzers/BaseAssertionAnalyzer.cs
```diff
@@ -16,24 +16,24 @@ public override void Initialize(AnalysisContext context)
...
-            return operation is IExpressionStatementOperation expressionOperation &&
-                expressionOperation.Operation is IInvocationOperation invocationOperation &&
-                IsAssert(invocationOperation);
+            return this.IncludeClassic()
+                ? invocationOperation.TargetMethod.ContainingType.IsAnyAssert()
+                : invocationOperation.TargetMethod.ContainingType.IsAssert();
         }
 
-        protected static bool IsAssert(IInvocationOperation invocationOperation)
+        protected virtual bool IncludeClassic()
```
manfred-brands, line 26: Not the `virtual` method I meant. I would like `IsAssert` to become virtual, so there is no 'if' at all.
The base class implementation calls the `IsAssert` and the derived override calls `IsAnyAssert`

eatdrinksleepcode: Oh ok, got it

manfred-brands (approved):
Thanks, looks good to me.

src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageAnalyzerTests.cs
```diff
@@ -311,5 +311,162 @@ public void Test()
...
+        }
+
+        [Test]
+        public void AnalyzeWhenStringEmptyArgumentIsProvidedForStringAssertContains()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+                public void Test()
+                {
+                    string actual = ""act"";
+                    StringAssert.Contains(actual, string.Empty);
```
mikkelbu, line 394:
```suggestion
                    StringAssert.Contains(actual, ↓string.Empty);
```

mikkelbu (approved):
Look good to me. Thanks for the contribution @eatdrinksleepcode and thanks for the review @manfred-brands 👍 . I'll just add one minor correction and then merge the PR

PR #460, opened 2022-05-25
Limit suggestion to use Has.Length/Count

### LINKED ISSUE #459: False positives for NUnit2046 on non-array-like types
I really like the `Has.Count` and `Has.Length` syntax, so I use it a lot.

In some projects, I have types that similar to the following example
```
class Pipe
{
    double Diameter { get; set; }

    double Length { get; set; }
}

Assert.That(pipe.Diameter, Is.EqualTo(1));
Assert.That(pipe.Length, Is.EqualTo(2)); // NUnit2046 Use Assert.That(<collection>, Has.Length.EqualTo(<value>)
```

I know it is not a big deal, but I was wondering if the analyzer could check for the actual type instead of raising the suggestion for all Asserts including a `Length` (or `Count`) property. Not all types having such a property are actually Arrays, Lists or Collections :-)

manfred-brands:
Fixes #459 

- Do not raise it if the type is not integral
- Do not raise it if the type does not implement IEnumerable

### DIFF
--- a/src/nunit.analyzers.sln
+++ b/src/nunit.analyzers.sln
@@ -96,7 +96,6 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		..\.editorconfig = ..\.editorconfig
 		Directory.Build.props = Directory.Build.props
 		global.json = global.json
-		..\stylecop.json = ..\stylecop.json
 	EndProjectSection
 EndProject
 Global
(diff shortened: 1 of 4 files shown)


mikkelbu (commented):
Looks good. Is there any reason for removing the `VerifyNotUsingEmptyOnNonEnumerable` test ?

manfred-brands:
> Looks good. Is there any reason for removing the `VerifyNotUsingEmptyOnNonEnumerable` test ?

Yes, as it doesn't suggest to even use `Has.Length` anymore if the type is not `IEnumerable`.

mikkelbu:
> > Looks good. Is there any reason for removing the `VerifyNotUsingEmptyOnNonEnumerable` test ?
> 
> Yes, as it doesn't suggest to even use `Has.Length` anymore if the type is not `IEnumerable`.

Sorry :). I were too tired (and distracted by other work) when I reviewed this, so I didn't think. I'll be merging the change now. Thanks for the work 👍

mikkelbu (approved):

PR #476, opened 2022-09-11
[add] GitHub Actions workflow for running test

get-me-power:
I wrote a patch to run nunit.analyzers's tests on GitHub Actions. 
This enables automated testing with CI.

see also:
CI's log
  - https://github.com/get-me-power/nunit.analyzers/actions/runs/3032762761

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,19 @@
+name: CI
+
+on:
+  push:
+    branches:
+    - master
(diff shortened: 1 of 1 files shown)


mikkelbu (approved):
Thanks for the contribution @get-me-power . I've been a little unsure about whether we should do this or not (and then I completely forgot about the PR), but I see now harm in adding this, so I'll be merging it now

PR #477, opened 2022-09-11
Do not raise diagnostics when using == etc on 'ref struct'

### LINKED ISSUE #475: Code Fix for NUnit2010 on Ref Structs Creates CS1503
This might be similar to an issue [I just submitted](https://github.com/nunit/nunit.analyzers/issues/474)

Write this test:

```csharp
[Test]
public static void EqualityWithSpans()
{
  var span1 = new[] { 1 }.AsSpan();
  var span2 = new[] { 1 }.AsSpan();

  Assert.That(span1 == span2, Is.True);
}
```

This triggers NUnit2010. The code fix creates this code:

```csharp
[Test]
public static void EqualityWithSpans()
{
  var span1 = new[] { 1 }.AsSpan();
  var span2 = new[] { 1 }.AsSpan();

  Assert.That(span1, Is.EqualTo(span2));
}
```

But then this raises the CS1503 compiler error.

manfred-brands:
Fixes #475

### DIFF
--- a/src/nunit.analyzers.tests/ConstraintsUsage/ComparisonConstraintUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/ConstraintsUsage/ComparisonConstraintUsageAnalyzerTests.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Globalization;
 using Gu.Roslyn.Asserts;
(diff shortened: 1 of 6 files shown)


mikkelbu (approved):
Looks great. Thanks for the work @manfred-brands

PR #498, opened 2022-12-18
Issue 2045 assert.multiple

### LINKED ISSUE #496: NUnit2045 code-fix does not correctly lift asynchronous assertions into Assert.Multiple
I have NUnit2045 (Call independent assert statements from inside an Assert.Multiple) triggering on the following code:

```csharp
Assert.That(response.StatusCode, Is.EqualTo(HttpStatusCode.OK));
Assert.That(response.Content.Headers.ContentType?.MediaType, Is.EqualTo("application/json"));
Assert.That(response.Content.Headers.ContentType?.CharSet, Is.EqualTo("utf-8"));
Assert.That(await response.Content.ReadAsStringAsync(), Is.EqualTo(responseBody));
```

If I apply the code-fix, I am left with the following invalid code:

```csharp
Assert.Multiple(() =>
{
	Assert.That(response.StatusCode, Is.EqualTo(HttpStatusCode.OK));
	Assert.That(response.Content.Headers.ContentType?.MediaType, Is.EqualTo("application/json"));
	Assert.That(response.Content.Headers.ContentType?.CharSet, Is.EqualTo("utf-8"));
	Assert.That(await response.Content.ReadAsStringAsync(), Is.EqualTo(responseBody));
});
```

This code binds to `Assert.Multiple(TestDelegate)` which causes the following error:

> CS4034: The 'await' operator can only be used within an async lambda expression. Consider marking this lambda expression with the 'async' modifier.

I would expect that we should use the `Assert.Multiple(AsyncTestDelegate)` overload by adding `async` to the lambda, like so:

```csharp
Assert.Multiple(async () =>
{
	Assert.That(response.StatusCode, Is.EqualTo(HttpStatusCode.OK));
	Assert.That(response.Content.Headers.ContentType?.MediaType, Is.EqualTo("application/json"));
	Assert.That(response.Content.Headers.ContentType?.CharSet, Is.EqualTo("utf-8"));
	Assert.That(await response.Content.ReadAsStringAsync(), Is.EqualTo(responseBody));
});
```

manfred-brands:
Fixes #496

### DIFF
--- a/src/nunit.analyzers.tests/UseAssertMultiple/UseAssertMultipleCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/UseAssertMultiple/UseAssertMultipleCodeFixTests.cs
@@ -89,5 +89,50 @@ namespace NUnit.Analyzers.Tests.UseAssertMultiple
         }");
             RoslynAssert.FixAll(analyzer, fix, expectedDiagnostic, code, fixedCode);
         }
+
+        [Test]
+        public void AddsAsyncWhenAwaitIsUsed()
+        {
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void Test()
+        {
+            var configuration = new Configuration();
+            Assert.That(configuration, Is.Not.Null);
+            ↓Assert.That(configuration.Value1, Is.EqualTo(0));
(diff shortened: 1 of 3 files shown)


mikkelbu (approved):
Looks greate. A nice and simple solution. I'll merge it now

PR #501, opened 2023-01-14
Detect TupleDeconstruction assignment for uninitialized suppression

### LINKED ISSUE #499: DiagnosticSuppressor doesn't detect tuple deconstruction assignments
The diagnostic suppressor detects assignments in `SetUp` methods.
It doesn't however detect tuple deconstruction assignments:

```csharp
(m_Name, m_Description) = CreateInstances()
```

It does work if manually deconstructed:

```csharp
(string Name, string Description) instances = CreateInstances();
m_Name = instances.Name;
m_Description= instances.Description;
```

We need to detect tuple deconstruction assignments.

manfred-brands:
Fixes #499


mikkelbu (approved):
LGTM. Thanks for the report and the fix @manfred-brands 👍

PR #502, opened 2023-01-14
Issue442 test case source parameter check

### LINKED ISSUE #442: Analyzer for TestCaseSource does not check Test method parameters
The `TestCaseUsageAnalyzer` warns if the test method doesn't have the same number of parameters.
The `TestCaseSourceUsesStringAnalyzer` only test if the parameters for the test source method match.

Either create a new or add to the existing `TestCaseSourceUsesStringAnalyzer` code to check test parameters.

```csharp
private static readonly IEnumerable<string> NUnitNameSpaces = new[] { ".NUnit", ".NUnitExtensions" };

[TestCaseSource(nameof(NUnitNameSpaces))]
public void IsNUnit()
{
}
```

This should raise an issue that the test method doesn't expect parameters.

manfred-brands:
Fixes #442 

Only supports TestCaseSource with single parameter, where the TestCaseSource returns a non-object type enumeration.

Without analyzing the actual TestSource method/property, we cannot verify type/parameters as everything is either `object` or `TestData`.
The actual code might be an array or arrays of `object` or a `yield return new object[] { ... }`.
Probably to complex and expensive to check.

### DIFF
--- a/documentation/NUnit1030.md
+++ b/documentation/NUnit1030.md
@@ -18,7 +18,7 @@ Note that the current implementation only works for single parameters.
 
 ## Motivation
 
-A `TestCaseSourceAttribute` is used to pass parameters to a test method, but the test method does expect different type of parameter.
+A `TestCaseSourceAttribute` is used to pass parameters to a test method, but the test method expects a different type of parameter.
 
 ```charp
 private static readonly IEnumerable<string> NUnitNameSpaces = new[] { ".NUnit", ".NUnitExtensions" };
--- a/documentation/NUnit1030.md
+++ b/documentation/NUnit1030.md
@@ -1,0 +1,74 @@
+# NUnit1030
+
+## The type of parameter provided by the TestCaseSource does not match the type of the parameter in the Test method
+
+| Topic    | Value
+| :--      | :--
+| Id       | NUnit1030
+| Severity | Error
+| Enabled  | True
+| Category | Structure
+| Code     | [TestCaseSourceUsesStringAnalyzer](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs)
+
+## Description
+
+The type of parameters provided by the TestCaseSource must match the type of parameters in the Test method.
+
+Note that the current implementation only works for single parameters.
+
+## Motivation
+
+A `TestCaseSourceAttribute` is used to pass parameters to a test method, but the test method does expect different type of parameter.
+
+```charp
+private static readonly IEnumerable<string> NUnitNameSpaces = new[] { ".NUnit", ".NUnitExtensions" };
+
+[TestCaseSource(nameof(NUnitNameSpaces))]
+public void IsNUnit(int n)
+{
+}
+```
+
+## How to fix violations
+
+Match the type of parameters between the test data and the test method.
+
+
+<!-- start generated config severity -->
+## Configure severity
+
+### Via ruleset file
+
+Configure the severity per project, for more info see [MSDN](https://msdn.microsoft.com/en-us/library/dd264949.aspx).
+
+### Via .editorconfig file
+
+```ini
+# NUnit1030: The type of parameter provided by the TestCaseSource does not match the type of the parameter in the Test method
+dotnet_diagnostic.NUnit1030.severity = chosenSeverity
+```
+
+where `chosenSeverity` can be one of `none`, `silent`, `suggestion`, `warning`, or `error`.
+
+### Via #pragma directive
+
+```csharp
+#pragma warning disable NUnit1030 // The type of parameter provided by the TestCaseSource does not match the type of the parameter in the Test method
+Code violating the rule here
+#pragma warning restore NUnit1030 // The type of parameter provided by the TestCaseSource does not match the type of the parameter in the Test method
+```
+
+Or put this at the top of the file to disable all instances.
+
+```csharp
+#pragma warning disable NUnit1030 // The type of parameter provided by the TestCaseSource does not match the type of the parameter in the Test method
+```
+
+### Via attribute `[SuppressMessage]`
+
+```csharp
+[System.Diagnostics.CodeAnalysis.SuppressMessage("Structure",
+    "NUnit1030:The type of parameter provided by the TestCaseSource does not match the type of the parameter in the Test method",
+    Justification = "Reason...")]
+```
+<!-- end generated config severity -->
--- a/src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
@@ -549,7 +549,7 @@ namespace NUnit.Analyzers.Tests.TestCaseSourceUsage
         }
 
         [Test]
-        public void AnalyzeWhenNumberOfParametersOfTestIsLessThanEvidentFromTestSource()
+        public void AnalyzeWhenNumberOfParametersOfTestIsLessThanProvidedByTestCaseSource()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
     [TestFixture]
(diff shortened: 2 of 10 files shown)


documentation/NUnit1030.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Description
+
+The type of parameters provided by the TestCaseSource must match the type of parameters in the Test method.
+
+Note that the current implementation only works for single parameters.
+
+## Motivation
+
+A `TestCaseSourceAttribute` is used to pass parameters to a test method, but the test method does expect different type of parameter.
```
mikkelbu, line 21:
```suggestion
A `TestCaseSourceAttribute` is used to pass parameters to a test method, but the test method expects a different type of parameter.
```

src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
```diff
@@ -547,5 +547,114 @@ public class InnerClass
             var message = "Consider using nameof(AnotherClass.InnerClass.TestCases) instead of \"TestCases\"";
             RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic.WithMessage(message), testCode, fixedCode);
         }
+
+        [Test]
+        public void AnalyzeWhenNumberOfParametersOfTestIsLessThanEvidentFromTestSource()
```
mikkelbu, line 552:
```suggestion
        public void AnalyzeWhenNumberOfParametersOfTestIsLessThanProvidedByTestCaseSource()
```

src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
```diff
@@ -547,5 +547,114 @@ public class InnerClass
...
+    }", additionalUsings: "using System.Collections.Generic;");
+
+            var expectedDiagnostic = ExpectedDiagnostic
+                .Create(AnalyzerIdentifiers.TestCaseSourceMismatchInNumberOfTestMethodParameters)
+                .WithMessage("The TestCaseSource provides '1' parameter(s), but the Test method expects '0' parameter(s)");
+            RoslynAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenNumberOfParametersOfTestIsMoreThanEvidentFromTestSource()
```
mikkelbu, line 580:
```suggestion
        public void AnalyzeWhenNumberOfParametersOfTestIsMoreThanProvidedByTestCaseSource()
```

src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
```diff
@@ -547,5 +547,114 @@ public class InnerClass
...
+    }", additionalUsings: "using System.Collections.Generic;");
+
+            var expectedDiagnostic = ExpectedDiagnostic
+                .Create(AnalyzerIdentifiers.TestCaseSourceMismatchInNumberOfTestMethodParameters)
+                .WithMessage("The TestCaseSource provides '1' parameter(s), but the Test method expects '2' parameter(s)");
+            RoslynAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenParameterTypeOfTestDiffersFromTestSource()
```
mikkelbu, line 607:
```suggestion
        public void AnalyzeWhenParameterTypeOfTestDiffersFromTestCaseSource()
```

src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
```diff
@@ -547,5 +547,114 @@ public class InnerClass
...
+            yield return new object[] { 1, 2 };
+            yield return new object[] { 2, 3 };
+        }
+    }", additionalUsings: "using System.Collections.Generic;");
+
+            var expectedDiagnostic = ExpectedDiagnostic
+                .Create(AnalyzerIdentifiers.TestCaseSourceMismatchInNumberOfTestMethodParameters)
+                .WithMessage("The TestCaseSource provides '>0' parameter(s), but the Test method expects '1' parameter(s)");
+            RoslynAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
+        }
```
mikkelbu, line 658: Shouldn't we have a couple of tests that checks that the Analyzers don't report a problem when the return type from the TestCaseSource is one of the following
* The non-generic IEnumerable
* A collection of objects
* A collection of TestCaseData (or perhaps all classes deriving from `TestCaseParameters`)
(and the test expects more than 0 parameters)

As we cannot determine if there is a problem or not when returning these types.

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
```diff
@@ -192,20 +214,80 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context)
...
             {
                 context.ReportDiagnostic(Diagnostic.Create(
                     sourceDoesNotReturnIEnumerable,
                     syntaxNode.GetLocation(),
                     typeSymbol));
+                return null;
+            }
+            else
+            {
+                return elementType ?? context.Compilation.GetSpecialType(SpecialType.System_Object);
```
mikkelbu, line 290: Why do we default to `Object` here and not `null` (e.g. just removing the null coalescing) ?

mikkelbu (approved):
Greate work again @manfred-brands. I've only added some minor comments and questions. I'm happy to merge when you are.

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
manfred-brands: This deals with the case where the type is the non-generic `IEnumerable` which iterates over `object`. When `elementType is null` it means the type is not enumerable at all.

It might have been better if this was dealt with in the `IsEnumerable` method, but that is an existing method with several usages.  The current implementation of that method does allows distinguishing between `IEnumerable` and `IEnumerable<object>`.  That fact is used in some of the usages of the method.

src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
manfred-brands: @mikkelbu I added a few more tests.

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
mikkelbu: Thanks for the explanation. I think that the solution make sense 👍

src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
```diff
@@ -547,5 +546,158 @@ public class InnerClass
...
+
+            // The TestCaseSource provides '>0' parameter(s), but the Test method expects '1' parameter(s)
+            // Analyzing the actual code inside the TestCaseSource method is beyond the scope of the analyzer.
+            RoslynAssert.Valid(analyzer, testCode);
+        }
+
+        [TestCase("IEnumerable", "object", "System.Collections")]
+        [TestCase("IEnumerable<object>", "object", "System.Collections.Generic")]
+        [TestCase("IEnumerable<TestCaseData>", "TestCaseData", "System.Collections.Generic")]
+        [TestCase("IEnumerable<int>", "int", "System.Collections.Generic")]
```
mikkelbu, line 661: I guess the type can be determined here, but I think it is fine to keep the case

mikkelbu (approved):
Thanks for all the work @manfred-brands. I'm happy to merge this if you are.

src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
manfred-brands: Correct.

manfred-brands:
@mikkelbu Yes, this can be merged.

PR #505, opened 2023-01-21
Detect path for conditional/coalescing expression for null suppression

### LINKED ISSUE #503: DiagnosticsSuppressor doesn't suppress null coalescing operator
Given the following:

```csharp
string? line1 = reader.ReadLine();
string? line2 = reader.ReadLine();

Assert.That(line1, Is.Not.Null);
string fatal = line2 ?? line1;
```

This results in an CS8600 error:
image

```csharp
Assert.That(line1, Is.Not.Null);
string fatal = line2 is not null ? line2 : line1;
```

However the below does suppress the error correctly:
```csharp
Assert.That(line1, Is.Not.Null);
string fatal = line1;
```

manfred-brands:
Fixes #503 

The compiler raises the error on the whole statement, not on the part that is possibly `null`.
Simple for coalescing operator as we need to check the right hand side.
Harder for conditional operator. Code tries to detect the side, but only for simple expressions against `null`.

This PR also contains a fix for when the if/else parts are not contained in a block.

### DIFF
--- a/src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs
+++ b/src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs
@@ -3,9 +3,9 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 using NUnit.Analyzers.Constants;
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs
```diff
@@ -61,38 +61,116 @@ private static bool ShouldBeSuppressed(SyntaxNode node)
...
                     if (IsKnownToBeNotNull(currentNode) ||
-                        (parent is not null && IsValidatedNotNullByPreviousStatementInSameBlock(possibleNullReference, parent, statement)))
+                        (block is not null && IsValidatedNotNullByPreviousStatementInSameBlock(possibleNullReference, block, statement)))
                     {
                         return true;
                     }
                 }
 
-                currentNode = parent;
+                currentNode = statement.Parent;
```
mikkelbu, line 106: Is there a reason that we don't use `parent` instead of `statement.Parent`

src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs
```diff
@@ -61,38 +61,116 @@ private static bool ShouldBeSuppressed(SyntaxNode node)
...
+                    isNot = true;
+                    pattern = unaryPattern.Pattern;
+                }
+
+                if (pattern is not ConstantPatternSyntax constantPattern || !constantPattern.Expression.IsKind(SyntaxKind.NullLiteralExpression))
+                {
+                    return null;
+                }
+
+                // identifier is (not) null
```
mikkelbu, line 146: I think this is a bit odd place for the comment. Should this have been on top of line 128?

src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs
```diff
@@ -61,38 +61,116 @@ private static bool ShouldBeSuppressed(SyntaxNode node)
...
+            }
+            else if (conditionalExpression.Condition is BinaryExpressionSyntax binaryExpression &&
+                binaryExpression.Right.IsKind(SyntaxKind.NullLiteralExpression) &&
+                binaryExpression.Left is IdentifierNameSyntax or MemberAccessExpressionSyntax)
+            {
+                testedAgainstNullExpression = binaryExpression.Left.ToString();
+                isNot = binaryExpression.IsKind(SyntaxKind.NotEqualsExpression);
+            }
+            else
```
mikkelbu, line 155: Should we also cater for the people having a C/C++ background of hence often place the `null` to the left - e.g. `string fatal = null != line2 ? line2 : line1;`. Although it is has been some time since I've seen this formatting used in C#

mikkelbu (approved):
Looks good @manfred-brands, and I'm happy to merge it now if you are. I've left a couple of comments, but nothing major.

Thanks for the contribution

src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs
manfred-brands: If they would have StyleCop enabled they would have been told SA1131: Constant values should appear on the right-hand side of comparisons. But yes, I have seen code like that :-( Added branch.

src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs
manfred-brands: I was indicating that if we get here, it was a 'is (not) null' expression and not an 'is some-other-constant'.
Updated comments.

src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs
manfred-brands: The `parent` variable no longer exists. It is now called `block` and when a statement is not inside a block (which is the use case that was previously not covered), `block` is now `null`. Previously it would keep looking to parents until it found a block.

Given:
```csharp
{
    Assert.That(possible-nullable-value, Is.Not.Null);
    if (condition)
       non-nullable-variable = possible-nullable-value;
}
```

The offending statement is the assignment operation.

The `parent` previously was the block containing the `if`. 
The method `IsValidatedNotNullByPreviousStatementInSameBlock` tries to find the assignment statement inside that block, but it isn't in there as a direct statement and therefore doesn't find it and hence doesn't look to previous statements for assertions.

Now the `block` is `null` and the code moves on to `statement.Parent` which is the `if` statement. It then find the block and in there that `if` statement and before that the `Assert`.

src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs
mikkelbu: My mistake I meant `block` and not `parent` in the question. Thanks for the thorough explanation, as far as I can tell it makes sense.

src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs
mikkelbu: My mistake I meant `block` and not `parent` in the question. Thanks for the thorough explanation, as far as I can tell it makes sense.

mikkelbu:
Thanks for the contribution @manfred-brands. I'll merge it now and then take a look at the other PR.

PR #511, opened 2023-01-28
Do not report errors on TestCase parameters types that are invalid.

### LINKED ISSUE #509: Giving TestCaseAttribute an explicit decimal for a parameter that is a decimal gives a Nunit 1001 error
I am working with nunit 3.13.3

This is the offending case
```
    [TestCase(3650,-0.000027m)]
    public void Test_CalculateDegradationRate(int numberOfCycles, decimal expectedDegradationRate)
    {
        Assert.That(WithinTolerance(_state.DegradationRate, expectedDegradationRate), Is.True);
    }
```
It gives the error if the decimal is explicitly declared as a decimal with either 'm' or 'M', but it works fine if no specifier is given or if a 'd' specifier is given.

image

I would of course expect that the TestCaseAttribute would at least take an explicit decimal.

### LINKED ISSUE #510: TestCaseAttribute gives a Nunit 1001 when declaring an explicit decimal where a decimal is needed.
I am working with nunit 3.13.3

This is the offending case
```
    [TestCase(3650,-0.000027m)]
    public void Test_CalculateDegradationRate(int numberOfCycles, decimal expectedDegradationRate)
    {
        Assert.That(WithinTolerance(_state.DegradationRate, expectedDegradationRate), Is.True);
    }
```
It gives the error if the decimal is explicitly declared as a decimal with either 'm' or 'M', but it works fine if no specifier is given or if a 'd' specifier is given.

image

I would of course expect that the TestCaseAttribute would at least take an explicit decimal.

manfred-brands:
Fixes #509
Fixes #510 

If the attribute argument has a type, but is marked as Error, do not report it.
The compiler already report the invalid argument.

### DIFF
--- a/src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
@@ -356,6 +356,18 @@ namespace NUnit.Analyzers.Tests.TestCaseUsage
         }
 
         [Test]
+        public void AnalyzeArgumentHasTypeNotAllowedInAttributes()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    class AnalyzeArgumentHasTypeNotAllowedInAttributes
+    {
+        [TestCase(1.0m)]
+        public void Test(decimal d) { }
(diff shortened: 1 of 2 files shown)


mikkelbu (approved):
Nice simple fix @manfred-brands

PR #513, opened 2023-02-02
chore: add macos, linux in ci

get-me-power:


manfred-brands:
Looks like the Microsoft.CodeAnalysis assembly tries to load SQLite and fails. But there is way to much output from normal succeeding tests.

src/nunit.analyzers.tests/nunit.analyzers.tests.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <RootNamespace>NUnit.Analyzers.Tests</RootNamespace>
-    <TargetFrameworks>net6.0;net462</TargetFrameworks>
+    <TargetFramework>net6.0</TargetFramework>
```
manfred-brands, line 4: This needs to be done conditionally on OS. We still want to test .net framework on Windows

get-me-power: I agree with your suggestion.
Therefore, I changed only ci.yml like this. 
[552b72c](https://github.com/nunit/nunit.analyzers/pull/513/commits/552b72c686da6de7b06c53efe9f757c615c0432a)

manfred-brands:
I'm currently traveling and only have access to my phone.
For coverage we should align with the PR currently underway for nunit.framework. Note that it is no longer necessary to include a NuGet package to output coverage files from dotnet test. I have to lookup what I did for my work.

.github/workflows/ci.yml
```diff
@@ -8,12 +8,28 @@ on:
...
         with:
           global-json-file: ./global.json
-      - name: dotnet test
-        run: dotnet test ./src/nunit.analyzers.tests/
+
+      - name: .NET6.0 Test
+        run: |
+            dotnet tool install --global dotnet-coverage
+            dotnet test --collect:"XPlat Code Coverage" -f net6.0 ./src/nunit.analyzers.tests/
+            dotnet-coverage merge -o merged.cobertura.xml -f cobertura -r coverage.cobertura.xml
```
manfred-brands, line 26: There is no need for the coverlet nuget package. [Dotnet can output cobertura natively.](https://github.com/microsoft/vstest/issues/2874#issuecomment-973994514).
You can call it like: `dotnet test --collect:"Code Coverage;Format=Cobertura"`

src/nunit.analyzers.tests/nunit.analyzers.tests.csproj
```diff
@@ -5,6 +5,10 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="coverlet.collector" Version="3.2.0">
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
+    </PackageReference>
```
manfred-brands, line 11: Not necessary anymore.

.octocov.yml
```diff
@@ -0,0 +1,21 @@
...
+  if: is_pull_request
+
+report:
+  if: is_default_branch
+  datastores:
+    - artifact://${GITHUB_REPOSITORY}
+
+diff:
+  datastores:
+    - artifact://${GITHUB_REPOSITORY}
```
manfred-brands, line 21: This needs align with the [nunit changes](https://github.com/nunit/nunit/pull/4286/) to use coverlet.

.github/workflows/ci.yml
```diff
@@ -8,12 +8,28 @@ on:
...
+
+      - name: .NET Framework Test
+        if: runner.os == 'Windows'
+        run: |
+            dotnet test -f net462 ./src/nunit.analyzers.tests/
+
+      - name: Post Coverage
+        uses: k1LoW/octocov-action@v0
+        if: runner.os == 'Linux'
```
manfred-brands, line 35: This needs align with the https://github.com/nunit/nunit/pull/4286/ to use coverlet.

PR #516, opened 2023-02-14
Replace link to ruleset docs

SeanKilleen:
I noticed as part of https://github.com/nunit/docs/pull/723 that the analyzers appear to have an older link.

The MSDN link listed now redirects to <https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2015/code-quality/walkthrough-configuring-and-using-a-custom-rule-set?view=vs-2015&redirectedfrom=MSDN>

I'm proposing using this newer link, which I think achieves the same purpose: https://learn.microsoft.com/en-us/visualstudio/code-quality/using-rule-sets-to-group-code-analysis-rules?view=vs-2022

The analyzer docs eventually flow into the NUnit docs so I figured I'd suggest the change here.

### DIFF
--- a/documentation/NUnit1001.md
+++ b/documentation/NUnit1001.md
@@ -65,7 +65,7 @@ public void SampleTest(int numberValue)
 
 ### Via ruleset file
 
-Configure the severity per project, for more info see [MSDN](https://msdn.microsoft.com/en-us/library/dd264949.aspx).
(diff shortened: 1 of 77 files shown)


manfred-brands (approved):
Thanks for doing this.

stevenaw (approved):

stevenaw:
Nice catch!

mikkelbu (approved):
Thanks for the PR Sean. I'll merge it now (and then update the two new documentation documents)

PR #524, opened 2023-02-22
TestCaseSource now ignore all classes derived from TestCaseParameters

### LINKED ISSUE #523: False positive for NUnit1030 with `TestCaseParameters`
## Description

I've encountered what I believe is a false positive with the recently ([3.6.0](https://github.com/nunit/nunit.analyzers/releases/tag/3.6.0)) added [NUnit1030](https://github.com/nunit/nunit.analyzers/blob/master/documentation/NUnit1030.md). At least the code does work fine when disabling the error, and the error also does not occur when switching out `TestCaseParameters` with `TestCaseData`.

## Expected behavior

No error is produced (though I'm not sure if `TestCaseParameters` are supposed to be used this way, since they're in namespace `NUnit.Framework.Internal`, and using sub type `TestCaseData` instead works just fine).

## Reproduction

I've created a [small demo project](https://github.com/ehonda/TestCaseParametersNUnit1030/tree/main) that reproduces the false positive. It's [README](https://github.com/ehonda/TestCaseParametersNUnit1030/blob/main/README.md) has sample output.

The gist is the following:

```csharp
public class Tests
{
    private static class TestCases
    {
        public static IEnumerable<TestCaseParameters> All => new[]
        {
            new TestCaseData(1), new TestCaseData(2)
        };
    }

    [TestCaseSource(typeof(TestCases), nameof(TestCases.All))]
    public void Test(int n)
    {
        Assert.That(n, Is.GreaterThan(0));
    }
}
```

This produces error NUnit1030, although it works fine if the error is disabled, or if `IEnumerable<TestCaseParameters>` is switched out for `IEnumerable<TestCaseData>`.

manfred-brands:
Fixes #523

### DIFF
--- a/src/nunit.analyzers.tests/Constants/NUnitFrameworkConstantsTests.cs
+++ b/src/nunit.analyzers.tests/Constants/NUnitFrameworkConstantsTests.cs
@@ -6,6 +6,7 @@ using NUnit.Analyzers.Constants;
 using NUnit.Framework;
 using NUnit.Framework.Constraints;
 using NUnit.Framework.Interfaces;
+using NUnit.Framework.Internal;
 
 namespace NUnit.Analyzers.Tests.Constants
 {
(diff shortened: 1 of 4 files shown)


mikkelbu (approved):
LGTM. A little odd that Roslyn don't have an inbuilt `IsOrDerivesFrom`, but the implementation is as expected.

PR #531, opened 2023-03-11
Markdown fixes

SeanKilleen:
The analyzer docs flow into the NUnit docs, and the NUnit docs showed some markdown linting issues (which I didn't catch because of https://github.com/nunit/docs/issues/746. 

This PR should fix those items at the source.

PS if you want the markdown linting built directly into the GitHub CI here, let me know and I can make that happen.

### DIFF
--- a/documentation/NUnit1030.md
+++ b/documentation/NUnit1030.md
@@ -33,7 +33,6 @@ public void IsNUnit(int n)
 
 Match the type of parameters between the test data and the test method.
 
-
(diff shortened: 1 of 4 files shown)


mikkelbu (approved):
Thanks for the PR @SeanKilleen.

mikkelbu:
I'll merge this now and when you have time, then it would be great to have the markdown linting built directly into the GitHub CI here.

PR #536, opened 2023-04-26
Suppress parameters that have been verified to be not-null

### LINKED ISSUE #535: DiagnosticSuppressor doesn't suppress values passed as arguments
If a nullable variable is asserted to be not null and then passed to a method that expects a non-nullable instance, the errors are not suppressed, even if the variable is assigned to a local variable first.

```csharp
[Test]
public void TestPassDirectly()
{
    string? possibleNull = GetNext();
    Assert.That(possibleNull, Is.Not.Null);
    AssertEmpty((string)possibleNull);
}

[Test]
public void TestPassViaLocal()
{
    string? possibleNull = GetNext();
    Assert.That(possibleNull, Is.Not.Null);
    string s = (string)possibleNull;
    AssertEmpty(s);
}

private static string? GetNext() => string.Empty;

private static void AssertEmpty(string s) => Assert.That(s, Is.EqualTo(string.Empty));
```

manfred-brands:
Fixes #535

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,7 +1,7 @@
 {
   "sdk": {
-    "version": "6.0.100",
-    "rollForward" : "latestPatch",
+    "version": "7.0.100",
+    "rollForward" : "latestFeature",
     "allowPrerelease": false
   }
 }
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -22,7 +22,7 @@
 
   <ItemGroup>
     <PackageReference Include="StyleCop.Analyzers.Unstable" Version="1.2.0.435" PrivateAssets="all" />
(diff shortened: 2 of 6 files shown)


mikkelbu (approved):
LGTM. Thanks for the work @manfred-brands. I'll see if I can get a release out this weekend

manfred-brands:
> I'll see if I can get a release out this weekend

@mikkelbu Please hold-off. I want to see if I can get some other tickets addressed first.

mikkelbu:
> > I'll see if I can get a release out this weekend
> 
> @mikkelbu Please hold-off. I want to see if I can get some other tickets addressed first.

@manfred-brands No problem 👍

PR #537, opened 2023-04-30
Within usage analyzer + tests + codefix

### LINKED ISSUE #344: Add a rule informing that .Within is not valid for non-numeric types.

manfred-brands:
Fixes #344 

Rebased original code from #362 from @Antash, fixed rebase issues.
Then addressed remaining issues raised by @mikkelbu

### DIFF
--- a/src/nunit.analyzers/WithinUsage/WithinUsageAnalyzer.cs
+++ b/src/nunit.analyzers/WithinUsage/WithinUsageAnalyzer.cs
@@ -76,11 +76,7 @@ namespace NUnit.Analyzers.WithinUsage
         private static bool IsTypeSupported(ITypeSymbol type, HashSet<ITypeSymbol>? checkedTypes = null)
         {
             // Protection against possible infinite recursion
-            // TODO Remove suppression when the below is released:
-            // https://github.com/dotnet/roslyn-analyzers/issues/4568
-#pragma warning disable RS1024 // Compare symbols correctly
             checkedTypes ??= new HashSet<ITypeSymbol>(SymbolEqualityComparer.Default);
-#pragma warning restore RS1024 // Compare symbols correctly
             if (!checkedTypes.Add(type))
                 return false;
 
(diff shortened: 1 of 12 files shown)


src/nunit.analyzers/WithinUsage/WithinUsageAnalyzer.cs
```diff
@@ -0,0 +1,160 @@
...
+
+                var withinSuffix = constraintPart.GetSuffix(NUnitFrameworkConstants.NameOfEqualConstraintWithin) as IInvocationOperation;
+
+                if (withinSuffix == null)
+                    continue;
+
+                var expectedType = constraintPart.GetExpectedArgument()?.Type;
+
+                if (expectedType == null || expectedType.TypeKind == TypeKind.Error)
+                    return;
```
mikkelbu, line 60:
```suggestion
                if (withinSuffix is null)
                    continue;

                var expectedType = constraintPart.GetExpectedArgument()?.Type;

                if (expectedType is null || expectedType.TypeKind == TypeKind.Error)
                    return;
```

mikkelbu (approved):
Thanks for rebasing and adding the last part. I only have minor nit-pick (probably due to the age of the original change).

src/nunit.analyzers/WithinUsage/WithinUsageAnalyzer.cs
manfred-brands: I'm so relying on analyzers helping me to create consistent code, but here it failed me.

Even after enabling `dotnet_style_prefer_is_null_check_over_reference_equality_method = true:warning` I get none.
I hoped this would triggered [IDE0041](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0041) but it doesn't. It only triggers when using `ReferenceEquals(expectedType, null)`.
I see that the roslyn issue to [discourage ' == null'](https://github.com/dotnet/roslyn/issues/56859) is still open after two years. But the original reports also had created its own, so I added 'CSharpIsNullAnalyzer' to the project.

PR #538, opened 2023-04-30
Keep leading and trailing trivia when replacing statements with Assert.Multiple

### LINKED ISSUE #534: QuickFix for Assert.Multiple looses white space before and comments after
Given:
```csharp
... some statement

Assert.That(a, ...);
Assert.That(b, ...);
// Some comment
```

The `Assert.Multiple` quickfix converts this into:
```csharp
... some statement
Assert.Multiple(() =>
{
    Assert.That(a, ...);
    Assert.That(b, ...);
});
```

manfred-brands:
Fixes #534 

The empty line before the first Assert that can be merged is part of that statement. It needs to be moved to the new Assert.Multiple call.

Comment at the end of the block are part of block's closing brace, maintain that to keep the comments.

### DIFF
--- a/src/nunit.analyzers.tests/UseAssertMultiple/UseAssertMultipleCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/UseAssertMultiple/UseAssertMultipleCodeFixTests.cs
@@ -134,5 +134,63 @@ namespace NUnit.Analyzers.Tests.UseAssertMultiple
         }");
             RoslynAssert.FixAll(analyzer, fix, expectedDiagnostic, code, fixedCode);
         }
+
+        [TestCase("            // ", "")]
+        [TestCase("\r\n            // ", "")]
+        [TestCase("\r\n            ", "")]
+        [TestCase("\r\n            ", " // Same line Comment")]
+        [TestCase("\r\n            ", "\r\n            // Final Comment on next line")]
+        public void VerifyKeepsTrivia(string separation, string comment)
+        {
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@$"
+        public void TestMethod()
+        {{
+            const bool True = true;
+            const bool False = false;
+
+            // Verify that our bool constants are correct
+            ↓Assert.That(True, Is.True);
+            Assert.That(False, Is.False);
+{separation}Console.WriteLine(""Next Statement"");{comment}
+        }}");
+            var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@$"
+        public void TestMethod()
+        {{
+            const bool True = true;
+            const bool False = false;
+
+            Assert.Multiple(() =>
+            {{
+                // Verify that our bool constants are correct
+                Assert.That(True, Is.True);
+                Assert.That(False, Is.False);
+            }});
+{separation}Console.WriteLine(""Next Statement"");{comment}
+        }}");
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers/UseAssertMultiple/UseAssertMultipleCodeFix.cs
```diff
@@ -113,8 +129,21 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
                         })))).WithAdditionalAnnotations(Formatter.Annotation);
 
+            if (endOfLineTrivia is not null)
+            {
+                // Add the remembered blank line.
+                assertMultiple = assertMultiple.WithLeadingTrivia(endOfLineTrivia.Value);
+            }
+
+            // Add new line after the Assert.Multiple statement.
+            assertMultiple = assertMultiple.WithTrailingTrivia(SyntaxFactory.CarriageReturnLineFeed);
```
mikkelbu, line 139: Will this also work for codebases that use LF or do we need to either extract the current setting or use ElasticCarriageReturnLineFeed (I tried to read and understand dotnet/roslyn#202, dotnet/roslyn#19306, but I ended up more confused).

mikkelbu (approved):
Looks good. I only have a couple of questions

src/nunit.analyzers.tests/UseAssertMultiple/UseAssertMultipleCodeFixTests.cs
```diff
@@ -134,5 +134,63 @@ private sealed class Configuration
...
+            RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode);
+        }
+
+        [Test]
+        public void VerifyKeepsTrivia()
+        {
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@$"
+        public void TestMethod()
+        {{
+                // Verify that boolean work as expected
```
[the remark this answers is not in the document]
manfred-brands: Yes, to prevent getting trailing spaces when adding the extra new-line.

src/nunit.analyzers/UseAssertMultiple/UseAssertMultipleCodeFix.cs
manfred-brands: You an me both. I initially was looking for `Environment.NewLine` but that raised an issue that analyzers are not allowed to depend on environment variables. 

Using `ElasticCarriageReturnLineFeed` they were [removed](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.syntaxfactory.elasticcarriagereturnlinefeed?view=roslyn-dotnet-4.3.0): (**emphasis** mine)

> Elastic trivia are used to denote trivia that was not produced by parsing source text, and **are usually not preserved during formatting**."

According to a [comment](https://github.com/dotnet/roslyn/issues/19306#issuecomment-920139354) to the issue you referenced: (**emphasis** mine)

> Note that Roslyn implements a formatting pass after the application of a code fix or refactoring. During this pass, elastic nodes and **nodes marked with the formatter annotation** will be updated to use the correct line endings.

The `assertMultiple` statement is annotated with the `Formatter.Annotation`.

This seems to be correct as when updating the code to use `LineFeed`  the test still work and introduce CRLF.

Updated the code to add the CRLF before mentioning the Formatter to make it clearer.

PR #554, opened 2023-06-03
Add suppression of type parameters that have notnull constraints

### LINKED ISSUE #462: DiagnosticsSuppress does not suppress CS8634
Given the following constructed example:
```csharp
[Test]
public void Test()
{
    object? possibleNull = GetNext();
    Assert.NotNull(possibleNull);
    DoNothing((object)possibleNull);
}

private static object? GetNext() => new object();

private void DoNothing<T>(T obj)
    where T : class
{}
```
Result in error: 
```
CS8634 The type 'object?' cannot be used as type parameter 'T' in the generic type or method 'TestClass.DoNothing<T>(T)'. Nullability of type argument 'object?' doesn't match 'class' constraint.
```

Even trying to be explicit, doesn't remove the error:
```csharp
    object? possibleNull = GetNext();
    Assert.NotNull(possibleNull);
    object notNull = (object)possibleNull;
    DoNothing(notNull);
```

This might be a limitation on the `DiagnosticSuppressors` where we suppress the error generated for the cast, but we don't change the compiler's understanding of what could be null. We don't do a complete flow analysis on `notNull`

The only option that seems to work is by either using the null suppression operator: `object notNull = possibleNull!;`
or by doing the nullability conversion in a separate method:

```csharp
[Test]
public void Test()
{
    object? possibleNull = GetNext();
    object notNull = EnsureNotNull(possibleNull);
    DoNothing(notNull);
}

private static object? GetNext() => new object();

private static object EnsureNotNull(object? possibleNull)
{
    Assert.NotNull(possibleNull);
    return (object)possibleNull;
}

private void DoNothing<T>(T obj)
    where T : class
{}
```

manfred-brands:
Fixes #462

### DIFF
--- a/src/nunit.analyzers.tests/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressorTests.cs
+++ b/src/nunit.analyzers.tests/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressorTests.cs
@@ -1,6 +1,7 @@
 using System.IO;
 using System.Threading.Tasks;
 using Gu.Roslyn.Asserts;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using NUnit.Analyzers.DiagnosticSuppressors;
 using NUnit.Framework;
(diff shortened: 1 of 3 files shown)


src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs
```diff
@@ -115,6 +148,23 @@ private static bool ShouldBeSuppressed(SyntaxNode node)
             return false;
         }
 
+        private static bool ShouldBeSuppressed(IMethodSymbol method, string nonMatchedParameterType, InvocationExpressionSyntax invocationExpression)
+        {
+            // if (!Debugger.IsAttached) Debugger.Launch();
```
mikkelbu, line 153:
```suggestion
```

src/nunit.analyzers.tests/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressorTests.cs
```diff
@@ -771,5 +772,113 @@ public Lines(string? line1, string? line2)
...
+            ");
+
+            RoslynAssert.Suppressed(suppressor,
+                ExpectedDiagnostic.Create(DereferencePossiblyNullReferenceSuppressor.SuppressionDescriptors[diagnostic]),
+                testCode);
+        }
+
+        [TestCase("class", "CS8634")]
+        [TestCase("notnull", "CS8714")]
+        public void TestIssue462SuppressesMultiple(string constraint, string diagnostic)
```
mikkelbu, line 831: I originally had a question about this in the production code, but this test answered it 👍

mikkelbu (approved):
Only one minor nitpick (the tests answered the rest :) ). I'm happy to merge when you are 👍

src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs
manfred-brands: Removed

PR #555, opened 2023-06-06
Only report the first statement of the ones that can be combined in an Assert.Multiple

### LINKED ISSUE #541: [NUnit2045] Incorrect refactoring
```cs
using NUnit.Framework;

[TestFixture]
public class Class1
{
	[Test]
	public void TestValueHandler_Counted()
	{
		var handler = new Handler();

		Assert.That(handler.Build(null), Is.EqualTo(null));
		Assert.That(handler.Build(new object()), Is.EqualTo(0d));
		Assert.That(handler.Build(new { }), Is.EqualTo(0d));
	}
}

internal class Handler
{
	internal object? Build(object? value)
	{
		throw new NotImplementedException();
	}
}
```

When single refactoring applied, code refactored properly, but when user selects to apply it to some scope, e.g. Document, it produce non-compilable results. Note duplicate `Assert.Multiple(() =>` lines

image

```cs
using NUnit.Framework;

[TestFixture]
public class Class1
{
	[Test]
	public void TestValueHandler_Counted()
	{
		var handler = new Handler();
		Assert.Multiple(() =>
		{
			Assert.That(handler.Build(null), Is.EqualTo(null));
			Assert.Multiple(() =>
		{
			Assert.That(handler.Build(new object()), Is.EqualTo(0d));
			Assert.That(handler.Build(new { }), Is.EqualTo(0d));
		});
	}
}

internal class Handler
{
	internal object? Build(object? value)
	{
		throw new NotImplementedException();
	}
}
```

manfred-brands:
Fixes #541

### DIFF
--- a/src/nunit.analyzers.tests/UseAssertMultiple/UseAssertMultipleAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/UseAssertMultiple/UseAssertMultipleAnalyzerTests.cs
@@ -51,7 +51,7 @@ namespace NUnit.Analyzers.Tests.UseAssertMultiple
             var configuration = new Configuration();
             Assert.That(configuration, Is.Not.Null);
             ↓Assert.That(configuration.Value1, Is.EqualTo(0));
-            ↓Assert.That(configuration.Value2, Is.EqualTo(0.0));
+            Assert.That(configuration.Value2, Is.EqualTo(0.0));
             Assert.That(configuration.Value11, Is.EqualTo(string.Empty));
         }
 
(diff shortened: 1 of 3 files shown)


mikkelbu (approved):
Sorry it took some time to review this, but this week has been fully booked. The change looks sound (although I probably need to work with `UseAssertMultipleAnalyzer` to fully get the logic). I'm happy to merge this when you are (and again thanks for the work @manfred-brands).

Ps. Can you see issue with creating a release this weekend ?

manfred-brands:
Thanks @mikkelbu 
It should be fine to create a release now.
I won't be able to get to anything for a while.

PR #559, opened 2023-06-26
FP NUnit1001: CustomTypeConverters could convert from anything

Corniel:
I stumbled into this, as I use this feature quite often:

``` C#
[TestCase(23.00)]
pulic void TestSomething(Qowaiv.Financial.Amount amount)
{
  // ...
}
```

Where [`Qowaiv.Financial.Amount`](https://github.com/Qowaiv/Qowaiv#amount) is a value type that has its own type converter that not only can convert from string, but also from `int`, `long`, `double`, `float`, and `decimal`. This works like a charm, so it too bad that this falsely flags NUnit1001.

I've provided a fix, But if the implementation should be changed, just let me no.

Note that none existing test started to fail with this changed behavior.

### DIFF
--- a/src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
@@ -279,25 +279,6 @@ namespace NUnit.Analyzers.Tests.TestCaseUsage
         }
 
         [Test]
-        public void AnalyzeArgumentIsSomeConvertedToTypeWithCustomConverter()
-        {
-            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
-    using System.ComponentModel;
-
-    class C
-    {
-        [TestCase(17)]
-        public void Test(CustomType p) { }
-    }
-
-    [TypeConverter(typeof(CustomTypeConverter))]
-    struct CustomType { }
-    class CustomTypeConverter : TypeConverter { }");
-
-            RoslynAssert.Valid(this.analyzer, testCode);
-        }
-
-        [Test]
         public void AnalyzeArgumentIsNullConvertedToTypeWithCustomConverter()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
@@ -344,7 +325,7 @@ namespace NUnit.Analyzers.Tests.TestCaseUsage
 
     class C
     {
-        [TestCase(↓2)]
+        [TestCase(2)]
         public void Test(CustomType p) { }
     }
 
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers/Extensions/AttributeArgumentTypedConstantExtensions.cs
```diff
@@ -210,6 +210,10 @@ private static bool HasBuiltInImplicitConversion(ITypeSymbol argumentType, IType
                     }
                 }
             }
+            else if (targetTypeSymbol.GetAttributes().Any(att => att.AttributeClass?.GetFullMetadataName() == typeof(TypeConverterAttribute).FullName))
```
manfred-brands, line 213: I'm not behind a computer, but isn't TypeConverter not already dealt with?
You added a vase when targetType is null?
The CustomTypeConverter in your test doesn't override the Can convert from tested in the existing code.

Corniel: TypeConverters are only dealt with for strings. And the code tries to check for types it can access the code from. In most cases that is not possible. So basically, you can not know if it is supported or not, but because of that, you should report on it. Specially because the default severity is an error, and it runs perfectly.

manfred-brands: Thanks. I am on leave without access to a development PC until middle of July. I will look further then.

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -271,6 +271,25 @@ class C
...
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    using System.ComponentModel;
+
+    class C
+    {
+        [TestCase(17)]
+        public void Test(CustomType p) { }
+    }
+
```
manfred-brands: There is already a test for this case, but one that assumes failure.

src/nunit.analyzers/Extensions/AttributeArgumentTypedConstantExtensions.cs
manfred-brands: Sorry, I forgot to look at.
Your added case can be removed. Instead we can remove the condition on the line below.

manfred-brands (approved):
I added a commit to your branch simplifying the solution.
Please check if it works for your use case.

Corniel:
@manfred-brands I saw you made the changes yourself. Would you like me to make any changes in this PR?

manfred-brands:
> @manfred-brands I saw you made the changes yourself. Would you like me to make any changes in this PR?

No. Are you happy with the changes?

mikkelbu (approved):

PR #569, opened 2023-08-11
Added a suppressor when CA1812 fires on NUnit Test classes.

### LINKED ISSUE #565: [TestFixture] attribute suppresses `CA1812`
We try to keep our code clean so we've enabled optional analyzers including [CA1812: Avoid uninstantiated internal classes](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1812).

Whenever we have an internal class used as a test fixture, this rule always lights up giving us false positives.

Much like there's an analyzer to suppress nullability warnings when doing `Assert.That(x, Is.Not.Null);`, can we get an analyzer suppressor which disables this rule in the presence of `[TestFixture]`?

manfred-brands:
Fixes #565

### DIFF
--- a/documentation/NUnit3003.md
+++ b/documentation/NUnit3003.md
@@ -19,7 +19,7 @@ Class is a NUnit TestFixture and called by reflection
 The default roslyn analyzer has rule [CA1812](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1812)
 which warns about internal classes not being used.
 That analyzer doesn't know about NUnit test classes.
-This suppressor catches the error, verifies the class is an NUNit TestFixture and if so suppresses the error.
+This suppressor catches the error, verifies the class is an NUnit TestFixture and if so suppresses the error.
 
 <!-- start generated config severity -->
 ## Configure severity
--- a/documentation/NUnit3003.md
+++ b/documentation/NUnit3003.md
@@ -1,0 +1,63 @@
+# NUnit3003
+
+## Class is a NUnit TestFixture and called by reflection
+
+| Topic    | Value
+| :--      | :--
+| Id       | NUnit3003
+| Severity | Info
+| Enabled  | True
+| Category | Suppressor
+| Code     | [AvoidUninstantiatedInternalClassesSuppressor](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/DiagnosticSuppressors/AvoidUninstantiatedInternalClassesSuppressor.cs)
+
+## Description
+
+Class is a NUnit TestFixture and called by reflection
+
+## Motivation
+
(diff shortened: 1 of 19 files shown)


documentation/NUnit3003.md
```diff
@@ -0,0 +1,63 @@
...
+## Description
+
+Class is a NUnit TestFixture and called by reflection
+
+## Motivation
+
+The default roslyn analyzer has rule [CA1812](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1812)
+which warns about internal classes not being used.
+That analyzer doesn't know about NUnit test classes.
+This suppressor catches the error, verifies the class is an NUNit TestFixture and if so suppresses the error.
```
mikkelbu, line 22:
```suggestion
This suppressor catches the error, verifies the class is an NUnit TestFixture and if so suppresses the error.
```

src/nunit.analyzers.tests/DiagnosticSuppressors/AvoidUninstantiatedInternalClassesSuppressorTests.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        [OneTimeSetUp]
+        public void OneTimeSetUp()
+        {
+            // Find the NetAnalyzers assembly (note version should match the one referenced)
+            string netAnalyzersPath = Path.Combine(PathHelper.GetNuGetPackageDirectory(),
+                "microsoft.codeanalysis.netanalyzers/7.0.3/analyzers/dotnet/cs/Microsoft.CodeAnalysis.CSharp.NetAnalyzers.dll");
+            Assembly netAnalyzerAssembly = Assembly.LoadFrom(netAnalyzersPath);
+            Type analyzerType = netAnalyzerAssembly.GetType("Microsoft.CodeQuality.CSharp.Analyzers.Maintainability.CSharpAvoidUninstantiatedInternalClasses", true)!;
+            this.analyzer = (DiagnosticAnalyzer)Activator.CreateInstance(analyzerType!)!;
```
mikkelbu, line 65: Is the first `!` needed in `Activator.CreateInstance(analyzerType!)!;` given that we also have used `!` on the previous line (line 64).

mikkelbu (approved):
Looks good. I only have the smallest suggestion and question.

src/nunit.analyzers.tests/DiagnosticSuppressors/AvoidUninstantiatedInternalClassesSuppressorTests.cs
manfred-brands: You are correct. Removed.

documentation/NUnit3003.md
manfred-brands: Thanks. Updated.

PR #570, opened 2023-08-12
Create ValuesUsageAnalyzer.

### LINKED ISSUE #52: Reuse TestCaseAttribute logic for ValuesAttribute
We should reuse the logic for testing whether the value of a `TestCaseAttribute` matches the argument of the test method to also have effect for `ValuesAttribute`. The following will currently give an analyzer error on `"a"`.

```csharp
    [TestCase("a")]
    [TestCase(-4.0)]
    [TestCase(-9.0)]
    public void Test1(double d)
    {
...
    }
```
But for the following there are no analyzer errors
```csharp
    [Test]
    public void Test2([Values("a", -4.0, -9.0)] double d)
    {
...
    }
```

andrewimcclement:
Aiming to resolve #52 , though I haven't actually shared any code at this point.  Refactoring suggestions are very welcome.
All tests are duplicated from TestCaseUsageAnalyzerTests.  `AnalyzeWhenArgumentPassesNullToValueType` demonstrates `[Values(null)]` reporting a slightly unhelpful type to the user - fixes/suggestions are welcome.


src/nunit.analyzers/ValuesUsage/ValuesUsageAnalyzer.cs
```diff
@@ -0,0 +1,91 @@
...
+                symbolContext.CancellationToken.ThrowIfCancellationRequested();
+                for (var index = 0; index < attribute.ConstructorArguments.Length; index++)
+                {
+                    var constructorArgument = attribute.ConstructorArguments[index];
+
+                    // If the compiler detects an illegal constant, we shouldn't check it.
+                    // Unfortunately the constant 'null' is marked as Error with a null type.
+                    if (constructorArgument.Kind == TypedConstantKind.Error && constructorArgument.Type is not null)
+                    {
+                        continue;
```
manfred-brands, line 61: We should still check whether `null` is a legal argument for the parameter.
It would be legal if the parameter is not a ValueType (struct or int etc).
I can't recall if we can detect nullability, in which case `null` would only be valid if the parameter is marked nullable.

manfred-brands (changes requested):
Thanks. It is a good start.
Please add the documentation .md file so the tests can pass.
If you run the test locally you get  a skeleton file you can rename and update.

I'm not sure how much code is sharable between the two analyzers, except for the tests to see if a parameter can be assigned from the supplied value.
The analyzers are different that one has multiple values for one parameter, the other a single value for each parameter.

andrewimcclement:
@manfred-brands I used NUnit2047 as it was simply the next number in the list without skipping to the 3000s, but should this be in the 1000s? I'm not really sure of the overall scheme.

src/nunit.analyzers/ValuesUsage/ValuesUsageAnalyzer.cs
andrewimcclement: Huh, I just tried out
```csharp
[Test]
public void Blah([Values(null, 0)] int numberValue)
{
    Assert.That(numberValue, Is.EqualTo(0));
}
```
which passed.  So the behaviour for null is a little weird (seems to end up being `default(T)`).  I can't imagine we want to encourage this behaviour, but equally the analyzer claiming something will fail when it will run correctly isn't particularly helpful either.

This means that my current test `AnalyzeWhenArgumentTypeIsIncorrect` actually warns incorrectly, since that test would run without issues.  Amusingly enough, the TestCaseUsageAnalyzer has the same bug, in that null ends up being converted to `default(T)` so the test will actually run

documentation/NUnit2048.md
```diff
@@ -0,0 +1,101 @@
+# NUnit2048
```
andrewimcclement, line 1: Somehow I've ended up pushing with windows line endings (causing the appveyor build to fail), no idea how.

src/nunit.analyzers/ValuesUsage/ValuesUsageAnalyzer.cs
manfred-brands: >  I can't imagine we want to encourage this behaviour, but equally the analyzer claiming something will fail when it will run correctly isn't particularly helpful either.

Analyzers not only warn against failing code, but are also for improving. You could create a separate diagnostic saying that `null` should be `default(T)`.

andrewimcclement: For the general point, `CanAssignTo` does handle null conversions correctly (including nullable structs), so the logic is correct as far as I can tell.  (We do not check for nullability annotations in that method.)

andrewimcclement:
Now using NUnit1031 for consistency with other structural issues.

documentation/NUnit2048.md
manfred-brands: Check your git settings for `git core.autocrlf`
We should probably fix this in a .got attributes so it doesn't depend on a user's settings

andrewimcclement: Or not?  Everything seems happy now.  Not sure what happened.

documentation/NUnit1031.md
```diff
@@ -0,0 +1,101 @@
...
+{
+    Assert.That(numberValue, Is.AnyOf(0, 1));
+}
+```
+
+### Problem
+
+In the test above, `numberValue` is declared as an integer. However, `[Values(0.0, 1.0)]` provides values as doubles.  This will lead to a runtime failure.
+
+### Fix
```
manfred-brands, line 37:
```suggestion
## How to fix violations
```

documentation/NUnit1031.md
```diff
@@ -0,0 +1,101 @@
...
+## Description
+
+The individual arguments provided by a ValuesAttribute must match the type of the corresponding parameter of the method.
+
+## Motivation
+
+To prevent tests that will fail at runtime due to improper construction.
+
+## How to fix violations
+
```
manfred-brands, line 22:
```suggestion
```

documentation/NUnit1031.md
```diff
@@ -0,0 +1,101 @@
...
+// Both use type double
+[Test]
+public void SampleTest([Values(0.0, 1.0)] double numberValue)
+{
+    Assert.That(numberValue, Is.AnyOf(0, 1));
+}
+```
+
+ADD HOW TO FIX VIOLATIONS HERE
+
```
manfred-brands, line 63:
```suggestion
```

src/nunit.analyzers.tests/ValuesUsage/ValuesUsageAnalyzerTests.cs
```diff
@@ -0,0 +1,355 @@
...
+    {
+        [Test]
+        public void Test([Values(2)] CustomType p) { }
+    }
+
+    [TypeConverter(typeof(CustomTypeConverter))]
+    public struct CustomType { }
+    public class CustomTypeConverter : TypeConverter { }");
+
+            RoslynAssert.Diagnostics(this.analyzer,
```
manfred-brands, line 292: Note to self: This use case will become valid when #559 is checked in.

src/nunit.analyzers/ValuesUsage/ValuesUsageAnalyzer.cs
manfred-brands: We will leave that for another separate issue.

manfred-brands (changes requested):
Small suggested changes to documentation.

documentation/NUnit1031.md
andrewimcclement: I copied the style from NUnit1001.md (https://github.com/nunit/nunit.analyzers/blob/master/documentation/NUnit1001.md?plain=1).  I'll push changes in line with your suggestions, just fyi that we should probably update NUnit1001 (and/or others) when we have time.

src/nunit.analyzers/ValuesUsage/ValuesUsageAnalyzer.cs
```diff
@@ -0,0 +1,91 @@
...
+            foreach (var attribute in attributes.Where(x => x.ApplicationSyntaxReference is not null
+                                                       && SymbolEqualityComparer.Default.Equals(x.AttributeClass, valuesType)))
+            {
+                symbolContext.CancellationToken.ThrowIfCancellationRequested();
+                for (var index = 0; index < attribute.ConstructorArguments.Length; index++)
+                {
+                    var constructorArgument = attribute.ConstructorArguments[index];
+
+                    // If the compiler detects an illegal constant, we shouldn't check it.
+                    // Unfortunately the constant 'null' is marked as Error with a null type.
```
andrewimcclement, line 58: @manfred-brands This seems incorrect - when I try using `null`, I see it as Kind == Primitive and Type as object (or Kind == Array and Type == object?[], etc.).  I copied this from the relevant code for TestCaseUsageAnalyzer.  I can remove this comment and the check `constructorArgument.Type is not null` if you want?

documentation/NUnit1031.md
manfred-brands: Thanks, I reverted the changes so the two align.

src/nunit.analyzers.tests/ValuesUsage/ValuesUsageAnalyzerTests.cs
manfred-brands: Fixed test.

src/nunit.analyzers/ValuesUsage/ValuesUsageAnalyzer.cs
manfred-brands: @andrewimcclement I indeed don't get it in your analyzer.
It was added to `TestCaseUsageAnalyzer` for #510 and there I still get it for the tests with `[TestCase(null)]`
The test cases that still triggers this is: `AnalyzeWhenArgumentPassesNullToValueType`

image

The same test case in the `ValuesUsageAnalyzer` shows:
image

Investigating further, the root of the cause is the `AdjustArguments` call done in `TestCaseUsageAnalyzer`. 
This tries to reflect the case were the `TestCaseAttribute` constructor converts a `null` argument into `new object?[] { null }`.
Note that this is also the case for the `ValuesAttribute`.

I have converted the local  `AdjustArguments` into an extension method and call it from your analyzer.
I also added test cases to cover this scenario, although I don't see why anyone would. But it works in nunit.

Whilst I was there, I added support for nullable reference types with appropriate unit tests.

manfred-brands (approved):

manfred-brands:
@mikkelbu As I also made changes, I would appreciate your review when you have time.

mikkelbu:
@manfred-brands I'll take a look at it now. And thank you both for all the work

src/nunit.analyzers/Extensions/AttributeArgumentTypedConstantExtensions.cs
```diff
@@ -222,5 +230,36 @@ private static bool HasBuiltInImplicitConversion(ITypeSymbol argumentType, IType
...
+                && attributePositionalArguments[0] is { Kind: TypedConstantKind.Array } arrayArgument
+                && arrayArgument.Type is IArrayTypeSymbol arrayType
+                && arrayType.ElementType.SpecialType == SpecialType.System_Object)
+            {
+                // params overload, need to check array values instead
+
+                if (arrayArgument.IsNull)
+                {
+                    // If null is provided - analyze it as single null value, not as null array
+                    // Thebelow creates an invalid TypedConstant (Error, NullType)
```
mikkelbu, line 250:
```suggestion
                    // The code below creates an invalid TypedConstant (Error, NullType)
```

src/nunit.analyzers/ValuesUsage/ValuesUsageAnalyzer.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    [DiagnosticAnalyzer(LanguageNames.CSharp)]
+    public sealed class ValuesUsageAnalyzer : DiagnosticAnalyzer
+    {
+        private static readonly DiagnosticDescriptor parameterTypeMismatch = DiagnosticDescriptorCreator.Create(id: AnalyzerIdentifiers.ValuesParameterTypeMismatchUsage,
+                                                                                                                title: ValuesUsageAnalyzerConstants.ParameterTypeMismatchTitle,
+                                                                                                                messageFormat: ValuesUsageAnalyzerConstants.ParameterTypeMismatchMessage,
+                                                                                                                category: Categories.Structure,
+                                                                                                                defaultSeverity: DiagnosticSeverity.Error,
+                                                                                                                description: ValuesUsageAnalyzerConstants.ParameterTypeMismatchDescription);
```
mikkelbu, line 18: I'm often working on a small laptop, so cannot see all the code without scrolling

```suggestion
        private static readonly DiagnosticDescriptor parameterTypeMismatch = DiagnosticDescriptorCreator.Create(
            id: AnalyzerIdentifiers.ValuesParameterTypeMismatchUsage,
            title: ValuesUsageAnalyzerConstants.ParameterTypeMismatchTitle,
            messageFormat: ValuesUsageAnalyzerConstants.ParameterTypeMismatchMessage,
            category: Categories.Structure,
            defaultSeverity: DiagnosticSeverity.Error,
            description: ValuesUsageAnalyzerConstants.ParameterTypeMismatchDescription);
```

mikkelbu (approved):
Looks great. I only found a couple of nitpicks.

src/nunit.analyzers.sln
```diff
@@ -42,6 +42,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "documentation", "documentat
 		..\documentation\NUnit1028.md = ..\documentation\NUnit1028.md
 		..\documentation\NUnit1029.md = ..\documentation\NUnit1029.md
 		..\documentation\NUnit1030.md = ..\documentation\NUnit1030.md
+		..\documentation\NUnit1030.md = ..\documentation\NUnit1031.md
```
manfred-brands, line 45:
```suggestion
		..\documentation\NUnit1031.md = ..\documentation\NUnit1031.md
```

PR #581, opened 2023-09-05
Within is allowed on Nullable<T> when T  is supported.

### LINKED ISSUE #580: False positive for WithinUsageAnalyzer
We get `false` positives for when nullable numeric types are used:

```
[TestCase(null, null)]
[TestCase(1, 1)]
[TestCase(1.01, 1)]
public void Nullables(double? actual, double? expected)
{
    Assert.That(actual, Is.EqualTo(expected).Within(0.1));
}
```

If the values are not `null`, the `NumericsComparer` will use the _tolerance_.

Allow `Nullable<T>` where `T` itself is allowed.

manfred-brands:
Fixes #580


mikkelbu (approved):
LGTM. A nice and simple fix

PR #583, opened 2023-09-05
Recognize assignments inside try/finally statements.

### LINKED ISSUE #582: NonNullableFieldOrPropertyIsUninitializedSuppressor does not detect assignments in try/finally blocks

manfred-brands:
Fixes #582

### DIFF
--- a/src/nunit.analyzers.tests/DiagnosticSuppressors/NonNullableFieldOrPropertyIsUninitializedSuppressorTests.cs
+++ b/src/nunit.analyzers.tests/DiagnosticSuppressors/NonNullableFieldOrPropertyIsUninitializedSuppressorTests.cs
@@ -124,6 +124,51 @@ namespace NUnit.Analyzers.Tests.DiagnosticSuppressors
         }
 
         [Test]
+        public void FieldAssignedInTryFinally()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers.tests/DiagnosticSuppressors/NonNullableFieldOrPropertyIsUninitializedSuppressorTests.cs
```diff
@@ -123,6 +123,51 @@ public void Test()
             RoslynAssert.Suppressed(suppressor, testCode);
         }
 
+        [Test]
+        public void FieldAssignedInTryFinally()
```
mikkelbu, line 127: Should we rename this to `FieldAssignedInTry()` and then also have a test that tests when the assignment is in the `finally` clause?

mikkelbu (approved):
Just a comment, but I'm also happy to merge it as it is

src/nunit.analyzers.tests/DiagnosticSuppressors/NonNullableFieldOrPropertyIsUninitializedSuppressorTests.cs
manfred-brands: I have instead simplified the test and have fields assigned in both the `try` and the `finally` clauses.

PR #584, opened 2023-09-06
CanBeEqual: Add special case for Uri.Equals(string)

### LINKED ISSUE #561: NUnit2021 error shown when comparing Uri and string
If you try to `Assert` that a `Uri` is equal to a `string`, the NUnit Analyzer shows an NUnit2021 error:

> The EqualTo constraint always fails as the actual and the expected value cannot be equal csharp(NUnit2021)

But `Uri` must have some overloaded `Equals()` logic that lets you compare `Uri` and `string`:

```csharp
string testString = "test";
Uri testUri = new Uri("test", UriKind.Relative);

bool result = testUri.Equals(testString); // True
```

https://dotnetfiddle.net/seNXx9

I recently discovered this when I enabled the NUnit Analyzer in my test projects and discovered that the tests have always been passing when comparing two separate objects.

I know this is similar to #186

manfred-brands:
Fixes #561


mikkelbu (approved):
LGTM. Nice and simple (although a bit confusing behaviour by `Uri`)

PR #589, opened 2023-09-09
Descend into local asynchronous methods.

### LINKED ISSUE #585: NonNullableFieldOrPropertyIsUninitializedSuppressor doesn't check async methods called from SetUp
```csharp
public async Task SetUp()
{
    ...
    await MoreInitialization().ConfigureAwait(false);
    ....
}

private async Task MoreInitialization()
{
    ....
    m_Field = value;
    ....
}```

It currently follows non-async method, so it would only be fair to also follow async method calls.

manfred-brands:
Fixes #585


manfred-brands:
@mikkelbu This is the same as #586 where I had accidentally deleted the remote branch.

mikkelbu (approved):
LGTM 👍

PR #597, opened 2023-09-22
fix: IndexOutOfRangeException threw by WithinUsageAnalyzer

### LINKED ISSUE #596: WithinUsageAnalyzer threw an exception of type 'System.IndexOutOfRangeException'
`NUnit.Analyzers="3.7.0"`

### Minimal reproducible case

```c#
using System.Collections;
using NUnit.Framework;

Assert.That(new DoubleCollection(1), Is.EqualTo(new DoubleCollection(2)).Within(0.001));

public class DoubleCollection : IEnumerable<double>
{
    public DoubleCollection(params double[] numbers) { }
    public IEnumerator<double> GetEnumerator() => throw new NotImplementedException();
    IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
}
```

Error Message: `Analyzer 'NUnit.Analyzers.WithinUsage.WithinUsageAnalyzer' threw an exception of type 'System.IndexOutOfRangeException' with message 'Index was outside the bounds of the array.'.`

HenryZhang-ZHY:
fix #596

### DIFF
--- a/src/nunit.analyzers.tests/WithinUsage/WithinUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/WithinUsage/WithinUsageAnalyzerTests.cs
@@ -273,6 +273,31 @@ namespace NUnit.Analyzers.Tests.WithinUsage
             RoslynAssert.Valid(analyzer, testCode);
         }
 
+        [Test]
+        public void AnalyzeNonGenericSubclassOfGenericIEnumerable()
+        {
+            string testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+                 [TestFixture]
+                 public class TestClass
+                 {
+                     public void TestMethod()
+                     {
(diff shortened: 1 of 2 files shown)


HenryZhang-ZHY:
@dotnet-policy-service agree

HenryZhang-ZHY:
PTAL @manfred-brands

manfred-brands (approved):
Thank you @HenryZhang-ZHY for not only reporting this, but also supplying the simple fix

HenryZhang-ZHY:
Will this fix be released with version 3.8? If so, when is version 3.8 scheduled to be released?

mikkelbu:
> Which version number will include this fix? When will it be released? @manfred-brands

@HenryZhang-ZHY  I think I'll create a version 3.7.1 and release it today or tomorrow (I just want to get one change

PR #598, opened 2023-09-23
Update NUnit1032.md to fix typo `Dipose` -> `Dispose`.

andrewimcclement:


andrewimcclement:
By the way, out of curiosity, how often does the website https://docs.nunit.org/articles/nunit-analyzers get updated?  (I note that e.g. NUnit1031/NUnit1032 are missing - I realise that 3.7.0 has only just gone out.)

manfred-brands (approved):
Thanks

mikkelbu:
> By the way, out of curiosity, how often does the website https://docs.nunit.org/articles/nunit-analyzers get updated? (I note that e.g. NUnit1031/NUnit1032 are missing - I realise that 3.7.0 has only just gone out.)

I depends on how fast we are at creating a PR for the docs repo. Unfortunately, the process is manual at the moment, so sometimes it take some time.

PR #599, opened 2023-09-24
Do not suppress CA1001 if Fixture uses LifeCycle.InstancePerTestCase

### LINKED ISSUE #595: NUnit1032 false positive in `LifeCycle.InstancePerTestCase` test fixture using constructor
Hi all,

First of all, thank you for the great work you've been doing on this library. It's much appreciated.

It seems the new rule [NUnit1032](https://github.com/nunit/nunit.analyzers/blob/master/documentation/NUnit1032.md) introduced in `v3.7.0` is flagging false positives and the wrong fix suggestion when dealing with test fixtures with `LifeCycle.InstancePerTestCase` and constructors.

## Works as expected with `LifeCycle.SingleInstance`

It works as expected with `LifeCycle.SingleInstance`. Given this disposable class:

```csharp
public class TestDisposable : IDisposable { public void Dispose() { /* do nothing */ } }
```

and this `LifeCycle.SingleInstance` failing fixture class:

```csharp
[FixtureLifeCycle(LifeCycle.SingleInstance)]
internal sealed class FailingNUnit1032TestsSingleInstance
{
    // Correctly flags NUnit1032 with correct fix "The field _testDisposable should be Disposed in a method
    // annotated with [OneTimeTearDownAttribute]"
    private readonly TestDisposable _testDisposable;

    public FailingNUnit1032TestsSingleInstance()
    {
        _testDisposable = new TestDisposable();
    }
    
    [Test]
    public void SomeTest() { /* do something */ /*}
}
```

The suggested fix works:

```csharp
[FixtureLifeCycle(LifeCycle.SingleInstance)]
internal sealed class FixedNUnit1032TestsSingleInstance
{
    private readonly TestDisposable _testDisposable;

    public FixedNUnit1032TestsSingleInstance()
    {
        _testDisposable = new TestDisposable();
    }

    [OneTimeTearDown]
    public void TearDown()
    {
        _testDisposable.Dispose();
    }

    [Test]
    public void SomeTest() { /* do something */ }
}
```

## `LifeCycle.InstancePerTestCase` introduces some issues

However, if we switch to `LifeCycle.InstancePerTestCase`.

```csharp
[FixtureLifeCycle(LifeCycle.InstancePerTestCase)]
internal sealed class FailingNUnit1032TestsInstancePerTestCase
{
    // Correctly flags NUnit1032 but with the wrong fix "The field _testDisposable should be Disposed 
    // in a method annotated with [OneTimeTearDownAttribute]"
    private readonly TestDisposable _testDisposable;

    public FailingNUnit1032TestsInstancePerTestCase()
    {
        _testDisposable = new TestDisposable();
    }
    
    [Test]
    public void SomeTest() { /* do something */ /*}
}
```

The first issue is that a OneTimeTearDownAttribute [needs to be static](https://docs.nunit.org/articles/nunit/writing-tests/attributes/fixturelifecycle.html) when using `LifeCycle.InstancePerTestCase` as it only runs once per test fixture, which prevents the disposal of the non-static field as suggested by the fix.

The second problem is that doing a disposal in `[TearDown]` does not remove the NUnit1032 violation, which I believe it should as it's called after the execution of every test as in `Constructor > test > Teardown`:

```csharp
[FixtureLifeCycle(LifeCycle.InstancePerTestCase)]
internal sealed class ShouldWorkNUnit1032TestsInstancePerTestCase
{
    // NUnit1032 still being flagged
    private readonly TestDisposable _testDisposable;

    public ShouldWorkNUnit1032TestsInstancePerTestCase()
    {
        _testDisposable = new TestDisposable();
    }
    
    [TearDown]
    public void TearDown()
    {
        _testDisposable.Dispose();
    }

    [Test]
    public void SomeTest() { /* do something */ }
}
```

The following does work, but then we lose some of the benefits of the constructor (e.g. readonly fields) and we need to disable nullability warnings (or disable/handle it in the tests or global level in `.editorconfig`  or similar).

```csharp
[FixtureLifeCycle(LifeCycle.InstancePerTestCase)]
internal sealed class FixedNUnit1032TestsInstancePerTestCase
{
    // NUnit1032 is gone but we have other warnings to handle
#pragma warning disable CS8618
    private TestDisposable _testDisposable;
#pragma warning enable CS8618
    
    [SetUp]
    public void SetUp()
    {
        _testDisposable = new TestDisposable();
    }
    
    [TearDown]
    public void TearDown()
    {
        _testDisposable.Dispose();
    }

    [Test]
    public void SomeTest() { /* do something */ }
}
```

manfred-brands:
Fixes #595 

When a TestFixture uses `LifeCycle.InstancePerTestCase` it can use an `IDispose` implementation.

Fields can be initialized in the constructor and disposed in the `Dispose` method.

### DIFF
--- a/documentation/NUnit3003.md
+++ b/documentation/NUnit3003.md
@@ -8,7 +8,7 @@
 | Severity | Info
 | Enabled  | True
 | Category | Suppressor
-| Code     | [TestFixtureSuppressor](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/DiagnosticSuppressors/TestFixtureSuppressor.cs)
+| Code     | [AvoidUninstantiatedInternalClassSuppressor](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/DiagnosticSuppressors/AvoidUninstantiatedInternalClassSuppressor.cs)
 
 ## Description
 
--- a/documentation/NUnit3004.md
+++ b/documentation/NUnit3004.md
@@ -8,7 +8,7 @@
 | Severity | Info
 | Enabled  | True
 | Category | Suppressor
-| Code     | [TestFixtureSuppressor](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/DiagnosticSuppressors/TestFixtureSuppressor.cs)
+| Code     | [TypesThatOwnDisposableFieldsShouldBeDisposableSuppressor](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/DiagnosticSuppressors/TypesThatOwnDisposableFieldsShouldBeDisposableSuppressor.cs)
 
 ## Description
 
--- a/src/nunit.analyzers.tests/Constants/NUnitFrameworkConstantsTests.cs
+++ b/src/nunit.analyzers.tests/Constants/NUnitFrameworkConstantsTests.cs
@@ -155,6 +155,9 @@ namespace NUnit.Analyzers.Tests.Constants
             (nameof(NUnitFrameworkConstants.FullNameOfTypeSetUpAttribute), typeof(SetUpAttribute)),
             (nameof(NUnitFrameworkConstants.FullNameOfTypeTearDownAttribute), typeof(TearDownAttribute)),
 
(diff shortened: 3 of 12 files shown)


mikkelbu (approved):

PR #613, opened 2023-10-17
Also look at static methods for NUnit1032

### LINKED ISSUE #607: NUnit1032 false positive when test class is static
Discovered on NUnit.Analysers v3.8.0.

The following code raises [NUnit1032](https://docs.nunit.org/articles/nunit-analyzers/NUnit1032.html) on the declaration of Enumerator.

```
using FluentAssertions;
using NUnit.Framework;

public static class ExampleTests
{
    private static readonly IEnumerator<string> Enumerator = new List<string>().GetEnumerator();

    [OneTimeTearDown]
    public static void TearDown()
    {
        Enumerator.Dispose();
    }

    [Test]
    public static void StubTest()
    {
        true.Should().BeTrue();
    }
}
```

I suggest that either NUnit1032 should not be raised at all as Enumerator is static, or it should be cleared by the existence of the TearDown function.

It seems to be the `static` keyword causing issues here. Making the class and the TearDown function non-static removes the NUnit1032 error.

edit: typos

manfred-brands:
Fixes #607 

Change is small, new tests were more work,


mikkelbu (approved):

PR #616, opened 2023-10-20
Do not consider Assert.That with TestDelegate to be a candidate for Assert.Multiple

### LINKED ISSUE #614: `NUnit2045` does not respect lambda captures
I have a test like this

``` csharp
[Test]
public void MyTest()
{
    object? actualDeserialized = null;

    Assert.That(() => actualDeserialized = Calculate(), Throws.Nothing);
    Assert.That(actualDeserialized, Is.Not.Null);
}

private static object? Calculate() { return new(); }
```

This triggers `NUnit2045`. However, because the lambda captures the `actualDeserialized` variable, these two assertions are not independent.

Is there any way Roslyn can delect when a lambda captures a certain variable, so `NUnit2045` can be disabled for this circumstance?

edit: my minimal repro did demonstrate the issue, but was a little nonsensical. I tweaked it.

manfred-brands:
Fixes #614 

The analyzer will not try to delve into `TestDelegate` to see if it is independent or not.
It will consider this `Assert` to be not mergeable inside an `Assert.Multiple`

### DIFF
--- a/src/nunit.analyzers/UseAssertMultiple/UseAssertMultipleAnalyzer.cs
+++ b/src/nunit.analyzers/UseAssertMultiple/UseAssertMultipleAnalyzer.cs
@@ -82,7 +82,13 @@ namespace NUnit.Analyzers.UseAssertMultiple
                 var previousArguments = new HashSet<string>(StringComparer.Ordinal);
 
                 // No need to check argument count as Assert.That needs at least one argument.
-                var assertArgument = assertOperation.Arguments[0].Syntax.ToString();
+                IArgumentOperation assertArgumentOperation = assertOperation.Arguments[0];
+                if (assertArgumentOperation.Value is IDelegateCreationOperation)
+                {
+                    return;
+                }
+
+                var assertArgument = assertArgumentOperation.Syntax.ToString();
 
                 IOperation? statementBefore = null;
                 int firstAssert = -1;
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers/UseAssertMultiple/UseAssertMultipleAnalyzer.cs
```diff
@@ -139,7 +145,16 @@ private static bool IsIndependentAssert(HashSet<string> previousArguments, IOper
             if (currentAssertOperation is not null)
             {
                 // No need to check argument count as Assert.That needs at least one argument.
-                string currentArgument = currentAssertOperation.Arguments[0].Syntax.ToString();
+                IArgumentOperation argumentOperation = currentAssertOperation.Arguments[0];
+                if (argumentOperation.Value is IDelegateCreationOperation)
+                {
+                    // Assert.That(() => { SomeCode }, Throws.Nothing);
+                    // TODO: Should we delve into the lambda?
+                    // For now state that it isn't mergable inside an Assert.Multiple.
```
mikkelbu, line 153:
```suggestion
                    // For now state that it isn't mergeable inside an Assert.Multiple.
```

manfred-brands:
@mikkelbu I cannot merge this PR because it still depends on appveyor which has been misbehaving, aka not starting,

PR #636, opened 2023-11-10
Suppress ValuesUsageAnalyzer when using null-forgiving operator

### LINKED ISSUE #631: NUnit1001/NUnit1031 don't observe null forgiveness operator
Given a test that doesn't accept `null` as an argument;
```csharp
[TestCase(null)]
[TestCase("bad|path.png", IncludePlatform = "Win")]
public void InvalidFilePathsThrowsArgumentException(string filePath)
{
    Assert.That(() => TestContext.AddTestAttachment(filePath), Throws.InstanceOf<ArgumentException>());
}
```

The first `TestCase` correctly fires NUnit1001.
Changing the case to `TestCase(null!)` should not fire the warning as the user explicitly stated what they wanted.

The current workaround is to declared _filePath_ as nullable and used the null forgiveness operator when _filePath_ is passed.

CollinAlpert:
Fixes #631 

This teaches the `ValuesUsageAnalyzer` to respect the null-forgiving operator.

The test passes when using .NET 8, however I wasn't able to build locally with the current setup, so it would be great if someone who is able to run the tests as-is could try it out.

### DIFF
--- a/src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs
+++ b/src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs
@@ -173,13 +173,19 @@ namespace NUnit.Analyzers.TestCaseUsage
                 if (methodParameterType.IsTypeParameterAndDeclaredOnMethod())
                     continue;
 
+                var argumentSyntax = attribute.GetAdjustedArgumentSyntax(i, attributePositionalArguments, context.CancellationToken);
+
+                if (argumentSyntax is null)
+                    continue;
+
                 ITypeSymbol? argumentType = attributeArgument.Type;
 
                 var argumentTypeMatchesParameterType = attributeArgument.CanAssignTo(
                     methodParameterType,
                     context.Compilation,
                     allowImplicitConversion: true,
-                    allowEnumToUnderlyingTypeConversion: true);
+                    allowEnumToUnderlyingTypeConversion: true,
+                    suppressNullableWarning: argumentSyntax.IsSuppressNullableWarning());
 
                 if (methodParameterParamsType is null && argumentTypeMatchesParameterType)
                     continue;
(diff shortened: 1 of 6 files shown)


src/nunit.analyzers/ValuesUsage/ValuesUsageAnalyzer.cs
```diff
@@ -83,6 +84,12 @@ private static void AnalyzeParameter(SymbolAnalysisContext symbolContext, INamed
                         continue;
                     }
 
+#if NETSTANDARD2_0_OR_GREATER
+                    if (argumentSyntax.IsKind(SyntaxKind.SuppressNullableWarningExpression))
+                    {
+                        continue;
+                    }
+#endif
```
manfred-brands: This is the wrong spot to put this test. This allows code like `public void Test([Values(null!)] int i)`.
It now even allows `public void Test([Values(2!)] string s)`.
The analyzer detects a type mismatch, but the operator suppresses all errors when the null forgiveness operator is used.

manfred-brands (commented):
Thanks @CollinAlpert for your contribution.
Although you fix works for your use case, it would be hiding other errors.

I have changed to code to add an extra parameter to the `CanAssignTo` method.
The extra parameter indicates if a null forgiveness operator was used.
Its value is only checked if the value passed is `null`.

I also updated the TestCaseUsageAnalyzer as it has the same problem.

src/nunit.analyzers/ValuesUsage/ValuesUsageAnalyzer.cs
CollinAlpert: Nice catch, thanks!

mikkelbu (approved):
LGTM. Just a single questions. Thanks for the contribution @manfred-brands

src/nunit.analyzers/TestCaseUsage/TestCaseUsageAnalyzer.cs
```diff
@@ -190,7 +196,8 @@ private static bool IsTypeAnObjectArray(ITypeSymbol? typeSymbol)
                         methodParameterParamsType,
                         context.Compilation,
                         allowImplicitConversion: true,
-                        allowEnumToUnderlyingTypeConversion: true);
+                        allowEnumToUnderlyingTypeConversion: true,
+                        suppressNullableWarning: false);
```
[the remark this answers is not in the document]
manfred-brands: Good find.
I added tests and updated the code to allow a null forgiveness operator to elements of a `params` array.

PR #646, opened 2023-11-18
chore: Remove VSIX project

### LINKED ISSUE #429: Drop the VSIX project
The VSIX project only allows the analyzers to run in Visual Studio, which means that CLI build would not use them.

They have been [removed ](https://github.com/dotnet/roslyn-analyzers/issues/3314#issuecomment-747598315) from the VS extension marketplace for the roslyn analyzers.

NuGet package allows different rules/versions for different projects and can be used in CI processes.

mikkelbu:
fixes #429


manfred-brands (approved):
Nice. Thanks @mikkelbu

PR #647, opened 2023-11-19
chore: Add tooling from docs project

### LINKED ISSUE #630: Update NUnit Analyzer docs to respect 120-character docs rule
We adopted 120-character line rules in https://github.com/nunit/docs/issues/830.

I've changed the analyzer docs in the docs repo as part of that PR, but I understand they need to be changed here too. 

Let me know how I can best assist with that change.

mikkelbu:
Fixes #630


mikkelbu:
@SeanKilleen I've made all the linting work in this repo except for one
```
documentation/NUnit2028.md:3:121 MD013/line-length Line length [Expected: 120; Actual: 130]
```
We have a title in NUnit2028 that is too long.
```md
## Consider using Assert.That(actual, Is.GreaterThanOrEqualTo(expected)) instead of ClassicAssert.GreaterOrEqual(actual, expected)
```
I cannot understand why this is not failing in the docs repo as it is also too long there (note that we have now prepended "Classic" to `Assert.GreaterOrEqual` (due to the changes in NUnit4), but in the docs repo it is still 123 characters.

As far as I can tell it is not possible to break a line in the header, so I'm considering the alternatives:
* increasing the limit to 130, but then the build will fail when we transport the MD files to the docs repo - unless the limit is also increased there
* rephrasing the title so it is shorter, but it is hard due to the "method calls"
* ignore this validation for this file/line (I've not looked into this yet).

Do you have any suggestions?

SeanKilleen:
@mikkelbu I'm not near technology today but I set a reminder to look at this tomorrow.

Definitely surprised that it's failing for you and not in the docs repo too if the configs match. Maybe check the actions used and their versions?

mikkelbu:
> @mikkelbu I'm not near technology today but I set a reminder to look at this tomorrow.
> 
> Definitely surprised that it's failing for you and not in the docs repo too if the configs match. Maybe check the actions used and their versions?

I just tried it in the docs repo - https://github.com/nunit/docs/actions/runs/6926833928/job/18839626771?pr=838 - and it fails with the same error. The odd part is that I made it 7 characters longer and it fails with `[Expected: 120; Actual: 130]` which indicates that it was also longer to begin with, but perhaps it is due to the header symbols or the special symbols in the line?

manfred-brands:
> * ignore this validation for this file/line (I've not looked into this yet).

Can we use the below for the header?
```
<!-- eslint-disable-next-line max-len>
```

mikkelbu:
I guess the reason it does not complain with the existing length is the exception

https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md013---line-length
> This rule has an exception where there is no whitespace beyond the configured line length. This allows you to still include items such as long URLs without being forced to break them in the middle.

@manfred-brands As far as I can tell from the code at https://github.com/markdownlint/markdownlint/tree/main it is not possible to turn off validation - the most relevant I could find was this unanswered question https://github.com/markdownlint/markdownlint/issues/421, so I think  our alternatives are:
* Create our own rule (that special case headers)
* Increase the limit to 130 characters
* Rephrase the NUnit2028 (the simple hack could just be to remove the space in `ClassicAssert.GreaterOrEqual(actual, expected)` although this feels like a major hack

I think I most for increasing the limit although I would have liked for a better solution

manfred-brands:
It does seem to support a comment similar to my initial one:

> Disable all rules for the next line: <!-- markdownlint-disable-next-line

SeanKilleen:
@mikkelbu @manfred-brands with my change, the linting is now passing.

PR #642, opened 2023-11-16
Add support for Assert.MultipleAsync

manfred-brands:
Fixes #615 

This affects:
- DereferencePossiblyNullReferenceSuppressor
- UseAssertMultipleAnalyzer

PR #648, opened 2023-11-19
chore: Skip branch builds on PRs

mikkelbu:


appveyor.yml
```diff
@@ -32,3 +32,6 @@ deploy:
     skip_symbols: true
     on:
       branch: master
+
+# For PRs, skip the branch run and only run against the virtual PR merge.
+skip_branch_with_pr: true
```
manfred-brands, line 37: If we do this, then we loose the  NuGet packages artifacts. Should we add an upload-artifact step to the ci build?

mikkelbu: I didn't think about the artifacts. We still get NuGet packages artifacts for the "virtual" PR build. Do you use the packages from the branch build? (it was mostly to avoid having to run both (sequentially) on PRs).

Alternatively, we can try to increase `max_jobs` to 2?

manfred-brands: I have used the artifacts of PR builds in the past, but probably can just create the package locally as I need to have them in my local NuGet folder anyhow.

mikkelbu: Note that we will still have artifacts of the PR builds, but now only from the "PR" build - before we had two AppVeyor builds per PR (one for branch build and one for the merge build)
image
So this PR should remove the one build - `continuous-integration/appveyor/branch`

manfred-brands: I've only seen one, e.g. #642.

mikkelbu: Odd. I wonder if the branch builds are removed after some time? I can see two on #646, but only one on e.g. #642 or #641 (note that is the missing one that I'm trying to remove)

mikkelbu: I'll try this out and if we loose the artifacts then I'll revert. Hopefully, this will just skip one of the builds, thus reducing the wait on AppVeyor on PRs.

PR #657, opened 2023-11-28
Allow public override methods.

### LINKED ISSUE #656: NUnit1028 warns about overridden methods
# Description

When overriding a method from a (potentially abstract) base-class, `NUnit1028` (A fixture should not contain any public non-test methods) fires.

We can not change the visibility of a method when overriding it. Therefore, this analyzer should not be warning about methods that are overriding something from their base class.

# Reproduction

Set up a test-project with following fixture in it:

```csharp
using NUnit.Framework;

namespace SomeNamespace;

public sealed class SomeTests {
    [Test]
    public void Test() => Assert.That(new object(), Is.Not.Null);

    public override string ToString() => "Test";
}
```

You will be able to see that the last method has a warning from `NUnit1028`.

# Solution

Check `method.OverriddenMethod` for a null value on line 48 in `NonTestMethodAccessibilityLevelAnalyzer.cs`.

Of course, I am willing to send a merge request to fix this issue. Please let me know whether I shall do that.

---

Thank you for looking into this.

manfred-brands:
Fixes #656


mikkelbu (approved):
Nice and simple fix. Thanks for the PR @manfred-brands

PR #670, opened 2023-12-22
Also support codefix for .Within

### LINKED ISSUE #669: NUnit2007 does not provide codefix when .Within is used
When having something like this
```csharp
double expected = 5d;
Assert.That(1d, Is.EqualTo(expected).Within(double.Epsilon));
```

NUnit2007 warns correctly but no codefix is provided.

Laniusexcubitor:
fixes #669

### DIFF
--- a/src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageCodeFixTests.cs
@@ -45,6 +45,27 @@ namespace NUnit.Analyzers.Tests.ConstActualValueUsage
                 fixTitle: ConstActualValueUsageCodeFix.SwapArgumentsDescription);
         }
 
+        [TestCase(nameof(ClassicAssert.AreEqual))]
+        public void LiteralArgumentIsProvidedForClassicAssertWithDeltaCodeFix(string classicAssertMethod)
+        {
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+                public void Test()
+                {{
+                    int expected = 5;
+                    ClassicAssert.{classicAssertMethod}(expected, ↓1, 1);
+                }}");
+
+            var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+                public void Test()
+                {{
+                    int expected = 5;
+                    ClassicAssert.{classicAssertMethod}(1, expected, 1);
+                }}");
+
+            RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode,
+                fixTitle: ConstActualValueUsageCodeFix.SwapArgumentsDescription);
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers.tests/ConstActualValueUsage/ConstActualValueUsageCodeFixTests.cs
```diff
@@ -45,6 +45,27 @@ public void Test()
...
         }
 
+        [TestCase(nameof(ClassicAssert.AreEqual))]
+        public void LiteralArgumentIsProvidedForClassicAssertWithDeltaCodeFix(string classicAssertMethod)
+        {
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+                public void Test()
+                {{
+                    int expected = 5;
+                    ClassicAssert.{classicAssertMethod}(expected, ↓1, 1);
```
manfred-brands: It is never good to use two arguments with the same value as it doesn't proof that the correct argument is exchanged.
I have update the code to use `double.Epsilon` for the accuracy.

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageCodeFix.cs
```diff
@@ -157,6 +155,32 @@ private static bool IsSupportedClassicAssert(IMethodSymbol methodSymbol)
...
 
+        private static bool IsSupportedIsConstraint(ref InvocationExpressionSyntax invocationExpression, [NotNullWhen(true)]out MemberAccessExpressionSyntax? memberAccessExpression)
+        {
+            if (invocationExpression.Expression is not MemberAccessExpressionSyntax mae)
+            {
+                memberAccessExpression = null;
+                return false;
+            }
+
+            memberAccessExpression = mae;
```
manfred-brands: I simplified this a tiny bit. Mainly because I don't like shorthand names like `mae`

src/nunit.analyzers/ConstActualValueUsage/ConstActualValueUsageCodeFix.cs
```diff
@@ -157,6 +155,32 @@ private static bool IsSupportedClassicAssert(IMethodSymbol methodSymbol)
...
+
+            if (memberAccessExpression.Name.ToString() == NUnitFrameworkConstants.NameOfEqualConstraintWithin)
+            {
+                // e.g. Is.EqualTo(1).Within(1) or Is.Not.EqualTo(1).Within(1)
+                if (memberAccessExpression.Expression is not InvocationExpressionSyntax ies
```
manfred-brands: Renamed `ies` into `innerInnovationExpression`

manfred-brands (approved):
@Laniusexcubitor Thanks for your contribution.

I have made some tiny changes.
One to ensure that the test validates that the correct parameter is swapped.
The other because I'm pedantic and don't like abbreviated variable names.

PR #675, opened 2024-01-20
feat: Drop support of VS before 2019 version 16.3

### LINKED ISSUE #609: Drop support for VS before 2019
We currently have separate releases 2.x and 3.x, where the first is the code without the `DiagnosticSuppressors`.
Those were introduced in [roslyn version 3.3.1](https://learn.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.diagnostics.diagnosticsuppressor?view=roslyn-dotnet-3.3.1)
which matches [Visual Studio 2019 16.3](https://learn.microsoft.com/en-us/visualstudio/extensibility/roslyn-version-support?view=vs-2019) which is 4 years old and [predates .NET Core 3.1](https://learn.microsoft.com/en-us/dotnet/core/porting/versioning-sdk-msbuild-vs).

I expect all users, even those addressing .NET Framework to use later SDKs to take advantage of new C# language feates.

mikkelbu:
Fixes #609


manfred-brands (approved):
A few suggestions in the documentation changes otherwise looks fine to me.

PR #678, opened 2024-01-20
Update Test method parameter checks to account for CancellationToken

### LINKED ISSUE #663: NUnit1027 fired when `CancellationToken` and `[CancelAfter]` is given
Repro (basically just the example from the [docs](https://docs.nunit.org/articles/nunit/writing-tests/attributes/cancelafter.html):
```c#
[TestFixture]
public class MyTests
{
    [Test, CancelAfter(2000)]
    public void RunningTestUntilCanceled(CancellationToken token)
    {
        while (!token.IsCancellationRequested)
        {
            /* */
        }
    }
}
```

**Expected behavior**: the analyzer understand the `[CancelAfter]` attribute, so that the value for the argument will be given by the NUnit framework.

**Actual behavior**: the analyzer triggers 
grafik

manfred-brands:
Fixes #663

### DIFF
--- a/src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
@@ -708,7 +708,7 @@ namespace NUnit.Analyzers.Tests.TestCaseSourceUsage
 
 #if NUNIT4
         [Test]
-        public void AnalyzeWhenNumberOfParametersMatchExcludingImplicitSuppliedCancellationToken()
+        public void AnalyzeWhenNumberOfParametersMatchExcludingImplicitSuppliedCancellationTokenDueToCancelAfterOnMethod()
         {
             var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
     [TestFixture]
@@ -716,6 +716,33 @@ namespace NUnit.Analyzers.Tests.TestCaseSourceUsage
     {
         [TestCaseSource(nameof(TestData), new object[] { 1, 3, 5 })]
         [CancelAfter(10)]
+        public void ShortName(int number, CancellationToken cancellationToken)
+        {
+            if (cancellationToken.IsCancellationRequested)
+                Assert.Ignore(""Cancelled"");
+            Assert.That(number, Is.GreaterThanOrEqualTo(0));
+        }
+
+        static IEnumerable<int> TestData(int first, int second, int third)
+        {
+            yield return first;
+            yield return second;
+            yield return third;
+        }
+    }", additionalUsings: "using System.Collections.Generic;using System.Threading;");
+
+            RoslynAssert.Valid(analyzer, testCode);
+        }
+
+        [Test]
+        public void AnalyzeWhenNumberOfParametersMatchExcludingImplicitSuppliedCancellationTokenDueToCancelAfterOnClass()
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
+    [TestFixture]
+    [CancelAfter(100)]
+    public class AnalyzeWhenNumberOfParametersMatch
+    {
+        [TestCaseSource(nameof(TestData), new object[] { 1, 3, 5 })]
         public void ShortName(int number, CancellationToken cancellationToken)
         {
             if (cancellationToken.IsCancellationRequested)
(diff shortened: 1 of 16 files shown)


src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
```diff
@@ -755,5 +755,60 @@ public sealed class AnalyzeWhenTestMethodHasTypeParameterArgumentType
     }");
             RoslynAssert.Valid(this.analyzer, testCode);
         }
+
+#if NUNIT4
```
mikkelbu, line 759: Do we need this when we have declared the CancelAfterAttribute for NUNIT3 ?

src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
```diff
@@ -705,5 +705,63 @@ public void ShortName(int first, int second)
 
             RoslynAssert.Valid(analyzer, testCode);
         }
+
+#if NUNIT4
```
mikkelbu, line 709: Do we need this when we have declared the `CancelAfterAttribute` for NUNIT3 ?

mikkelbu:
General comment. It seems like all the code (analyzer and tests) only checks the case where the CancelAfterAttribute is placed directly on the test-method, but I guess this attribute could also have been placed on the test fixture?

src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
manfred-brands: Yes, the test project has this class only because of the `NUnitFrameworkConstantsTests` class.
These test here are compiled against the real NUnit assembly which for NUnit 3 doesn't have the `CancelAfterAttribute`.

src/nunit.analyzers.tests/TestCaseUsage/TestCaseUsageAnalyzerTests.cs
manfred-brands: Yes, the test project has this class only because of the `NUnitFrameworkConstantsTests` class.
These test here are compiled against the real NUnit assembly which for NUnit 3 doesn't have the `CancelAfterAttribute`.

src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs
```diff
@@ -238,7 +245,9 @@ private static void AnalyzeAttribute(SyntaxNodeAnalysisContext context, INamedTy
                         IMethodSymbol? testMethod = context.SemanticModel.GetDeclaredSymbol(testMethodDeclaration);
                         if (testMethod is not null)
                         {
-                            var (methodRequiredParameters, methodOptionalParameters, methodParamsParameters) = testMethod.GetParameterCounts();
+                            var hasCancelAfterAttribute = testMethod.GetAttributes().Any(a => SymbolEqualityComparer.Default.Equals(a.AttributeClass, cancelAfterType));
+
+                            var (methodRequiredParameters, methodOptionalParameters, methodParamsParameters) = testMethod.GetParameterCounts(true, cancellationTokenType);
```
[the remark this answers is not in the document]
manfred-brands: Yes indeed. Thanks.

src/nunit.analyzers/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -15,7 +15,9 @@ internal static class IMethodSymbolExtensions
         /// and the last is the <see langword="params" /> count.
         /// </returns>
         internal static (uint requiredParameters, uint optionalParameters, uint paramsCount) GetParameterCounts(
-            this IMethodSymbol @this)
+            this IMethodSymbol @this,
+            bool hasCancelAfterAttribute,
+            INamedTypeSymbol? cancellationTokenType)
```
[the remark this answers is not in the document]
manfred-brands: Done. I am so used to StyleCop prompting me that I forgot.

src/nunit.analyzers/Extensions/IMethodSymbolExtensions.cs
```diff
@@ -39,6 +41,15 @@ internal static class IMethodSymbolExtensions
                 }
             }
 
+            var hasCancellationToken = parameters.Length > 0 &&
+                                       SymbolEqualityComparer.Default.Equals(parameters[parameters.Length - 1].Type, cancellationTokenType);
+            if (hasCancelAfterAttribute && hasCancellationToken)
+            {
+                // This parameter is optional, it not specified it will be supplied by NUnit.
```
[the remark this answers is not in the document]
manfred-brands: Done.

manfred-brands:
> General comment. It seems like all the code (analyzer and tests) only checks the case where the CancelAfterAttribute is placed directly on the test-method, but I guess this attribute could also have been placed on the test fixture?

Yes. The `CancelAfterAttribute` is allowed on the class. Myself I don't see proper use of it, what if you add more tests, then the fixture needs more time even though all individual tests are fine. 

But as NUnit supports it, I added tests and support for looking up the attribute on the class.

mikkelbu (approved):
+1LGTM. I'm happy to merge this when you are @manfred-brands. Thanks for the work 👍

PR #688, opened 2024-02-04
Remove name colon on tolerance argument with code fix

verdie-g:
Fix https://github.com/nunit/nunit.analyzers/issues/689.

### DIFF
--- a/src/nunit.analyzers/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFix.cs
+++ b/src/nunit.analyzers/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFix.cs
@@ -32,7 +32,7 @@ namespace NUnit.Analyzers.ClassicModelAssertUsage
 
             if (hasToleranceValue)
             {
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFix.cs
```diff
@@ -32,13 +32,17 @@ protected override void UpdateArguments(Diagnostic diagnostic, List<ArgumentSynt
 
             if (hasToleranceValue)
             {
+                // The tolerance argument should be renamed from delta to amount but with the model constraint it the
```
manfred-brands, line 35:
```suggestion
                // The tolerance argument should be renamed from 'delta' to 'amount' but with the model constraint the
```

manfred-brands (approved):
Thank you. The code changes look good. I only have a small change request for the comment

mikkelbu (approved):
Thanks for the fix @verdie-g and for the review @manfred-brands.

PR #694, opened 2024-02-21
Switch to using license expression

lahma:
See https://github.com/nunit/nunit/pull/4617 , this will align analyzer project with core projects and test adapter.


mikkelbu (approved):
Thanks for PR @lahma

PR #698, opened 2024-03-04
Remove null message in AreEqual code fix

verdie-g:
```
Assert.AreEqual(2d, 3d, null);
```
would generate
```
Assert.That(3d, Is.EqualTo(2d), null);
```
which does not compile
```
The call is ambiguous between the following methods or properties:
'Assert.That<TActual>(TActual, IResolveConstraint, FormattableString,
string, string)' and 'Assert.That<TActual>(TActual, IResolveConstraint,
Func<string>, string, string)'
```

### DIFF
--- a/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
@@ -57,6 +57,22 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         }
 
         [Test]
+        public void VerifyAreEqualFixWithNullMessage()
+        {
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void TestMethod()
+        {
+            ↓ClassicAssert.AreEqual(2d, 3d, null);
+        }");
+            var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void TestMethod()
+        {
+            Assert.That(3d, Is.EqualTo(2d));
+        }");
+            RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
+        }
+
+        [Test]
         public void VerifyAreEqualFixWithMessageAndParams()
         {
             var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFix.cs
```diff
@@ -55,6 +55,12 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
             // Now, replace the arguments.
             List<ArgumentSyntax> arguments = invocationNode.ArgumentList.Arguments.ToList();
 
+            // Remove null message to avoid ambiguous calls.
+            if (arguments.Count == 3 && arguments[2].Expression.Kind() == SyntaxKind.NullLiteralExpression)
+            {
+                arguments.RemoveAt(2);
+            }
+
```
manfred-brands, line 63: This needs to move after the call to `UpdateArguments` below so that it also works if there is a tolerance value specified.

src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
```diff
@@ -56,6 +56,22 @@ public void TestMethod()
             RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
         }
 
+        [Test]
+        public void VerifyAreEqualFixWithNullMessage()
```
manfred-brands, line 60: Please add another test with a tolerance value: `ClassicAssert.AreEqual(2d, 3d, 0.1, null);`

manfred-brands (changes requested):
With a small change, your fix would also work for the other use case.

manfred-brands:
Thank you @verdie-g for the PR. For administration sake I have created an issue (#700) for this.

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFix.cs
```diff
@@ -55,6 +55,12 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
             // Now, replace the arguments.
             List<ArgumentSyntax> arguments = invocationNode.ArgumentList.Arguments.ToList();
 
+            // Remove null message to avoid ambiguous calls.
+            if (arguments.Count == 3 && arguments[2].Expression.Kind() == SyntaxKind.NullLiteralExpression)
```
manfred-brands, line 59:
```suggestion
            if (arguments.Count == 3 && arguments[2].Expression.IsKind(SyntaxKind.NullLiteralExpression))
```

manfred-brands (approved):
Thanks you for your quick fix. Nothing else.

PR #702, opened 2024-03-11
Support Assume.That(x, Is.Not.Null) to suppress nullable warning

### LINKED ISSUE #693: Possibly Null Reference Warning should be suppressed for Assume
The [DereferencePossiblyNullReferenceSuppressor](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressor.cs) class suppresses the `Possible 'System.NullReferenceException' ` warning in case of:

```cs
Assert.That(exception,           Is.Not.Null);
Assert.That(exception.ParamName, Is.EqualTo("udpClient")); //<< No warning
```
Would it be possible for this to work for `Assume` as well?

```cs
Assume.That(exception, Is.Not.Null)
string name = exception.ParamName; //<< Possible 'System.NullReferenceException' warning
```

manfred-brands:
Fixes #693


mikkelbu (approved):
LGTM. Thanks for the contribution @manfred-brands 👍

PR #703, opened 2024-03-11
NUnit2010: Add support for detecting use of 'is' pattern inside 'Assert.That'

### LINKED ISSUE #691: Extent rule NUnit2010 to detect 'is null'
Rule NUnit2010 fires for: `Assert.That(instance == null)` but not for `Assert.That(instance is null)`.
It should also detected `is not null` to mean `!= null`.
Code fixes should create: `Assert.That(instance, Is.Null)` and `Assert.That(instance, Is.Not.Null)` respectively.

manfred-brands:
Fixes #691

### DIFF
--- a/src/nunit.analyzers.tests/ConstraintsUsage/EqualConstraintUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/ConstraintsUsage/EqualConstraintUsageAnalyzerTests.cs
@@ -42,6 +42,26 @@ namespace NUnit.Analyzers.Tests.ConstraintsUsage
         }
 
         [Test]
+        public void AnalyzeWhenIsOperatorUsed()
+        {
+            var testCode = TestUtility.WrapInTestMethod(@"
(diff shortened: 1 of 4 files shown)


mikkelbu (approved):
Looks good @manfred-brands, and I think it is fine to keep it within the existing analyzer.

My only _minimal_ concern is that the And/OrConstraints does not exactly match the semantics of the patterns. For instance, we have the note in https://docs.nunit.org/articles/nunit/writing-tests/constraints/OrConstraint.html#evaluation-order-and-precedence, but one might argue that this failure in the OrConstraint is an error and an implementation detail. 

I've also tried to test some more complicated scenarios - combining multiple And and Ors - as I thought I could remember that we also had some problems with the precedence in NUnit, but I was not able to find the problem (if there is one).

So all in all I think the PR is an improvement and I'm not too worried about the differences as I don't think most people have very complicated patterns in their asserts

manfred-brands:
Thanks @mikkelbu

PR #704, opened 2024-03-12
Ignore null suppression operator when matching Is.Not.Null

### LINKED ISSUE #679: Null suppression doesn't work when the body has a null suppression
A null suppression throws the null warning suppression out of whack.

One might say why is `str` not getting checked for null. The following is a minimal repro, the actual scenarios I've ran into are less trivial. Suffice it to say, the existence of the null-suppression operator should be evidence enough that it needs to be used sometimes.

``` csharp
internal static string GetString(HasString str)
{
    Assert.That(str.Inner, Is.Not.Null);
    return str.Inner; // works great
}

internal static string GetStringSuppress(HasString? str) // argument is nullable
{
    Assert.That(str!.Inner, Is.Not.Null);
    return str.Inner; // warning: possible null reference return
}

public record HasString(string? Inner);
```

manfred-brands:
Fixes #679


mikkelbu (approved):
LGTM. Nice and simple 👍 . I was considering if the replace of `!` could remove some instances of the unary negation operator, but I don't think that this can be the case in this context.

PR #710, opened 2024-03-26
Nunit1032 - consider dispose of a type having explicit interface implementation

matode:
In case a class does inherit from IDisposable, but Dispose is implemented with explicit interface implementation (void IDisposable.Dispose()), then the Dispose call in TearDown method is not considered by NUnit1032.

This change considers casting of given type to IDisposable and doesn't report a finding in such case

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,4 +1,37 @@
-﻿NUnit Analyzers 4.0.1 - February 1, 2024
+﻿NUnit Analyzers 4.1 - March 16, 2024
+
+This release of the Analyzers extends the suppression of nullable warnings to also respect assumptions - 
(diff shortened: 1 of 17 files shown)


matode:
@dotnet-policy-service agree

mikkelbu (approved):
LGTM. I considering if we could make the code in `GetIdentifier` more general to e.g. also support cases such like
```
private DummyExplicitDisposable field2;
...
[OneTimeTearDown]
public void Dispose()
{
    IDisposable x = field2;
    x.Dispose();
}
```

But I think this is already an improvement, and then we can always improve on this later on

mikkelbu:
Thanks for the contribution @matode. I'll merge it now

Issue #713, opened 2024-03-31
[bug] Code fix for NUnit2049 places the comma at a wrong place and messes up indentation


Bartleby2718:
```cs
// before the code fix
CollectionAssert.AreEquivalent(
	new[] { "This is such a long sentence that I don't want the other parameter on the same line." },
	new[] { "This is such a long sentence that I don't want the other parameter on the same line." }
);
```
is transformed into
```cs
// after the current code fix
Assert.That(
	new[] { "This is such a long sentence that I don't want the other parameter on the same line." }
, Is.EquivalentTo(new[] { "This is such a long sentence that I don't want the other parameter on the same line." }));
```
There are two problems with this:
1. To be consistent with the original formatting, the comma should be trailing, not leading.
2. The second argument should have the same indentation as the first argument.
```cs
// after the proposed code fix
Assert.That(
	new[] { "This is such a long sentence that I don't want the other parameter on the same line." },
	Is.EquivalentTo(new[] { "This is such a long sentence that I don't want the other parameter on the same line." }));
```

Similarly,
```cs
// before the code fix
CollectionAssert.AreEquivalent(
	new[] { "This is such a long sentence that I don't want the other parameter on the same line." },
	new[] { "This is such a long sentence that I don't want the other parameter on the same line." });
```
is transformed into
```cs
// after the current code fix
Assert.That(
	new[] { "This is such a long sentence that I don't want the other parameter on the same line." }, Is.EquivalentTo(new[] { "This is such a long sentence that I don't want the other parameter on the same line." }));
```
but I believe the right formatting should be
```cs
// after the proposed code fix
Assert.That(
	new[] { "This is such a long sentence that I don't want the other parameter on the same line." },
	Is.EquivalentTo(new[] { "This is such a long sentence that I don't want the other parameter on the same line." }));
```
I believe fixing the former will also fix the latter.

manfred-brands:
@Bartleby2718 I agree with your right formatting.
However it isn't that simple as to which syntax node the compiler assigns white space.

```csharp
CollectionAssert.AreEquivalent(
    collection1,
    collection2
);
```

| Syntax | ToFullString |
|-|-|
| invocationNode.ArgumentList | "(\r\n                collection1,\r\n                collection2\r\n            )" |
| invocationNode.ArgumentList.OpenParenToken | "(\r\n" |
| invocationNode.ArgumentList.Arguments[0]| "                collection1"
[...]

PR #716, opened 2024-03-31
Fix logic around out-of-order arguments for classic assertions (#712)

### LINKED ISSUE #712: [bug] Code fix for NUnit2005 does not correctly fix Assert.AreEqual if named parameters are used in unexpected order
Given the following (note the order of named parameters)
```cs
    [Test]
    public void Test1()
    {
        var actual = 1;
        Assert.AreEqual(actual: actual, expected: 1);
    }
```
`NUnit.Analyzers` cannot fix `NUnit2005` correctly.

The above should suffice, but I also created a repo: https://github.com/Bartleby2718/NUnit2005-Assert-AreEqual-repro

Bartleby2718:
I found another bug while fixing the code, but in the interest of making the PR more targeted and reviewable, I'll file another issue and another PR to fix that.

Closes #712.

### DIFF
--- a/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
@@ -43,11 +43,11 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         [Test]
         public void VerifyAreEqualFixWhenToleranceExistsWithNamedArgumentButInNonstandardOrder()
         {
-            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
-        {{
+        {
             ↓ClassicAssert.AreEqual(delta: 0.0000001d, actual: 3d, expected: 2d);
-        }}");
+        }");
             var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
         {
@@ -59,11 +59,11 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         [Test]
         public void VerifyAreEqualFixWithNamedParametersInNonstandardOrder()
         {
-            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
-        {{
+        {
             ↓ClassicAssert.AreEqual(actual: 3d, expected: 2d);
-        }}");
+        }");
             var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
         {
@@ -117,6 +117,24 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         {
             Assert.That(3d, Is.EqualTo(2d), $""message-id: {Guid.NewGuid()}"");
         }");
+            RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
+        }
+
+        [Test]
+        public void VerifyAreEqualFixWithMessageAnd2ParamsInNonstandardOrder()
+        {
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void TestMethod()
+        {
+            var actual = 3d;
+            ↓ClassicAssert.AreEqual(delta: 0.0000001d, expected: 2d, actual: actual, args: new[] { ""Guid.NewGuid()"", Guid.NewGuid().ToString() }, message: ""message-id: {0}, {1}"");
+        }");
+            var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void TestMethod()
+        {
+            var actual = 3d;
+            Assert.That(actual: actual, Is.EqualTo(expected: 2d).Within(0.0000001d), $""message-id: {""Guid.NewGuid()""}, {Guid.NewGuid().ToString() }"");
+        }"); // TODO: Fix whitespace
             RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
         }
 
@@ -196,6 +214,22 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         public void TestMethod()
         {
             Assert.That(3d, Is.EqualTo(2d).Within(0.0000001d), $""message-id: {Guid.NewGuid()}"");
+        }");
+            RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
+        }
+
+        [Test]
+        public void VerifyAreEqualFixWhenToleranceExistsWithMessageAndParamsLiteralExpressionSyntax()
+        {
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void TestMethod()
+        {
+            ↓ClassicAssert.AreEqual(2d, 3d, 0.0000001d, ""message-id: {0}"", ""mystring"");
+        }");
+            var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void TestMethod()
+        {
+            Assert.That(3d, Is.EqualTo(2d).Within(0.0000001d), $""message-id: {""mystring""}"");
         }");
             RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
         }
--- a/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
@@ -139,6 +139,24 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         }
 
         [Test]
+        public void VerifyAreEqualFixWithMessageAnd2ParamsInStandardOrder()
+        {
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void TestMethod()
+        {
+            var actual = 3d;
+            ↓ClassicAssert.AreEqual(expected: 2d, actual: actual, delta: 0.0000001d, ""message-id: {0}, {1}"", ""Guid.NewGuid()"", Guid.NewGuid());
+        }");
+            var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void TestMethod()
+        {
+            var actual = 3d;
+            Assert.That(actual: actual, Is.EqualTo(expected: 2d).Within(0.0000001d), $""message-id: {""Guid.NewGuid()""}, {Guid.NewGuid()}"");
+        }");
+            RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
+        }
+
+        [Test]
         public void VerifyAreEqualFixWhenToleranceExists()
         {
             var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
--- a/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
@@ -51,7 +51,7 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
             var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
         {
-            Assert.That(actual: 3d, Is.EqualTo(expected: 2d).Within(0.0000001d));
+            Assert.That(3d, Is.EqualTo(2d).Within(0.0000001d));
         }");
             RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
         }
@@ -67,7 +67,7 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
             var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
         {
-            Assert.That(actual: 3d, Is.EqualTo(expected: 2d));
+            Assert.That(3d, Is.EqualTo(2d));
         }");
             RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
         }
@@ -133,7 +133,7 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         public void TestMethod()
         {
             var actual = 3d;
-            Assert.That(actual: actual, Is.EqualTo(expected: 2d).Within(0.0000001d), $""message-id: {""Guid.NewGuid()""}, {Guid.NewGuid().ToString()}"");
+            Assert.That(actual, Is.EqualTo(2d).Within(0.0000001d), $""message-id: {""Guid.NewGuid()""}, {Guid.NewGuid().ToString()}"");
         }");
             RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
         }
@@ -151,7 +151,7 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         public void TestMethod()
         {
             var actual = 3d;
-            Assert.That(actual: actual, Is.EqualTo(expected: 2d).Within(0.0000001d), $""message-id: {""Guid.NewGuid()""}, {Guid.NewGuid()}"");
+            Assert.That(actual, Is.EqualTo(2d).Within(0.0000001d), $""message-id: {""Guid.NewGuid()""}, {Guid.NewGuid()}"");
         }");
             RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
         }
@@ -183,7 +183,7 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
             var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
         {
-            Assert.That(actual: 3d, Is.EqualTo(expected: 2d).Within(0.0000001d));
+            Assert.That(3d, Is.EqualTo(2d).Within(0.0000001d));
         }");
             RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
         }
--- a/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
@@ -27,11 +27,11 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         [Test]
         public void VerifyAreEqualFix()
         {
-            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
-        {{
+        {
             ↓ClassicAssert.AreEqual(2d, 3d);
-        }}");
+        }");
             var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
         {
@@ -75,11 +75,11 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         [Test]
         public void VerifyAreEqualFixWithMessage()
         {
-            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
-        {{
+        {
             ↓ClassicAssert.AreEqual(2d, 3d, ""message"");
-        }}");
+        }");
             var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
         {
@@ -105,13 +105,13 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         }
 
         [Test]
-        public void VerifyAreEqualFixWithMessageAndParams()
+        public void VerifyAreEqualFixWithMessageAndOneArgumentForParams()
         {
-            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
-        {{
-            ↓ClassicAssert.AreEqual(2d, 3d, ""message-id: {{0}}"", Guid.NewGuid());
-        }}");
+        {
+            ↓ClassicAssert.AreEqual(2d, 3d, ""message-id: {0}"", Guid.NewGuid());
+        }");
             var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
         public void TestMethod()
         {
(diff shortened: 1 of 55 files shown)


Bartleby2718:
@dotnet-policy-service agree

manfred-brands:
@Bartleby2718 You are right that most code fixes don't take named parameters into account and I rather not have 24 separate updates for each of the classic code fixes.

The code already makes assumptions before getting to the `UpdateArguments` method you updated:
```
            arguments[0] = CastIfNecessary(arguments[0]);
            if (arguments.Count > 1)
                arguments[1] = CastIfNecessary(arguments[1]);
```

Can I suggest to do this differently?
Instead of the `UpdateArguments` expecting a `List<ArgumentSyntax>` we could change it to a `Dictionary<string, ArgumentSyntax>`. That way the code fixes can change `arguments[0]` with `arguments[NUnitFrameworkConstants.NameOfExpectedParameter]`.

src/nunit.analyzers/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFix.cs
```diff
@@ -18,6 +18,15 @@ public sealed class AreEqualClassicModelAssertUsageCodeFix
...
         {
+            // If named parameters are used, we can't assume that the first argument is the expected one and the second is the actual.
+            // Therefore, start by finding the expected argument and the actual argument.
+            var expectedPosition = arguments.FindIndex(arg => arg.NameColon?.Name.Identifier.Text == NUnitFrameworkConstants.NameOfExpectedParameter);
+            var actualPosition = arguments.FindIndex(arg => arg.NameColon?.Name.Identifier.Text == NUnitFrameworkConstants.NameOfActualParameter);
+
+            // If named arguments are not used, the first argument is expected, and the second is actual.
+            var expectedArgument = expectedPosition != -1 ? arguments[expectedPosition] : arguments[0];
+            var actualArgument = actualPosition != -1 ? arguments[actualPosition] : arguments[1];
+
```
manfred-brands, line 29: Instead of first finding the index, this can be combined into:
```
var expectedArgument = arguments.SingleOrDefault(arg => arg.NameColon?.Name.Identifier.Text == NUnitFrameworkConstants.NameOfExpectedParameter) ?? arguments[0];
var actualArgument = arguments.SingleOrDefault(arg => arg.NameColon?.Name.Identifier.Text == NUnitFrameworkConstants.NameOfActualParameter) ?? arguments[1];
```

src/nunit.analyzers/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFix.cs
```diff
@@ -50,6 +59,16 @@ protected override void UpdateArguments(Diagnostic diagnostic, List<ArgumentSynt
             // Then we have to remove the 1st argument because that's now in the "Is.EqualTo()"
             arguments.RemoveAt(0);
```
manfred-brands, line 60: The `Insert(2)` might insert at the wrong spot.
This `RemoveAt` might remove the actual parameter instead of the expected one.

manfred-brands (changes requested):
Once `named` arguments are in play, none of the code using an index is valid.
See comment in ticket to change approach to replace underlying design approach.

Bartleby2718:
@manfred-brands I think there's a little more room for improvement, but I've pushed what I did today. Given the time difference, requesting a re-review now.

src/nunit.analyzers/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFix.cs
```diff
@@ -45,16 +50,18 @@ protected override void UpdateArguments(Diagnostic diagnostic, List<ArgumentSynt
...
+            var handledParameterNames = new[]
             {
-                arguments.RemoveAt(2);
-            }
+                NUnitFrameworkConstants.NameOfExpectedParameter,
+                NUnitFrameworkConstants.NameOfActualParameter,
+                NameOfDeltaParameter,
+            };
+            return arguments
+                .Concat(argumentNamesToArguments
```
manfred-brands, line 61: `arguments.AddRange` would be better.
However as the only remaining parameters for the _message_ would need the same treatment for all CodeFixes it is better dealt with in the caller.

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFix.cs
```diff
@@ -60,9 +60,13 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
             if (arguments.Count > 1)
                 arguments[1] = CastIfNecessary(arguments[1]);
 
+            var argumentNamesToArguments = arguments.ToDictionary(
+                argument => GetParameterName(argument),
```
manfred-brands, line 64: Calling GetParameter for every argument which goes through all arguments each time is not efficiënt. It would be better to loop through the arguments and use the index to obtain the parameter name.

manfred-brands (changes requested):
@Bartleby2718 Thanks for making the changed.
The actual codefix now uses parameter names iso indices is a good approach.

I put some remarks for changes in the PR.

src/nunit.analyzers/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFix.cs
Bartleby2718: @manfred-brands I think there was also `format`. Could you point me to the code that handles `message` and `format`?

Bartleby2718: I'm also curious why `AddRange` is better.

Bartleby2718: Duh, found the answer to my first question:
```cs
CodeFixHelper.UpdateStringFormatToFormattableString(arguments, this.MinimumNumberOfParameters);
```

manfred-brands: > I'm also curious why `AddRange` is better.

For one it is simpler. 
Second it doesn't have to copy the existing members (even though there are only two).
Third it doesn't allocate a complete new list but only extends the existing one (which might not require any allocations)

src/nunit.analyzers/Helpers/CodeFixHelper.cs
```diff
@@ -35,6 +35,42 @@ internal static class CodeFixHelper
...
+                return null;
+
+            // We only support converting if the format specification is a constant string.
+            if (messageAndParams.Count == 1 || formatSpecificationArgument is not LiteralExpressionSyntax literalExpression)
+                return messageAndParams[0];
+
+            var formatSpecification = literalExpression.Token.ValueText;
+
+            // var formatArgumentExpressions = new List<ExpressionSyntax>(capacity: messageAndParams.Count - 1);
+            var argsExpression = messageAndParams.Single(a => a != messageArgument).Expression;
```
manfred-brands, line 62: This becomes simpler when the argument only contains _args_.

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFix.cs
```diff
@@ -52,30 +54,65 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
 
-            // Now, replace the arguments.
-            List<ArgumentSyntax> arguments = invocationNode.ArgumentList.Arguments.ToList();
+            var methodSymbol = (IMethodSymbol)semanticModel.GetSymbolInfo(invocationNode).Symbol!;
 
-            // See if we need to cast the arguments when they were using a specific classic overload.
-            arguments[0] = CastIfNecessary(arguments[0]);
-            if (arguments.Count > 1)
-                arguments[1] = CastIfNecessary(arguments[1]);
+            Dictionary<string, ArgumentSyntax> argumentNamesToArguments = new();
```
manfred-brands, line 59: Extract the code below to a separate function to create a parameter dictionary from the invocation argument. The _args_ parameters can be returned as a separate list (tuple return).

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFix.cs
```diff
@@ -121,6 +158,13 @@ ArgumentSyntax CastIfNecessary(ArgumentSyntax argument)
             }
         }
 
+        protected virtual (ArgumentSyntax ActualArgument, ArgumentSyntax ConstraintArgument) UpdateArguments(
+            Diagnostic diagnostic,
+            IReadOnlyDictionary<string, ArgumentSyntax> argumentNamesToArguments)
+        {
+            throw new NotImplementedException($"Class must override {nameof(UpdateArguments)}");
```
manfred-brands, line 165: This causes all the other Classic CodeFixes to fail until they are updated.
Although once we have the framework in place, the update should be straight forward: Replacing indices access with parameter names.

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFix.cs
```diff
@@ -52,30 +54,65 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+            const string NameOfArgsParameter = "args";
+
+            // There can be 0 to any number of message and parameters.
+            List<ArgumentSyntax> messageAndParams = new();
+            var arguments = invocationNode.ArgumentList.Arguments.ToList();
+            for (var i = 0; i < arguments.Count; ++i)
+            {
+                var argument = arguments[i];
+                if (i < methodSymbol.Parameters.Length
+                    && (argument.NameColon?.Name.Identifier.Text ?? methodSymbol.Parameters[i].Name) is { } argumentName
```
manfred-brands, line 69:
```suggestion
                    && (argument.NameColon?.Name.Identifier.Text ?? methodSymbol.Parameters[i].Name) is string argumentName
```

I find the `is { }` not very readable. Specifying the expected type make it clearer.

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFix.cs
```diff
@@ -52,30 +54,65 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+            ArgumentSyntax actualArgument, constraintArgument;
 
             // Do the rule specific conversion
             if (typeArguments is null)
-                this.UpdateArguments(diagnostic, arguments);
+            {
+                (actualArgument, constraintArgument) = this.UpdateArguments(diagnostic, argumentNamesToArguments);
+                newArguments.Add(actualArgument);
+                newArguments.Add(constraintArgument);
+                if (CodeFixHelper.GetInterpolatedMessageArgumentOrDefault(messageAndParams) is { } interpolatedMessageArgument)
```
manfred-brands, line 101:
```suggestion
                if (CodeFixHelper.GetInterpolatedMessageArgumentOrDefault(messageAndParams) is ArgumentSyntax interpolatedMessageArgument)
```

manfred-brands (changes requested):
@Bartleby2718 I think you are on the correct path.
Change how to treat the _message_ parameter form List to Dictionary.
Some cleanups.

And then doing the other Classic Code Fixes ...

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFix.cs
Bartleby2718: @manfred-brands Yup, I'm in the process of updating all other `*CodeFix`es.

However, I'm not sure how many tests should be added to other `*CodeFixTests` files, although I'm sure we need _some_. When you get a chance, could you point out the list of tests that should be ported over (potentially after renaming) or added to other `*CodeFixTests` files?

manfred-brands:
> However, I'm not sure how many tests should be added to other *CodeFixTests files, although I'm sure we need some.

The exisiting tests should ensure that nothing gets broken. Then I suggest one new test per Codefix using named out of order parameters.

Bartleby2718:
@dotnet-policy-service agree

Bartleby2718:
Still working on the tests; there's a lot :)

src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
```diff
@@ -88,6 +120,24 @@ public void TestMethod()
...
+        {
+            var actual = 3d;
+            ↓ClassicAssert.AreEqual(delta: 0.0000001d, expected: 2d, actual: actual, args: new[] { ""Guid.NewGuid()"", Guid.NewGuid().ToString() }, message: ""message-id: {0}, {1}"");
+        }");
+            var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void TestMethod()
+        {
+            var actual = 3d;
+            Assert.That(actual: actual, Is.EqualTo(expected: 2d).Within(0.0000001d), $""message-id: {""Guid.NewGuid()""}, {Guid.NewGuid().ToString() }"");
+        }"); // TODO: Fix trailing trivia if possible
```
Bartleby2718, line 137: @manfred-brands Do you think this is worth looking into? I haven't spent too much time on this yet, but this is likely the same problem as https://github.com/nunit/nunit.analyzers/issues/713.

Bartleby2718: Ah, I think this was related https://github.com/nunit/nunit.analyzers/blob/master/src%2Fnunit.analyzers%2FHelpers%2FCodeFixHelper.cs#L44, but I'm not sure if this should be fixed in this PR or file another issue for this bug. I want to avoid a further scope creep.

src/nunit.analyzers/Helpers/CodeFixHelper.cs
```diff
@@ -35,6 +35,39 @@ internal static class CodeFixHelper
...
+            if (formatSpecificationArgument.IsKind(SyntaxKind.NullLiteralExpression))
+                return null;
+
+            // We only support converting if the format specification is a constant string.
+            if (args.Count == 0 || formatSpecificationArgument is not LiteralExpressionSyntax literalExpression)
+                return messageArgument;
+
+            var formatSpecification = literalExpression.Token.ValueText;
+
+            var argsExpression = args.Single(a => a != messageArgument).Expression;
```
Bartleby2718, line 59: Need to fix this logic

src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
Bartleby2718: Found the bug; it's simple enough that I'm okay with making it a part of this PR.

manfred-brands (changes requested):
Nice work!
A few changes to make "more standard" named parameters work.

I'm not convinced we should introduce named parameters for the constraint model, regardless if they were used in the classic model.
I rather drop all names (`.WithNameColon(null)`) than replace `arg1` with `actual`.

src/nunit.analyzers/Helpers/CodeFixHelper.cs
manfred-brands: I don't follow this logic:
1.  _messageArgument_ is not be part of _args_ anymore.
2. `Single` implies only one argument. This will fail in some case where named arguments are used *in order*:
`ClassicAssert.AreEqual(expected: 2.0, actual: 3.0, delta: 0.1, message: "Named parameters in order {0}, {1}", "first", "second");`
4.  The compiler doesn't allow unnamed argument when out of order parameters are used.  In that case _args_ would not have to be an array and this code should create `() => string.Format(messageArgument, args)` as per #720

src/nunit.analyzers.tests/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFixTests.cs
```diff
@@ -22,9 +22,10 @@ private sealed class TestableClassicModelAssertUsageCodeFix : ClassicModelAssert
         {
             public override ImmutableArray<string> FixableDiagnosticIds => ImmutableArray<string>.Empty;
 
-            protected override void UpdateArguments(Diagnostic diagnostic, List<ArgumentSyntax> arguments)
-            {
-            }
+            protected override (ArgumentSyntax ActualArgument, ArgumentSyntax? ConstraintArgument) ConstructActualAndConstraintArguments(
+                Diagnostic diagnostic,
+                IReadOnlyDictionary<string, ArgumentSyntax> argumentNamesToArguments) =>
+                (default!, default);
```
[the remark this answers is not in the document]
manfred-brands: Pushed up a commit for this.

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFix.cs
```diff
@@ -1,12 +1,14 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Reflection;
```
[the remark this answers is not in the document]
manfred-brands: Pushed up a commit for this.

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFix.cs
```diff
@@ -40,42 +42,43 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+            var (argumentNamesToArguments, args) = SplitUpOtherParametersAndParamParameter(semanticModel, invocationNode);
 
             // Remove null message to avoid ambiguous calls.
-            if (arguments.Count == 3 && arguments[2].Expression.IsKind(SyntaxKind.NullLiteralExpression))
+            if (argumentNamesToArguments.TryGetValue(NUnitFrameworkConstants.NameOfMessageParameter, out ArgumentSyntax? messageArgument)
+                && messageArgument.Expression.IsKind(SyntaxKind.NullLiteralExpression))
             {
-                arguments.RemoveAt(2);
+                argumentNamesToArguments.Remove(NUnitFrameworkConstants.NameOfMessageParameter);
             }
```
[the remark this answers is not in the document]
manfred-brands: Removed.

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFix.cs
```diff
@@ -86,11 +89,44 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
                     _ => Task.FromResult(context.Document.WithSyntaxRoot(newRoot)),
-                    this.Title), diagnostic);
+                    this.Title),
+                diagnostic);
+
+            (Dictionary<string, ArgumentSyntax> argumentNamesToArguments, List<ArgumentSyntax> args) SplitUpOtherParametersAndParamParameter(
+                SemanticModel semanticModel,
+                InvocationExpressionSyntax invocationNode)
+            {
+                var methodSymbol = (IMethodSymbol)semanticModel.GetSymbolInfo(invocationNode).Symbol!;
```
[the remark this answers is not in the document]
manfred-brands: Pushed up a commit for this.

src/nunit.analyzers/ClassicModelAssertUsage/ClassicModelAssertUsageCodeFix.cs
```diff
@@ -40,42 +42,43 @@ public sealed override async Task RegisterCodeFixesAsync(CodeFixContext context)
...
+            var (actualArgument, constraintArgument) = typeArguments is null
+                ? this.ConstructActualAndConstraintArguments(diagnostic, argumentNamesToArguments)
+                : this.ConstructActualAndConstraintArguments(diagnostic, argumentNamesToArguments, typeArguments);
+            newArguments.Add(actualArgument);
+            if (constraintArgument is not null)
+                newArguments.Add(constraintArgument);
+
             // Do the format spec, params to formattable string conversion
-            CodeFixHelper.UpdateStringFormatToFormattableString(arguments, this.MinimumNumberOfParameters);
+            if (CodeFixHelper.GetInterpolatedMessageArgumentOrDefault(messageArgument, args) is ArgumentSyntax interpolatedMessageArgument)
```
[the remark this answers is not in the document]
manfred-brands: Pushed up a commit for this.

src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
```diff
@@ -88,6 +120,24 @@ public void TestMethod()
...
+        }");
+            var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void TestMethod()
+        {
+            var actual = 3d;
+            Assert.That(actual: actual, Is.EqualTo(expected: 2d).Within(0.0000001d), $""message-id: {""Guid.NewGuid()""}, {Guid.NewGuid().ToString()}"");
+        }");
+            RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode, fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription);
+        }
+
```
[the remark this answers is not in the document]
manfred-brands: Pushed up a commit for this.

manfred-brands:
@Bartleby2718 I pushed up a commit addressing some of what I commented on in the PR.

Bartleby2718:
@manfred-brands Could you assign the PR to me?

manfred-brands:
I noticed that not all 'named parameters' where removed in the classic code fixes.
I added a commit to consistently use `.WithNameColon(null)` when obtaining parameters and updated the tests.

manfred-brands (approved):
Thanks for the cooperation.
I think we are there and the actual code fixes have become more straight forward.

Bartleby2718:
Thank you for all the guidance and explanations! I feel like I've learned a lot about both NUnit and Roslyn analyzers.

I have a couple more questions about this repo:
1. If I were to create another PR, should it target `master`? I wonder how you prevent merge conflicts if you don't merge PRs immediately.
2. When will the next version of NUnit.Analyzers be released? Will it be the last week of May to match the release schedule of NUnit (unless there is a patch released before that)?

manfred-brands:
> I have a couple more questions about this repo:
> 
> 1. If I were to create another PR, should it target `master`? I wonder how you prevent merge conflicts if you don't merge PRs immediately.

Normally a PR targets master. If PRs are for different areas there should normally not be a merge conflict.
If a PR build upon a previous one it can be on top of an other branch.

PRs are merged when approved. In this case I want to see if @mikkelbu has any comments before merging.

> 2. When will the next version of NUnit.Analyzers be released? Will it be the last week of May to match the release schedule of NUnit (unless there is a patch released before that)?

I didn't know there was a NUnit release schedule. I thought they are released when enough changes justify one. NUnit and the NUnit.Analyzers don't have to be in sync and are released independently at the moment.

src/nunit.analyzers.tests/ClassicModelAssertUsage/IsTrueAndTrueClassicModelAssertUsageCondensedCodeFixTests.cs
```diff
@@ -81,5 +81,45 @@ public void TestMethod()
...
+
+            var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+        public void TestMethod()
+        {{
+            ↓ClassicAssert.{assertion}(args: new[] {{ ""first"", ""second"" }}, message: ""{{0}}, {{1}}"", condition: true);
+        }}");
+            var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void TestMethod()
+        {
+            Assert.That(condition: true, $""{""first""}, {""second""}"");
```
mikkelbu, line 119: Isn't it wrong that we still have the `condition:` name here?

mikkelbu (commented):
I've only reviewed half the files, but so far it looks great. I won't have time to review the rest of the files before the end of the week as I've very little spare time currently - hopefully Thursday evening CET or perhaps Friday evening CET.

I've left a couple of comments, but nothing major.

And Manfred is correct about the release cadency of NUnit vs NUnit.Analyzers. We could release the Analyzers when this PR is done and all the current active PRs have been merged.

src/nunit.analyzers.tests/StringAssertUsage/StringAssertUsageCodeFixTests.cs
```diff
@@ -53,10 +53,19 @@ public void AnalyzeWhenFormatAndArgumentsAreUsed(string method)
...
 
-        private static string GetAdjustedConstraint(string method)
+        [TestCaseSource(nameof(StringAsserts))]
+        public void AnalyzeWhenFormatAndArgumentsAreUsedOutOfOrder(string method)
         {
-            return StringAssertUsageAnalyzer.StringAssertToConstraint[method]
-                                            .Replace("expected", "\"expected\"");
+            var firstParameterName = StringAssertUsageCodeFix.StringAssertToExpectedParameterName[method];
+            var code = TestUtility.WrapInTestMethod(@$"
+		    ↓StringAssert.{method}(args: new[] {{ ""first"", ""second"" }}, message: ""{{0}}, {{1}}"", actual: ""actual"", {firstParameterName}: ""expected"");");
```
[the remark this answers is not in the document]
manfred-brands: It is inside the string and when looking inside VS it aligned for me.

src/nunit.analyzers/ClassicModelAssertUsage/LessOrEqualClassicModelAssertUsageCodeFix.cs
```diff
@@ -16,19 +16,23 @@ public sealed class LessOrEqualClassicModelAssertUsageCodeFix
...
+            var constraintArgument = SyntaxFactory.Argument(
                 SyntaxFactory.InvocationExpression(
                     SyntaxFactory.MemberAccessExpression(
                         SyntaxKind.SimpleMemberAccessExpression,
                         SyntaxFactory.IdentifierName(NUnitFrameworkConstants.NameOfIs),
                         SyntaxFactory.IdentifierName(NUnitFrameworkConstants.NameOfIsLessThanOrEqualTo)))
                 .WithArgumentList(SyntaxFactory.ArgumentList(
-                    SyntaxFactory.SingletonSeparatedList(arguments[1])))));
+                    SyntaxFactory.SingletonSeparatedList(
+                        arg2Argument))));
```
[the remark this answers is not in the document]
manfred-brands: Thanks. So much for my consistency changes. Fixed.

src/nunit.analyzers.tests/ClassicModelAssertUsage/IsTrueAndTrueClassicModelAssertUsageCondensedCodeFixTests.cs
manfred-brands: Yes it is indeed. Missed that one. Fixed.

manfred-brands:
@mikkelbu Fixed your reported issues and found two items we forgot to clean up. I left them in as separate commits.

mikkelbu (approved):
Looks great @Bartleby2718 and @manfred-brands. I only noticed one little nit-pick otherwise I think we should merge.

I especially like how it simplified several of the existing analyzers/fixes.

Bartleby2718:
@mikkelbu @manfred-brands Addressed the last feedback. Thanks both for the feedback and answers and guidance!

PR #722, opened 2024-04-09
Use Markdown for CHANGES

Bartleby2718:
- Noticed that this repository uses `CHANGES.txt` even though its content is compatible with Markdown.
- Since the repository is hosted on GitHub, I added a `@` before each username.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,4 +1,4 @@
-﻿NUnit Analyzers 4.1 - March 16, 2024
+﻿# NUnit Analyzers 4.1 - March 16, 2024
 
 This release of the Analyzers extends the suppression of nullable warnings to also respect assumptions - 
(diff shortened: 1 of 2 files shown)


manfred-brands (changes requested):
Thanks for that.

When you rename a file, you need to make sure any references to it are also updated, e.g.:
https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/nunit.analyzers.nuspec#L22

```xml
<releaseNotes>See the release notes on https://github.com/nunit/nunit.analyzers/blob/master/CHANGES.txt.</releaseNotes>
```

Bartleby2718:
@manfred-brands That was already done in the initial commit :)

manfred-brands (approved):
Indeed it was. Sorry that I missed it.

mikkelbu (approved):
Thanks for the contribution @Bartleby2718. The change looks great. I'll merge it now

PR #730, opened 2024-04-14
NUnit2010: Exclusions

### LINKED ISSUE #728: NUnit2010 - do not consider Equals call if it doesn't override Object.Equals
If there is a class Foo, implementing Equals in this way:

 ```csharp
class Foo
{
    public bool Equals(Foo objectToCompare
    {
      bool areEqual;
      //some logic modifying areEqual;
      return areEqual;
    }
}
 ```

and Assert like:
 ```csharp
Foo expectedValue = new Foo();
Foo testeeResult = new Foo(); //for simplicity. Some calculation happened instead.
Assert.That(expectedValue.Equals(testeeResult));
 ```

NUnit2010 fix changes it to 
 ```csharp
Assert.That(expectedValue, Is.EqualTo(testeeResult));
 ```

But this test fails with error message:
Expected: Foo
But was: Foo

It would work, if Foo would implement public override bool Equals(object objectToCompare) instead. But this change is not possible for legacy code reason.

It would probably better, if the Analyzer would check, if the called Equals method is an override from object or just an isolated implementation and ignore it in such case.

manfred-brands:
Fixes #728 

We cannot convert `a.Equals(b)` into `a, Is.EqualTo(b)` if a doesn't implement `IEquatable<b>`

### DIFF
--- a/src/nunit.analyzers.tests/ConstraintsUsage/EqualConstraintUsageAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/ConstraintsUsage/EqualConstraintUsageAnalyzerTests.cs
@@ -305,5 +305,108 @@ namespace NUnit.Analyzers.Tests.ConstraintsUsage
 
             RoslynAssert.Valid(analyzer, testCode, Settings.Default.WithMetadataReferences(metadataReferences));
         }
+
+        [Test]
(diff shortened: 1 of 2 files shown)


mikkelbu (approved):
Again a nice improvement. Thanks @manfred-brands 👍

Issue #729, opened 2024-04-12
NUnit2050 should consider conditional operator


matode:
If the Assert looks like this:
```csharp
Assert.That(foo, Is.True, "Some message {0} correct ", someCheck ? "not" : "");
```
fix for NUnit2050 modifies it to
```csharp

Assert.That(foo, Is.True, $"Some message {someCheck ? "not" : ""} correct");
```

But the result is wrong and doesn't compile. It should enclose the conditional check into brackets:
```csharp
Assert.That(foo, Is.True, $"Some message {(someCheck ? "not" : "")} correct");
```

manfred-brands:
There will be more examples of code that is not allowed inside an `FormattableString`:
I'll fix them when reported.

manfred-brands:
@matode I was going to look into it and notice there was already code in place in the analyzer for this, but no NUnit test.
However adding the test shows the code is working correctly.

I then put your example in a new project and it correctly inserts the parenthesis around the conditional expression.
Can you provide me with a full example where it doesn't work as intended?

Bartleby2718:
The logic was added in 4.1.0: https://github.com/nunit/nunit.analyzers/commit/202f208b19498142291c14a65debb4ae4d70d9ab#diff-63af8d52a7383b035c890d3ac15699e7a330f5c2c8cee83fdff269ee3cbc1b2dR119-R126

so it may be worth checking whether you're using 4.1.0.

mikkelbu:
> The logic was added in 4.1.0: [202f208#diff-63af8d52a7383b035c890d3ac15699e7a330f5c2c8cee83fdff269ee3cbc1b2dR119-R126](https://github.com/nunit/nunit.analyzers/commit/202f208b19498142291c14a65debb4ae4d70d9ab#diff-63af8d52a7383b035c890d3ac15699e7a330f5c2c8cee83fdff269ee3cbc1b2dR119-R126)
> 
> so it may be worth checking whether you're using 4.1.0.

Actually, the diff above is already included in [3.10.0](https://github.com/nunit/nunit.analyzers/releases/tag/3.10.0)
![image](https://github.com/nunit/nunit.analyzers/assets/1007631/2655f8b8-2dd0-4676-81ea-3515b7a9018e)

manfred-brands:
@matode Can we close this or do you see a situation where this generates non-compilable code?

matode:
All good. The reported use case is not reproducible with latest version. 
Issue can be closed.

[...]

PR #731, opened 2024-04-14
Add test for wrapping conditional expression in parenthesis for NUnit2050

manfred-brands:
Adds a test to proof #729 works as intended.


mikkelbu (approved):
LGTM. Thanks for the test and the fix of typo @manfred-brands

PR #733, opened 2024-04-23
Missing full stops added to NUnit2050

maettu-this:
Full stops were missing at two locations.


mikkelbu (approved):
Thanks for the contribution @maettu-this. I'll merge it now.

PR #738, opened 2024-05-03
Move NUnit4 paragraph under table.

### LINKED ISSUE #734: Why dropping composite message formatting support also for Assert.Pass/Fail/... where there are no actual and constraint parameters?
[NUnit2050](https://github.com/nunit/nunit.analyzers/blob/master/documentation/NUnit2050.md) tells me to replace composite message formatting by a simple string literal (not possible where variable values need to be injected) or an interpolatable string (would be possible but adds too much noise as variable names are much longer than {0} or {1}).

I fully understand the benefits of using `CallerArgumentExpression` for `Assert.That` and `Assume.That`. But why also for...
```
Assert.Pass
Assert.Fail
Assert.Warn
Assert.Ignore
Assert.Inconclusive
```
...where there neither is a `condition` nor `actual`/`constraint` parameters?

manfred-brands:
Fixes #734 

The table belongs to the method that have 'params' overloads removed.

### DIFF
--- a/documentation/NUnit2050.md
+++ b/documentation/NUnit2050.md
@@ -30,12 +30,6 @@ This analyzer needs to be run when still building against NUnit3 as otherwise yo
 When usages of the new methods with `params` are detected, the associated CodeFix will convert the format specification
 into an interpolated string.
 
-Once you moved to NUnit4 the analyzer has some limited functionality as there are a few
-cases where your NUnit3 code will compile on NUnit4, but not the way you want it.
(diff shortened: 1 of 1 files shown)


maettu-this (changes requested):
Shouldn't `Assume.That` be added as well?

manfred-brands:
> Shouldn't `Assume.That` be added as well?

Added.

mikkelbu (approved):
LGTM

PR #740, opened 2024-05-17
Recognize Fully Qualified NUnit.Framework.Assert.That

### LINKED ISSUE #739: Null suppression does not work when `Assert` is fully qualified
The following code has a null warning on `M(x)`

```csharp
using NUnit.Framework;

class C
{
    [Test]
    public void T()
    {
        object? x = null;
        NUnit.Framework.Assert.That(x, Is.Not.Null);
        M(x);
    }

    void M(object o)
    {
    }
}
```

The following code does not
```csharp
using NUnit.Framework;

class C
{
    [Test]
    public void T()
    {
        object? x = null;
        Assert.That(x, Is.Not.Null);
        M(x);
    }

    void M(object o)
    {
    }
}
```

manfred-brands:
Fixes #739


mikkelbu (approved):
LGTM. Thanks for the contribution @manfred-brands

PR #744, opened 2024-05-28
Recognize "(field as IDisposable)?.Dispose()"

### LINKED ISSUE #743: NUnit1032 (missing Dispose), if dispose is wrapped in "(… as IDisposable)?.Dispose()"
OK, basically I had a test class. In the Test-Class was a member "IServiceProvider _serviceProvider". In the [TearDown]-method the service provider was disposed with the following code:

```csharp
(_serviceProvider as IDisposable)?.Dispose();
_serviceProvider = null;
```

When compiled I got the above mentioned error:  `Error NUnit1032: The field _serviceProvider should be Disposed in a method annotated with [TearDownAttribute] (https://github.com/nunit/nunit.analyzers/tree/master/documentation/NUnit1032.md)`

1. `IServiceProvider` doesn’t inherit the `IDisposable` interface, hence the `as` operator in the TearDown method.
2. I explicitly call Dispose() in the TearDown method.

So why do I get this error message!? This is probably a bug.

So for reference, this is the NuGet Config:
```xml
<PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.9.0" />
<PackageReference Include="NUnit" Version="4.1.0" />
<PackageReference Include="NUnit3TestAdapter" Version="4.5.0" />
<PackageReference Include="NUnit.Analyzers" Version="4.2.0">
    <PrivateAssets>all</PrivateAssets>
    <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
</PackageReference>
<PackageReference Include="coverlet.collector" Version="6.0.2">
    <PrivateAssets>all</PrivateAssets>
    <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
</PackageReference>
```

manfred-brands:
Fixes #743 

Also adds recognition of `System.IDisposable` in addition to `IDisposable`

### DIFF
--- a/src/nunit.analyzers/DisposeFieldsAndPropertiesInTearDown/DisposeFieldsAndPropertiesInTearDownAnalyzer.cs
+++ b/src/nunit.analyzers/DisposeFieldsAndPropertiesInTearDown/DisposeFieldsAndPropertiesInTearDownAnalyzer.cs
@@ -480,8 +480,7 @@ namespace NUnit.Analyzers.DisposeFieldsInTearDown
                     //     disposable.Dispose();
                     if (ifStatement.Condition is IsPatternExpressionSyntax isPatternExpression &&
                         isPatternExpression.Pattern is DeclarationPatternSyntax declarationPattern &&
-                        declarationPattern.Type is IdentifierNameSyntax identifierName &&
-                        identifierName.Identifier.Text.EndsWith("Disposable", StringComparison.Ordinal) &&
+                        IsDisposable(declarationPattern.Type) &&
                         declarationPattern.Designation is SingleVariableDesignationSyntax singleVariableDesignation)
                     {
                         string? member = GetIdentifier(isPatternExpression.Expression);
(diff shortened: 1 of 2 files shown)


src/nunit.analyzers/DisposeFieldsAndPropertiesInTearDown/DisposeFieldsAndPropertiesInTearDownAnalyzer.cs
```diff
@@ -555,19 +554,60 @@ private static void DisposedIn(Parameters parameters, HashSet<string> disposals,
...
 
-            // considering cast to IDisposable, e.g. in case of explicit interface implementation of IDisposable.Dispose()
-            else if (expression is ParenthesizedExpressionSyntax parenthesizedExpression &&
-                     parenthesizedExpression.Expression is CastExpressionSyntax castExpression &&
-                     castExpression.Expression is IdentifierNameSyntax castIdentifierName &&
-                     castExpression.Type is IdentifierNameSyntax typeIdentifierName &&
-                     typeIdentifierName.Identifier.Text.Equals("IDisposable", StringComparison.Ordinal))
+            // considering cast to I(Async)Disposable, e.g. in case of explicit interface implementation of IDisposable.Dispose()
+            // or in case of 'as IDisposable' or 'as IAsyncDisposable'
+            // or in case of 'is IDisposable' or 'is IAsyncDisposable'
```
mikkelbu, line 559: Mostly a comment - as I think the code below looks good for cast and `as`-pattern, but I don't think `is` is handled here (only in lines 481-486) or am I overlooking something?

mikkelbu (approved):
Looks good to me. I only have a single question, but I'm happy to merge it when you are. 

As usual a nice and simple addition @manfred-brands

src/nunit.analyzers/DisposeFieldsAndPropertiesInTearDown/DisposeFieldsAndPropertiesInTearDownAnalyzer.cs
manfred-brands: No you are right, I added the comment until my nunit tests proofed me wrong that it was done when checking `if` statement.

PR #757, opened 2024-06-07
NUnit1032: Explain how to dispose fields when using InstancePerTestCase

### LINKED ISSUE #756: error NUnit1032 is incorrect when InstancePerTestCase and constructor is used to initialize IDisposible
If you have the following:

- InstancePerTestCase
- An IDisposible created in your constructor 
- [TearDown] method which disposes

Then you get the following error:
error NUnit1032: The field _yourService should be Disposed in a method annotated with [OneTimeTearDownAttribute] (https://github.com/nunit/nunit.analyzers/tree/master/documentation/NUnit1032.md)

However [OneTimeTearDownAttribute] must now be a static method and the IDisposible isn't and shouldn't be so it can't dispose it.

It seems this error doesn't know about InstancePerTestCase and so gets this wrong.

And no I don't want to create the IDisposible in a [Setup] method instead of the constructor. Putting it in the constructor allows me to make in non nullable and readonly. Which means I don't need null checks in every test.

manfred-brands:
Fixes #756

### DIFF
--- a/documentation/NUnit1032.md
+++ b/documentation/NUnit1032.md
@@ -14,9 +14,11 @@
 
 An IDisposable field/property should be Disposed in a TearDown method.
 
-This analyzer rule only applies to a `TestFixture` which is using the default NUnit `SingleInstance` life cycle where the class is instantiated once for all tests.
+This analyzer rule only applies to a `TestFixture` which is using the default
+NUnit `SingleInstance` life cycle where the class is instantiated once for all tests.
 
-If you are using `LifeCycle.InstancePerTestCase` you should dispose the fields/properties in the `Dispose` method of the test class.
+If you are using `LifeCycle.InstancePerTestCase` you should dispose the fields/properties
+in the `Dispose` method of the test class.
 
 ## Motivation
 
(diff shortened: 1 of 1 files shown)


documentation/NUnit1032.md
```diff
@@ -14,16 +14,52 @@
...
+
+If you have `IDisposable` fields or properties, your class must implement the
+[`IDisposable`](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable?view=net-8.0) interface.
+
+Dispose any fields/properties that are initialized at declaration or in the constructor in the
+[`Dispose`](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose?view=net-8.0) method.
+
+The NUnit.Analyzer will not help you here as the functionality is available in a [Microsoft .NET Analyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.NetAnalyzers)
+
+These are the rules that will help you with this:
```
mikkelbu, line 45: Just some suggestions:
* Removed an "a"
* Added period and merged it into one paragraph.

```suggestion
The NUnit.Analyzer will not help you here as the functionality is available in [Microsoft .NET Analyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.NetAnalyzers). These are the rules that will help you with this:
```

mikkelbu (approved):
Looks good - I have some suggestions, but nothing major. Happy to merge when you are 👍

manfred-brands:
Thanks @mikkelbu

PR #761, opened 2024-06-11
Update nunit.analyzers.nuspec to specify that NUnit.Analyzers v4 is intended to be used with NUnit 4.

andrewimcclement:


Bartleby2718:
`NUnit.Analyzers` v4 can still be used to help folks _transition_ from NUnit v3 to NUnit v4, so I'm not sure if I agree with this.

andrewimcclement:
It can, but it references classes like ClassicAssert which only exist in NUnit v4, so the messages are often incorrect, even if the majority of quick fixes will be fine.

manfred-brands:
The analyzer works with both NUnit 3 and 4. It detects which version if NUnit is referenced in your project and changes behaviour accordingly.

It has to be used with NUnit 3 to prepare moving to NUnit 4 as the latter has breaking changes, especially regarding message formatting.

andrewimcclement:
By all means suggest an alternative phrasing - I feel v4 is at least more accurate than v3, but we may be able to do better.

andrewimcclement:
We could just say `3+` instead of `3`? Or maybe it is worth being explicit - `for use with v4 or for the purposes of migrating from v3 to v4` or something?

manfred-brands:
3+ is fine. Also mentioning the migration is good.

Where do you say it references `ClassicAssert` when using with v3?
If you mean the `Assert.AreEqual`, those conversion analyzers existed way before there was a `ClassicAssert` class.

The {v4 migration guide](https://docs.nunit.org/articles/nunit/release-notes/Nunit4.0-MigrationGuide.html#convert-classic-assert-to-constraint-model) says:

> Caveat: The analyzers only run when the code compiles, so execute and act on the analyzer before upgrading nunit to version 4.0.0!

andrewimcclement:
Oh sure, the analyzers existed, it's just they give an incorrect message (precisely because the message talks about the ClassicAssert class, which doesn't exist in v3). It's just a small peeve.

Bartleby2718 (approved):

manfred-brands (approved):
I'm ok with these changes

PR #773, opened 2024-07-26
Add support for checking TestCase<> parameters

### LINKED ISSUE #767: Augment NUnit1001 to recognized and check generic TestCase attributes
https://github.com/nunit/nunit/issues/4632 added support for generic test case parameters:

NUnit1001 detects if `TestCase` arguments don't match parameters.

```csharp
[TestCase(1)]
public void SomeTest(string s)
{
}
```

This has to be augmented to support generic versions:

```csharp
[TestCase<double>(1)]
public void SomeTest(string s)
{
}
```

manfred-brands:
Fixes #767 

Actual Analyzer code is finished.

Testcode can be simplified once NUnit 4.2.0 is released.

### DIFF
--- a/documentation/NUnit1001.md
+++ b/documentation/NUnit1001.md
@@ -29,6 +29,12 @@ public void SampleTest(int numberValue)
 {
     Assert.That(numberValue, Is.EqualTo(1));
 }
+
+[TestCase<double>(42)]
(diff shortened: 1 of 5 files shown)


mikkelbu (approved):
LGTM. Thanks for the contribution @manfred-brands

PR #774, opened 2024-07-26
Prevent exception when fields are declared in a different source file.

### LINKED ISSUE #766: Error when TearDown method is defined in partial test classes - Syntax node is not within syntax tree
When a test class is separated into different files via partial classes and contain a method marked with TearDown attribute, I get the following error:

    Error AD0001: Analyzer 'NUnit.Analyzers.DisposeFieldsInTearDown.DisposeFieldsAndPropertiesInTearDownAnalyzer' threw an exception of type 'System.ArgumentException' with message 'Syntax node is not within syntax tree'.

When I move the TearDown method into a base class, everything works.

manfred-brands:
Fixes #766


mikkelbu (approved):
LGTM - thanks for the fix @manfred-brands

PR #784, opened 2024-09-16
Fix trivia for NUnit2046

verdie-g:
```csharp
Assert.That(array.Length,
    Is.EqualTo(1));
```
would give
```csharp
Assert.That(array,
Has.Length.EqualTo(1));
```


manfred-brands (approved):
Thanks again @verdie-g for your contribution.

Issue #759, opened 2024-06-09
New diagnostic: Consider using Assert.ThatAsync instead of Assert.That


Bartleby2718:
Arguably, one of the biggest advantages of using NUnit 4 is: [Proper async/await, you can now await the Asserts](https://docs.nunit.org/articles/nunit/Towards-NUnit4.html#proper-asyncawait-you-can-now-await-the-asserts). However, it appears that `NUnit.Analyzers` doesn't have an analyzer or a codefix for that.

Consider the following sync-over-async tests:
```cs
[Test]
public async Task Before()
{
    Assert.That(await GetFourtyTwoAsync(), Is.EqualTo(42));
    Assert.That(await GetFourtyTwoAsync("hello world"), Is.EqualTo(42));
}

private static Task<int> GetFourtyTwoAsync() => Task.FromResult(42);

private static Task<int> GetFourtyTwoAsync(string stringToPrintBeforehand)
{
    Console.WriteLine(stringToPrintBeforehand);
    return Task.FromResult(42);
}
```

It'd be great if these could be fixed to:
```cs
[Test]
public async Task After()
{
    await Assert.ThatAsync(() => GetFourtyTwoAsync(), Is.EqualTo(42)); // the first argument could be simplified to just GetFourtyTwoAsync
    await Assert.ThatAsync(() => GetFourtyTwoAsync("hello world"), Is.EqualTo(42));
}
```

Personally, I think it'd be neat if we could bundle all NUnit 4-related changes to NUnit4XXX. Thoughts?

manfred-brands:
The original test more likely looks like:
```csharp
[Test]
public void Before()
{
    Assert.That(await GetFourtyTwoAsync(), Is.EqualTo(42));
    Assert.That(await GetFourtyTwoAsync("hello world"), Is.EqualTo(42));
}
```

So any CodeFix would also possibly have to change the method declaration.

manfred-brands:
Note that these are bad examples as the `Assert.That` doesn't do any async-over-sync.
The method is called asynchronously and the result is passed to the assert.

The only place where you would want to use `Assert.ThatAsync` is when NUnit is calling the delegate.
This would only be the case for two asserts:
1. ThrowsConstraints where NUnit need to call the method in order to catch any exception.
2. DelayedConstraint where NUnit needs to call the method repeatedly to verify the constraint.

manfred-brands:
[...]

PR #798, opened 2024-12-03
WIP

Bartleby2718:
@manfred-brands @mikkelbu I've been looking into #759 in the past couple of days, but I'm getting the following error message in tests:
> CS0103 The name 'Assert.ThatAsync' does not exist in the current context
  at line 24 and character 18 in file TestClass.cs | await ↓Assert.ThatAsync(() => GetResultAsync(), Is.EqualTo(42));

**I suppose this is because I didn't restrict the Analyzer/CodeFix to NUnit 4+. Is it possible to do so?** If not, should I consider this not startable yet?

I tried updating `TestUtility.WrapMethodInClassNamespaceAndAddUsings()` to add `#if NUNIT4` like this, but I'm still getting the error:
```cs
    #pragma warning disable CS8019
    using System;
    using System.Threading.Tasks;
    using NUnit.Framework;
    #if NUNIT4
    using NUnit.Framework.Legacy;
    #else
    using ClassicAssert = NUnit.Framework.Assert;
    #endif
    
    #pragma warning restore CS8019
    
    
    namespace NUnit.Analyzers.Tests.Targets.TestCaseUsage
    {
        [TestFixture]
        public class TestClass
        {
    #if NUNIT4
    
            [Test]
            public async Task Test1()
            {
                await Assert.ThatAsync(() => GetResultAsync(), Is.EqualTo(42));
            }
    
            private static Task<int> GetResultAsync() => Task.FromResult(42);
    #endif
        }
    }
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,7 +45,6 @@ jobs:
         uses: actions/setup-dotnet@main
         with:
           global-json-file: ./global.json
-          dotnet-version: 8.x
 
       - name: dotnet test (NUnit4)
         run: dotnet test --configuration=Release -p:NUnitVersion=4 ./src/nunit.analyzers.tests/
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,7 @@ jobs:
         uses: actions/setup-dotnet@main
         with:
           global-json-file: ./global.json
+          dotnet-version: 8.x
 
       - name: dotnet test (NUnit4)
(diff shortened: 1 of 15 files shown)


manfred-brands:
> I suppose this is because I didn't restrict the Analyzer/CodeFix to NUnit 4+.

Correct and yes this is possible.
Your analyzer needs to override the `AnalyzAssertInvocation` that has the `Version nunitVersion` parameter.
See `UpdateStringFormatToInterpolatableStringAnalyzer` for an example.
And then only continue if the user uses NUnit4: `nunitVersion.Major > 4`

PR #803, opened 2024-12-05
Update DisposeFieldsAndPropertiesInTearDownAnalyzer.cs to support Quit()

rhyous:
Update DisposeFieldsAndPropertiesInTearDownAnalyzer.cs to support Quit() as in Selenium that calls Dispose().

### DIFF
--- a/src/nunit.analyzers/DisposeFieldsAndPropertiesInTearDown/DisposeFieldsAndPropertiesInTearDownAnalyzer.cs
+++ b/src/nunit.analyzers/DisposeFieldsAndPropertiesInTearDown/DisposeFieldsAndPropertiesInTearDownAnalyzer.cs
@@ -22,7 +22,8 @@ namespace NUnit.Analyzers.DisposeFieldsInTearDown
             "Dispose",
             "DisposeAsync",
             "Close",
-            "CloseAsync");
+            "CloseAsync",
+            "Quit");
 
         // Types that even though they are IDisposable, don't need to be Disposed.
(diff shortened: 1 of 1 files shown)


rhyous:
@dotnet-policy-service agree [company="{Rhyous}"]

mikkelbu:
Hi @rhyous 

Thanks for the contribution. As far as I can remember we chose the current method names to match the CA2000 rule. I'm not against adding `Quit` to the list, but are there any special case you are considering? As far as I can tell `Quit` is used by selenium-webdriver, but I'm unsure if there are other cases.

rhyous:
Cool.

Yes, I am using Selenium. I also don't know of any other cases.

FYI, I checked for any pull requests for adding Quit() already, in case it had already been rejected, and I didn't see one. So I posted this suggestion/pull request and the unit team could either say yes, this is good, or no.

For now, I'm suppressing this. I'll accept whatever you think is best.

manfred-brands:
This seems to be specific to a library you use.
Have you tried setting `dotnet_diagnostic.NUnit1032.additional_dispose_methods: Quit` in your `.editorconfig`?

rhyous:
@M

> This seems to be specific to a library you use. Have you tried setting `dotnet_diagnostic.NUnit1032.additional_dispose_methods: Quit` in your `.editorconfig`?

I didn't try that. I just suppressed it. I will try that. Thanks!

mikkelbu:
@rhyous No need for apologies. We have forgotten to add this information to the docs. I'll close this PR and then open a ticket about adding this to the docs of NUnit1032

PR #804, opened 2024-12-06
Use Assert.ThatAsync

### LINKED ISSUE #759: New diagnostic: Consider using Assert.ThatAsync instead of Assert.That

Bartleby2718:
This closes #759.

Please assign the issue and the PR to me!

### DIFF
--- a/src/nunit.analyzers.tests/UseAssertThatAsync/UseAssertThatAsyncCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/UseAssertThatAsync/UseAssertThatAsyncCodeFixTests.cs
@@ -1,0 +1,154 @@
+#if NUNIT4
+using Gu.Roslyn.Asserts;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.Diagnostics;
+using NUnit.Analyzers.Constants;
+using NUnit.Analyzers.UseAssertThatAsync;
+using NUnit.Framework;
+
+namespace NUnit.Analyzers.Tests.UseAssertThatAsync;
+
+[TestFixture]
+public sealed class UseAssertThatAsyncCodeFixTests
+{
+    private static readonly DiagnosticAnalyzer analyzer = new UseAssertThatAsyncAnalyzer();
+    private static readonly CodeFixProvider fix = new UseAssertThatAsyncCodeFix();
+    private static readonly ExpectedDiagnostic diagnostic = ExpectedDiagnostic.Create(AnalyzerIdentifiers.UseAssertThatAsync);
+    private static readonly string[] configureAwaitValues =
+    {
+        "",
+        ".ConfigureAwait(true)",
+        ".ConfigureAwait(false)",
+    };
+
+    [Test]
+    public void VerifyGetFixableDiagnosticIds()
+    {
+        var fix = new UseAssertThatAsyncCodeFix();
+        var ids = fix.FixableDiagnosticIds;
+
+        Assert.That(ids, Is.EquivalentTo(new[] { AnalyzerIdentifiers.UseAssertThatAsync }));
+    }
+
+    [Test]
+    public void VerifyIntAndConstraint([ValueSource(nameof(configureAwaitValues))] string configureAwait, [Values] bool hasMessage)
+    {
+        var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@$"
+        public async Task Test()
+        {{
+            Assert.That(await GetIntAsync(){configureAwait}, Is.EqualTo(42){(hasMessage ? @", ""message""" : "")});
+        }}
+
+        private static Task<int> GetIntAsync() => Task.FromResult(42);");
+        var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
+        public async Task Test()
+        {{
+            await Assert.ThatAsync(() => GetIntAsync(), Is.EqualTo(42){(hasMessage ? @", ""message""" : "")});
+        }}
+
+        private static Task<int> GetIntAsync() => Task.FromResult(42);");
+        RoslynAssert.CodeFix(analyzer, fix, diagnostic, code, fixedCode);
+    }
+
+    [Test]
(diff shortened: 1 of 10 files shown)


src/nunit.analyzers.tests/UseAssertThatAsync/UseAssertThatAsyncCodeFixTests.cs
```diff
@@ -0,0 +1,154 @@
...
+        var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public async Task Test()
+        {
+            await Assert.ThatAsync(() => GetBoolAsync(), Is.True, ""message"");
+        }
+
+        private static Task<bool> GetBoolAsync() => Task.FromResult(true);");
+        RoslynAssert.CodeFix(analyzer, fix, diagnostic, code, fixedCode);
+    }
+}
```
manfred-brands, line 153: Please add an example for detecting:
```csharp
public void Test()
{
    Assert.That(() => Task.FromException<int>(new InvalidOperationException()), Throws.InvalidOperationException);
}
```
With a code fix looking like:
```csharp
public async Task Test()
{
    await Assert.ThatAsync(() => Task.FromException<int>(new InvalidOperationException()), Throws.InvalidOperationException);
}
```

Especially notice the need to change the return type of the test from `void` to `async Task`.

src/nunit.analyzers/UseAssertThatAsync/UseAssertThatAsyncAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+            return;
+
+        if (assertOperation.TargetMethod.Name != NUnitFrameworkConstants.NameOfAssertThat)
+            return;
+
+        var arguments = assertOperation.Arguments
+            .Where(a => a.ArgumentKind == ArgumentKind.Explicit) // filter out arguments that were not explicitly passed in
+            .ToArray();
+
+        // The first parameter is usually the "actual" parameter, but sometimes it's the "condition" parameter.
```
manfred-brands, line 44: The only argument name we want to catch is actually called `del` as used in:
` public static void That<TActual>(ActualValueDelegate<TActual> del, IResolveConstraint expr`

src/nunit.analyzers/UseAssertThatAsync/UseAssertThatAsyncAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+        // Since the order is not guaranteed, let's just call it "actualArgument" here.
+        var actualArgument = arguments.SingleOrDefault(a => firstParameterCandidates.Contains(a.Parameter?.Name))
+            ?? arguments[0];
+        if (actualArgument.Syntax is not ArgumentSyntax argumentSyntax || argumentSyntax.Expression is not AwaitExpressionSyntax awaitExpression)
```
manfred-brands, line 48: When the argument is _awaited_ then NUnit doesn't do async-over-sync.

manfred-brands (changes requested):
Although the PR fixes what #759 describes it is for the wrong reason.
None of the examples do async-over-sync.

It fixes a style issue and at most should be a `suggestion` or even disabled by default.
This allows you to use this rule, but don't force it on others.
If you were mistaken and you thought this was doing synchronous async we could drop it.

What we could use is a rule that detects the real use-cases where we do synchronous async calls which is the examples mentioned in my comments.
This could be a separate diagnostic supported by the same Analyzer/CodeFix pair.

Bartleby2718:
Thanks for the quick feedback @manfred-brands!

Besides the overload you mentioned [here](https://github.com/nunit/nunit.analyzers/pull/804#discussion_r1872469762), am I correct in assuming that the following overloads should also be converted?
```cs
public static void That(Func<bool> condition, NUnitString message, string actualExpression)
public static void That(TestDelegate code, IResolveConstraint constraint, NUnitString message, string actualExpression, string constraintExpression)
```

manfred-brands:
> Thanks for the quick feedback @manfred-brands!
> 
> Besides the overload you mentioned [here](https://github.com/nunit/nunit.analyzers/pull/804#discussion_r1872469762), am I correct in assuming that the following overloads should also be converted?
> 
> ```cs
> public static void That(Func<bool> condition, NUnitString message, string actualExpression)
> public static void That(TestDelegate code, IResolveConstraint constraint, NUnitString message, string actualExpression, string constraintExpression)
> ```

The first no. As NUnit calls `condition.Invoke()` and the code wouldn't accept an `Func<Task<bool>>`.
The second yes. So also test for parameter name _code_.

manfred-brands:
@Bartleby2718 Will you continue with this PR or closes it as the use-cases you highlighted were not actually doing async-over-sync.

Bartleby2718:
@manfred-brands I don't think I'll be able to get to this soon, so I've closed the PR.

PR #805, opened 2024-12-07
fix(BaseAssertionAnalyzer): Don't fail if there are multiple references to NUnit

### LINKED ISSUE #794: AD0001: Occasional InvalidOperationException error in analyzer in IDE
I'm not entirely sure what reproduces this, however I do occasionally get a ton of these analyzer errors in the same file:
```
AD0001: Analyzer 'NUnit.Analyzers.ConstActualValueUsage.ConstActualValueUsageAnalyzer' threw an exception of type 'System.InvalidOperationException' with message 'Sequence contains more than one matching element'. Exception occurred with following context: Compilation: ConfigService.Tests System.InvalidOperationException: Sequence contains more than one matching element at System.Linq.ThrowHelper.ThrowMoreThanOneMatchException() at System.Linq.Enumerable.TryGetSingle[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found) at NUnit.Analyzers.BaseAssertionAnalyzer.AnalyzeCompilationStart(CompilationStartAnalysisContext context) at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock[TArg](DiagnosticAnalyzer analyzer, Action`1 analyze, TArg argument, Nullable`1 info, CancellationToken cancellationToken) in Z:\BuildAgent\work\3b7ce003563d6f8f\src\Compilers\Core\Portable\DiagnosticAnalyzer\AnalyzerExecutor.cs:line 1191 ----- Suppress the following diagnostics to disable this analyzer: NUnit2007
```

It only seems to happen in my IDE (At least I have never seen it happen when doing `dotnet build`), and it usually goes away pretty fast. I would guess it happens because of some code that is invalid for a short moment. When the error occurs the entire screen overflows with errors at once. 

I can't send you a file it happens on, as I cannot share code from work, but it seems to usually be larger test files.

NUnit version: 4.2.2
NUnit.Analyzers version: 4.4.0

Using JetBrains Rider with dotnet version 8.0.303.

I'm sorry that it's not a whole lot to work from :(

mikkelbu:
Fixes #794

### DIFF
--- a/src/nunit.analyzers/BaseAssertionAnalyzer.cs
+++ b/src/nunit.analyzers/BaseAssertionAnalyzer.cs
@@ -4,6 +4,7 @@ using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 using Microsoft.CodeAnalysis.Operations;
+using NUnit.Analyzers.Constants;
 using NUnit.Analyzers.Extensions;
 
 namespace NUnit.Analyzers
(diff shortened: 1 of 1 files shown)


manfred-brands (approved):
Looks good. I had forgotten about this one.

mikkelbu:
I tried to reproduce the problem, but I was not able to do so, so hopefully this fixes the problem

PR #812, opened 2024-12-22
Add support for new Equal constraints and overloads from NUnit 4.3.1

### LINKED ISSUE #811: NUnit2021 ignores cast operation since NUnit 4.3.1
It seems that the changes in https://github.com/nunit/nunit/pull/4899 caused my test code to have build error NUnit2021.
This occurs in some of the test codes like following. (Not all, it seems.)
My codes can build successfully at 36b8b1d37ce76795272b0df4b7b279f8c3cae93c.

```cs
[Test]
public void Test()
{
  MyLong l0 = new(0);

  Assert.That(l0, Is.EqualTo((MyLong)0));
  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ error NUnit2021: The EqualTo constraint always fails as the actual and the expected value cannot be equal (https://github.com/nunit/nunit.analyzers/tree/master/documentation/NUnit2021.md)
}
```

I will report again when the situation can be clearly reproduced.

_Originally posted by @smdn in https://github.com/nunit/nunit/issues/4899#issuecomment-2551438500_

manfred-brands:
Fixes #811

### DIFF
--- a/src/nunit.analyzers.tests/Constants/NUnitFrameworkConstantsTests.cs
+++ b/src/nunit.analyzers.tests/Constants/NUnitFrameworkConstantsTests.cs
@@ -251,6 +251,12 @@ namespace NUnit.Analyzers.Tests.Constants
         }
 
         [Test]
+        public void TestPrefixOfAllEqualConstraints()
+        {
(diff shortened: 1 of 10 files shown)


mikkelbu (approved):
The change looks sound as far as I understand the problem from #811 (and the discussions in nunit/nunit#4899), and I thrust you to have a much better understanding of this than I do :)

PR #813, opened 2024-12-22
Optimize Are(Not)Equal when actually wanting to test for Empty.

### LINKED ISSUE #801: Allow `NUnit2005` to recognize `Is.Empty`
We have the following code:

```csharp
Assert.AreEqual(string.Empty, element.GetDomProperty("value"));
Assert.AreEqual("", element.GetDomProperty("value"));
```

The `NUnit2005` fixer transforms these into 

```csharp
Assert.That(element.GetDomProperty("value"), Is.EqualTo(string.Empty));
Assert.That(element.GetDomProperty("value"), Is.EqualTo(""));
```

Both of these could become `Is.Empty` instead

manfred-brands:
Fixes #801


mikkelbu (approved):
Looks good to me. Thanks for the contribution @manfred-brands 👍

PR #816, opened 2024-12-23
Add support for Assert.EnterMultipleScope

### LINKED ISSUE #769: Recognized the new `Assert.EnterMultipleScope()` concept.
The Analyzer for some rules recognizes that it is in an `Assert.Multiple` scope to warn for `null` references.
That rule must be updated to detect the `Assert.EnterMultipleScope()` methods and its disposal.

manfred-brands:
Fixes #769 

Detects that code inside a `using (Assert.EnterMultipleScope())` is inside a multiple context.
Adds a codefix for to use `Assert.EnterMultipleScope`.

### DIFF
--- a/src/nunit.analyzers.tests/UseAssertMultiple/UseAssertMultipleAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/UseAssertMultiple/UseAssertMultipleAnalyzerTests.cs
@@ -45,6 +45,38 @@ namespace NUnit.Analyzers.Tests.UseAssertMultiple
         }");
             RoslynAssert.Valid(this.analyzer, testCode);
         }
+
+#if WOULD_SOMEONE_ACTUALLY_USE_THIS
+        [Test]
+        public void AnalyzeWhenMultipleScopeDeclarationIsUsed()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void Test()
+        {
+            using IDisposable disposable = Assert.EnterMultipleScope();
+
+            Assert.That(true, Is.True);
+            disposable.Dispose();
+            Assert.That(false, Is.False);
+        }");
+            RoslynAssert.Valid(this.analyzer, testCode);
+        }
+#endif
+
+        [Test]
+        public void AnalyzeWhenMultipleScopeStatementIsUsed()
+        {
+            var testCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings(@"
+        public void Test()
+        {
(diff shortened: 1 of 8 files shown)


manfred-brands:
When referencing NUnit >= 4.2.0

{11E56E81-99D5-405D-A9BB-F312078DFAE7}

And 
{BE7F9BE2-6B67-4F8C-A87A-F909AF1127E2}

When referencing NUnit 3.14.0:
{6E962561-C638-4082-8F16-DCD0B05B85AA}

src/nunit.analyzers/UseAssertMultiple/UseAssertMultipleCodeFix.cs
```diff
@@ -125,32 +144,37 @@ public override async Task RegisterCodeFixesAsync(CodeFixContext context)
                     SyntaxFactory.ArgumentList(
                         SyntaxFactory.SeparatedList(new[]
                         {
-                            SyntaxFactory.Argument(parenthesizedLambdaExpression)
+                        SyntaxFactory.Argument(parenthesizedLambdaExpression)
```
mikkelbu, line 147:
```suggestion
                            SyntaxFactory.Argument(parenthesizedLambdaExpression)
```

src/nunit.analyzers.tests/UseAssertMultiple/UseAssertMultipleAnalyzerTests.cs
```diff
@@ -45,6 +45,38 @@ await Assert.MultipleAsync(async () =>
         }");
             RoslynAssert.Valid(this.analyzer, testCode);
         }
+
+#if WOULD_SOMEONE_ACTUALLY_USE_THIS
```
mikkelbu, line 49: Is there a reason for why we wouldn't want to run this test :) ?

src/nunit.analyzers/UseAssertMultiple/UseAssertMultipleCodeFix.cs
manfred-brands: Done.

src/nunit.analyzers.tests/UseAssertMultiple/UseAssertMultipleAnalyzerTests.cs
manfred-brands: Yes, it won't work as we don't have support for it.
It would require tracking the declared variable and check if it is manually disposed somewhere within the scope.

Unless someone has a good reason for using this pattern, I don't see any, I won't spend my time on it.

PR #818, opened 2024-12-27
chore: Replace "buildstats.info" with "img.shields.io"

mikkelbu:
It seems like buildstats.info is no longer working.


stevenaw (commented):
I'm on vacation with questionable wifi so it's tough to review this well on my phone. It looks like the myget badge may be showing a badge for the latest stable version (4.5.0) while it links to a pre release 4.6.x version. Is that expected?

mikkelbu:
> I'm on vacation with questionable wifi so it's tough to review this well on my phone. It looks like the myget badge may be showing a badge for the latest stable version (4.5.0) while it links to a pre release 4.6.x version. Is that expected?

No - but it has always been like this :D (I think I copied the badges originally from one of the other repos). I'll update the badge to use `vpre` instead of `v` in the url, then it should show the prerelease.

stevenaw (approved):
Aha! Changing that has fixed it. LGTM, thanks for noticing and fixing the badge issue @mikkelbu

PR #831, opened 2025-01-13
Split off CodeFixes from analyzer

### LINKED ISSUE #677: Build using .NET8.0 SDK
This issue is extracted from two comments in #609 

> At the same time I wanted to move to .NET8.0 SDK to match NUnit.
> However, the corresponding updated Roslyn Analyzer now insists on separating the CodeFix from the Analyzers.
> CodeFixes should only be loaded by IDE, not by the build process.
https://github.com/nunit/nunit.analyzers/issues/609#issuecomment-1815588823

and
> In addition to that, unit tests fail due some other changes they made: GuOrg/Gu.Roslyn.Asserts#320
https://github.com/nunit/nunit.analyzers/issues/609#issuecomment-1817820626

manfred-brands:
.NET8.0 says there should be no Workspace dependency in an analyzer. CodeFixes should only be loaded by an editor.

Fixes #677


manfred-brands:
@mikkelbu This splits the analyzer into two projects as per guidelines.
To prevent merge conflicts, I don't really want to start on tickets that create new analyzers until this is in main.
I'm on away on holiday afk for the next two weeks.

mikkelbu (approved):
Great work @manfred-brands. I checked the changes and the renamings and all look good. I also tested it out in VS and it works. So I'll be merging this now

Bartleby2718:
@manfred-brands Out of curiosity, where does it say so? The only thing I could find was this sample: https://github.com/dotnet/samples/tree/main/csharp/roslyn-sdk/Tutorials/MakeConst.

manfred-brands:
@Bartleby2718 It isn't as much that I found a description of it, but it was reported by Roslyn analyzers [RS1022: Do not use types from Workspaces assembly in an analyzer](https://github.com/dotnet/roslyn-analyzers/blob/main/docs/rules/RS1022.md) and [RS1038](https://github.com/dotnet/roslyn-analyzers/blob/main/docs/rules/RS1038.md)

The [tutorial](https://learn.microsoft.com/en-us/dotnet/csharp/roslyn-sdk/tutorials/how-to-write-csharp-analyzer-code-fix) uses a template that creates a separate assemblies.

PR #833, opened 2025-01-13
Fix Nunit1029 for generic test method (#832)

### LINKED ISSUE #832: False positive for Nunit1029 when only a type argument is use
The following code
```csharp
using NUnit.Framework;

namespace NUnitTesting
{
    public sealed class GenericTestMethod
    {
        public static IEnumerable<TestCaseData> Source
            => [new TestCaseData { TypeArgs = [typeof(int)] }];

        [TestCaseSource(nameof(Source))]
        public void Generic<T>()
        {

        }
    }
}
```
raises an Nunit1029 error "The TestCaseSource provides '>0' parameter(s), but the Test methods expects '0' parameters", which shouldn't be the case.
Tested with NUnit.Analyzers version 4.6.0.

In my opinion, the check in https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzer.cs#L260 
```csharp
if (methodRequiredParameters + methodOptionalParameters < 1)
```
is valid only for non-generic methods and must be changed to 
```csharp
if (methodRequiredParameters + methodOptionalParameters < 1 && !testMethod.IsGenericMethod)
```
I could provide a PR for this.

cbersch:
Fixes #832

### DIFF
--- a/src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
@@ -659,6 +659,30 @@ namespace NUnit.Analyzers.Tests.TestCaseSourceUsage
             RoslynAssert.Valid(analyzer, testCode);
         }
 
+#if NUNIT4
+        [Test]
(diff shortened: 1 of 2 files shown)


manfred-brands (approved):
Thanks @cbersch , this looks good to me.

PR #845, opened 2025-02-13
feat: Also simplify `Is.EqualTo(0)` to `Is.Zero`

### LINKED ISSUE #824: Use Is.False / Is.True instead of Is.EqualTo
It would be helpful to have analyzers that suggest converting `Is.EqualTo(false)` to `Is.False` and `Is.EqualTo(true)` to `Is.True`. It should probably only flag this when using the `false`/`true` keywords within `Is.EqualTo`, not when those values are coming from variables, since the variables could be a design choice to be reused between setup and assertion.

mikkelbu:
Fixes #824

### DIFF
--- a/src/nunit.analyzers/UseSpecificConstraint/UseSpecificConstraintAnalyzer.cs
+++ b/src/nunit.analyzers/UseSpecificConstraint/UseSpecificConstraintAnalyzer.cs
@@ -45,11 +45,12 @@ namespace NUnit.Analyzers.UseSpecificConstraint
 
                 if (argument is LiteralExpressionSyntax literalExpression)
                 {
-                    constraint = literalExpression.Kind() switch
+                    constraint = (literalExpression.Kind(), literalExpression.Token.ValueText) switch
                     {
-                        SyntaxKind.NullLiteralExpression => NUnitFrameworkConstants.NameOfIsNull,
-                        SyntaxKind.FalseLiteralExpression => NUnitFrameworkConstants.NameOfIsFalse,
-                        SyntaxKind.TrueLiteralExpression => NUnitFrameworkConstants.NameOfIsTrue,
+                        (SyntaxKind.NullLiteralExpression, _) => NUnitFrameworkConstants.NameOfIsNull,
+                        (SyntaxKind.FalseLiteralExpression, _) => NUnitFrameworkConstants.NameOfIsFalse,
+                        (SyntaxKind.TrueLiteralExpression, _) => NUnitFrameworkConstants.NameOfIsTrue,
+                        (SyntaxKind.NumericLiteralExpression, "0") => NUnitFrameworkConstants.NameOfIsZero,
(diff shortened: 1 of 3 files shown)


manfred-brands:
@mikkelbu I didn't do this because `Is.EqualTo(0)` is no longer the same as `Is.Zero`.
The first returns an `EqualNumericConstraint<int>` and the second `EqualConstraint`.

Although I think it isn't that relevant for this case, except it allows `Is.Zero.IgnoreCase`.

On option would be for `Is.Zero` to return `EqualNumericContraint<double>`.

Should we also detected other zero values such as `0.0`, `0.0f`, `0L`, etc.

src/nunit.analyzers/UseSpecificConstraint/UseSpecificConstraintAnalyzer.cs
```diff
@@ -45,11 +45,12 @@ invocationExpression.Expression is MemberAccessExpressionSyntax memberAccessExpr
...
-                    constraint = literalExpression.Kind() switch
+                    constraint = (literalExpression.Kind(), literalExpression.Token.ValueText) switch
                     {
-                        SyntaxKind.NullLiteralExpression => NUnitFrameworkConstants.NameOfIsNull,
-                        SyntaxKind.FalseLiteralExpression => NUnitFrameworkConstants.NameOfIsFalse,
-                        SyntaxKind.TrueLiteralExpression => NUnitFrameworkConstants.NameOfIsTrue,
+                        (SyntaxKind.NullLiteralExpression, _) => NUnitFrameworkConstants.NameOfIsNull,
+                        (SyntaxKind.FalseLiteralExpression, _) => NUnitFrameworkConstants.NameOfIsFalse,
+                        (SyntaxKind.TrueLiteralExpression, _) => NUnitFrameworkConstants.NameOfIsTrue,
+                        (SyntaxKind.NumericLiteralExpression, "0") => NUnitFrameworkConstants.NameOfIsZero,
```
manfred-brands, line 53: Suggest not to switch on both, but treat the numeric part different. Which allow checking some more 0 values.

mikkelbu:
> @mikkelbu I didn't do this because `Is.EqualTo(0)` is no longer the same as `Is.Zero`. The first returns an `EqualNumericConstraint<int>` and the second `EqualConstraint`.

Ahh. I didn't consider this - I thought it was an oversight in the PR

> Although I think it isn't that relevant for this case, except it allows `Is.Zero.IgnoreCase`.
> On option would be for `Is.Zero` to return `EqualNumericContraint<double>`.

I think this makes sense. It could break somebody, but only people that have some odd/wrong constraints (unless I'm overlooking something)

> Should we also detected other zero values such as `0.0`, `0.0f`, `0L`, etc.

Yes - I'll update the PR with your suggestion

manfred-brands:
Superceded by #858

PR #848, opened 2025-02-15
New rule to ensure that base class TestFixtures are abstract.

### LINKED ISSUE #840: Detect incorrect or questionable use of TestFixture inheritance.
When you derive a TestFixture from another TestFixture, NUnit generally handles the situation as follows:

1. It runs the base class tests against an instance of the base class.
2. It runs the base class tests a second time against an instance of the derived class.

Note that (2) is repeated for **every** derived class

This is almost never what the user intended. Normally, what is expected is for the base class tests to only be run once for each derived class, not n + 1 times. That behavior is correctly achieved by using the "Abstract Test Fixture" pattern. If the base class is abstract, then NUnit will not attempt to construct it or run tests against it.

A special case of this problem arises if the TestFixture is parameterized and  no paramterized instance of `[TestFixture()]` In that case, rather than running the contained tests n + 1 times, an error will arise indicating that no valid constructor can be found. 

Both situations (with or without an error arising) should probably be dealt with.

In reviewing issues and questions, I have occasionally found a user who actually __wanted__ the tests to run twice (or n + 1 times). I believe most of the examples used logic in the base class to make it behave differently when executed directly by examining the Type of the instance. This seems pathological to me and I would not recommend supporting it.

manfred-brands:
Fixes #840 

I've run this against the nunit.framework code where it raised 2 errors:

* `NUnit.Framework.Tests.Attributes.ThreadingTests` has 10 derived fixtures.
* `NUnit.Framework.Tests.Attributes.BaseRepeatableTestFixture` has 2 derived fixtures, although this might be intentional.

### DIFF
--- a/src/nunit.analyzers.codefixes/TestFixtureShouldBeAbstract/TestFixtureShouldBeAbstractCodeFix.cs
+++ b/src/nunit.analyzers.codefixes/TestFixtureShouldBeAbstract/TestFixtureShouldBeAbstractCodeFix.cs
@@ -1,0 +1,75 @@
+using System.Collections.Immutable;
+using System.Composition;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using NUnit.Analyzers.Constants;
+
+namespace NUnit.Analyzers.TestFixtureShouldBeAbstract
+{
+    [ExportCodeFixProvider(LanguageNames.CSharp)]
+    [Shared]
+    public class TestFixtureShouldBeAbstractCodeFix : CodeFixProvider
+    {
+        internal const string MakeBaseTestFixtureAbstract = "Make base test fixture abstract";
+
+        public override ImmutableArray<string> FixableDiagnosticIds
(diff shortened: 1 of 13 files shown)


src/nunit.analyzers.codefixes/TestFixtureShouldBeAbstract/TestFixtureShouldBeAbstractCodeFix.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            var modifiers = originalExpression.Modifiers;
+
+            var abstractSyntax = SyntaxFactory.Token(SyntaxKind.AbstractKeyword)
+                                              .WithTrailingTrivia(SyntaxTriviaList.Create(SyntaxFactory.Whitespace(" ")));
+
+            if (!modifiers.Any())
+            {
+                // Get leading trivia from declaration
+                abstractSyntax = abstractSyntax.WithLeadingTrivia(originalExpression.GetLeadingTrivia());
```
mikkelbu, line 69: Can we also add this case to the `TestFixtureShouldBeAbstractCodeFixTests.cs` (or parameterise the existing test even more)

mikkelbu (approved):
Nit-pick - we usually add the documentation files to the solution file otherwise this looks good and I'm happy to merge it when you are

src/nunit.analyzers.codefixes/TestFixtureShouldBeAbstract/TestFixtureShouldBeAbstractCodeFix.cs
manfred-brands: Done.

PR #858, opened 2025-04-05
Rewrite UseSpecificConstraintAnalyzer as an IOperation analyzer

### LINKED ISSUE #856: NUnit4002 shouldn't trigger for unknown types
In my case it is triggered for value with `dynamic` type, but I suspect object also affected

```cs
// Is.Default will not be able to infer default value properly
Assert.That(someDynamic, Is.EqualTo(default(byte)));

// same for object
Assert.That(someObject, Is.EqualTo(default(byte)));
```

manfred-brands:
Fixes #856

This allows us to check the type of TActual and not suggest Is.Default if the type is object.

@mikkelbu I also incorporated the Is.Zero changes you started in #845


mikkelbu (approved):
Looks great and I've no comments except for "thanks for the work" :). I'll merge it now

PR #863, opened 2025-04-25
fix: Don't suggest `Nunit2045` on NUnit versions below version 3.7

### LINKED ISSUE #862: NUnit.Analyzers doesn't recognize the version of NUnit in use
I came across this issue while trying to resolve [Nunit2045](https://github.com/nunit/nunit.analyzers/blob/master/documentation/NUnit2045.md) suggestions in my test code. That is, NUnit.Analyzers was suggesting that I wrap multiple asserts in an `Assert.Multiple()` call. However, when I tried to apply this suggestion, I got compiler errors that `Multiple()` was undefined. Turns out, I was using NUnit 3.5, but `Assert.Multiple()` [was not added until](https://docs.nunit.org/articles/nunit/release-notes/framework.html#nunit-37---may-29-2017) NUnit 3.7. This is a Unity project, so I am constrained to NUnit 3.5 as that is what the [Unity Test Framework](https://docs.unity3d.com/Packages/com.unity.test-framework@1.1/manual/index.html) contains. While I can easily suppress `Nunit2045` across this project, it seems strange that I have to.

**TL;DR** Shouldn't NUnit.Analyzers be aware of the NUnit version in use, and not recommend APIs that don't exist in a particular version?

mikkelbu:
Fixes #862

### DIFF
--- a/src/nunit.analyzers/UseAssertMultiple/UseAssertMultipleAnalyzer.cs
+++ b/src/nunit.analyzers/UseAssertMultiple/UseAssertMultipleAnalyzer.cs
@@ -14,6 +14,7 @@ namespace NUnit.Analyzers.UseAssertMultiple
     public class UseAssertMultipleAnalyzer : BaseAssertionAnalyzer
     {
         private static readonly Version firstNUnitVersionWithEnterMultipleScope = new(4, 2);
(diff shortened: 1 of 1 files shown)


OsirisTerje (approved):
lgtm

manfred-brands (approved):
Looks good to me too

PR #871, opened 2025-05-28
UseSpecificConstraintAnalyzer: Only try convertible.ToDouble on primitive types.

### LINKED ISSUE #870: Compilation error caused by exception in an analyzer after updating to v4.8.0
After updating from v4.7.0 to v4.8.0, we've started getting compilation errors in our solution:

```
CSC: Error AD0001 : Analyzer 'NUnit.Analyzers.UseSpecificConstraint.UseSpecificConstraintAnalyzer' threw an exception of type 'System.FormatException' with message 'Input string was not in a correct format.'.
```

There isn't much to go on as the error doesn't link to a line of code, but it appears to be coming from usages of `Is.EqualTo("...")`. Having turned on detailed logging during the build, this is the error that is happening:

```
System.FormatException: Input string was not in a correct format.
at System.Number.ParseDouble(String value, NumberStyles options, NumberFormatInfo numfmt)
at System.String.System.IConvertible.ToDouble(IFormatProvider provider)
at NUnit.Analyzers.UseSpecificConstraint.UseSpecificConstraintAnalyzer.AnalyzeAssertInvocation(Version nunitVersion, OperationAnalysisContext context, IInvocationOperation assertOperation)
at NUnit.Analyzers.BaseAssertionAnalyzer.AnalyzeInvocation(Version nunitVersion, OperationAnalysisContext context)
at NUnit.Analyzers.BaseAssertionAnalyzer.<>c__DisplayClass5_0.<AnalyzeCompilationStart>b__1(OperationAnalysisContext context)
at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.<>c.<ExecuteOperationAction>b__53_0(ValueTuple`2 data)
at Microsoft.CodeAnalysis.Diagnostics.AnalyzerExecutor.ExecuteAndCatchIfThrows_NoLock[TArg](DiagnosticAnalyzer analyzer, Action`1 analyze, TArg argument, Nullable`1 info, CancellationToken cancellationToken)
```

This lines up with [this line of code](https://github.com/nunit/nunit.analyzers/blob/master/src/nunit.analyzers/UseSpecificConstraint/UseSpecificConstraintAnalyzer.cs#L79) in `UseSpecificConstraintAnalyzer`.

It appears to be that using `Assert.That("some string", Is.EqualTo("some other string"))` is causing this error by incorrectly attempting to convert the `string` to a `double`.

Is there a different assertion format that should be used to check the values of strings now, or is this a bug in the analyzer?

manfred-brands:
Fixes #870

### DIFF
--- a/src/nunit.analyzers.tests/UseSpecificConstraint/UseSpecificConstraintAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/UseSpecificConstraint/UseSpecificConstraintAnalyzerTests.cs
@@ -6,7 +6,7 @@ using NUnit.Framework;
 
 namespace NUnit.Analyzers.Tests.UseSpecificConstraint
 {
-    public class UseSpecificConstraintAnalyzerTests
+    public sealed class UseSpecificConstraintAnalyzerTests
     {
         private static readonly DiagnosticAnalyzer analyzer = new UseSpecificConstraintAnalyzer();
         private static readonly ExpectedDiagnostic expectedDiagnostic =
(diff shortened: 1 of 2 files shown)


manfred-brands:
@mikkelbu I have based to PR off the 4.8.0 tag. Not sure how you release it as a 4.8.1 version.

mikkelbu (approved):
Looks good to me 👍.

I'll just release this via `master` as there are no functional changes on master since 4.8 (and the `tag` should overwrite the version number).

PR #878, opened 2025-06-05
feat: Add analyzer and codefix for `Assert.Multiple` => `Assert.EnterMultipleScope`

### LINKED ISSUE #866: When Assert.Multiple is found, should suggest to convert to Assert.EnterMultipleScope
A new rule should be made that helps the user to convert from Assert.Multiple to Assert.EnterMultipleScope.

Also see issue #865

mikkelbu:
And similarly for `Assert.MultipleAsync` => `Assert.EnterMultipleScope`.

Fixes #866

### DIFF
--- a/src/nunit.analyzers.tests/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeCodeFixTests.cs
@@ -100,7 +100,7 @@ namespace NUnit.Analyzers.Tests.Targets.UseAssertEnterMultipleScope
             {
                 var i = 4;
                 var j = 67;
-                await System.Threading.Tasks.Task.Yield();
+                await Task.Yield();
                 Assert.That(i, Is.EqualTo(j));
             });
         }
@@ -122,7 +122,7 @@ namespace NUnit.Analyzers.Tests.Targets.UseAssertEnterMultipleScope
             {
                 var i = 4;
                 var j = 67;
-                await System.Threading.Tasks.Task.Yield();
+                await Task.Yield();
                 Assert.That(i, Is.EqualTo(j));
             }
         }
@@ -151,7 +151,7 @@ namespace NUnit.Analyzers.Tests.Targets.UseAssertEnterMultipleScope
             {
                 var i = 4;
                 var j = 67;
-                await System.Threading.Tasks.Task.Yield();
+                await Task.Yield();
                 Assert.That(i, Is.EqualTo(j));
             });
         }
@@ -173,7 +173,7 @@ namespace NUnit.Analyzers.Tests.Targets.UseAssertEnterMultipleScope
             {
                 var i = 4;
                 var j = 67;
-                await System.Threading.Tasks.Task.Yield();
+                await Task.Yield();
                 Assert.That(i, Is.EqualTo(j));
             }
         }


src/nunit.analyzers.codefixes/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeCodeFix.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+
+            var blockSyntax = FindBlockSyntax(node);
+            if (blockSyntax is null)
+            {
+                return;
+            }
+
+            var newSyntax = CreateUsingStatementSyntax(blockSyntax, expressionStatementSyntax.GetTrailingTrivia());
+
```
manfred-brands, line 53: If the node is an InvocationExpression of `Assert.Multiple` and  `ParenthesizedLambdaExpressionSyntax` has an `async` modifier and the return value of the method this statement is in is not `async Task` then you have to:
* Create a new MethodDeclarationSyntax replacing the return type with `async Task`
* Copy all statements, except replacing the original `expressionStatementSyntax` with `newSyntax`. See `UseAssertMultipleCodeFix` which also needs to copy statements.

You might think you can do two calls to `ReplaceNode` on for the MethodDeclaration and one for the statement, but I don't think that is allowed from what I remember.

You also need to check whether `System.Threading` is in the `using` list or use a full type name for  the return type.

Have fun.

manfred-brands (changes requested):
What is there is great, you have forgotten one use case where users passed an `async` method to `Assert.Multiple`.
The use case is treated as non-async code, so the issue is flagged, but the code fix generated invalid code as the return type of the test method needs to be `async Task`.

src/nunit.analyzers/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeAnalyzer.cs
```diff
@@ -0,0 +1,43 @@
...
+            title: UseAssertEnterMultipleScopeConstants.Title,
+            messageFormat: UseAssertEnterMultipleScopeConstants.Message,
+            category: Categories.Assertion,
+            defaultSeverity: DiagnosticSeverity.Info,
+            description: UseAssertEnterMultipleScopeConstants.Description);
+
+        public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(descriptor);
+
+        protected override bool IsAssert(bool hasClassicAssert, IInvocationOperation invocationOperation) =>
+            base.IsAssert(false, invocationOperation);
```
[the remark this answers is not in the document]
mikkelbu: > Is this worth it?

You are right that this "optimization" probably adds more cognitive overhead than it improves the flow. So I think I'll just remove the override

src/nunit.analyzers.codefixes/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeCodeFix.cs
mikkelbu: Thanks for the review Manfred. I opted for the solution with two `ReplaceNode` calls as I found this to easier (although I needed to find the new node in the updated document).

src/nunit.analyzers.codefixes/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeCodeFix.cs
```diff
@@ -0,0 +1,159 @@
...
+            NameSyntax name = SyntaxFactory.IdentifierName(parts[0]);
+            for (int i = 1; i < parts.Length; i++)
+            {
+                name = SyntaxFactory.QualifiedName(name, SyntaxFactory.IdentifierName(parts[i]));
+            }
+
+            return name;
+        }
+
+        private static T? FindNearestParentOfType<T>(SyntaxNode node)
```
manfred-brands, line 147: This method already has something by default: `node.FirstAncestorOrSelf<T>`

src/nunit.analyzers.codefixes/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeCodeFix.cs
```diff
@@ -0,0 +1,159 @@
...
+            if (root is null)
+            {
+                return;
+            }
+
+            context.CancellationToken.ThrowIfCancellationRequested();
+
+            var node = root.FindNode(context.Span);
+
+            var expressionStatementSyntax = FindNearestParentOfType<ExpressionStatementSyntax>(node);
```
manfred-brands, line 44: There is a standard method that does the same:
```suggestion
            var expressionStatementSyntax = node.FirstAncestorOrSelf<ExpressionStatementSyntax>();
(node);
```

src/nunit.analyzers.codefixes/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeCodeFix.cs
```diff
@@ -0,0 +1,159 @@
...
+                            SyntaxFactory.IdentifierName(NUnitFrameworkConstants.NameOfAssert),
+                            SyntaxFactory.IdentifierName(NUnitV4FrameworkConstants.NameOfEnterMultipleScope))))
+                .WithTrailingTrivia(trailingTrivia);
+
+        private static SyntaxNode UpdateTestMethodSignatureIfNecessary(
+            SyntaxNode newRoot,
+            SyntaxNode newSyntaxInTree,
+            bool lambdaHasAsyncKeyword)
+        {
+            var methodDeclaration = FindNearestParentOfType<MethodDeclarationSyntax>(newSyntaxInTree);
```
manfred-brands, line 102: Use existing method:
```suggestion
            var methodDeclaration = newSyntaxInTree.FirstAncestorOrSelf<MethodDeclarationSyntax>();
(newSyntaxInTree);
```

src/nunit.analyzers.codefixes/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeCodeFix.cs
```diff
@@ -0,0 +1,159 @@
...
+
+            if (!lambdaHasAsyncKeyword || methodDeclaration is null || IsAsyncTaskMethod(methodDeclaration))
+            {
+                return newRoot;
+            }
+
+            var systemThreadingTasksUsingExists = newRoot.DescendantNodes()
```
manfred-brands, line 109: `DescendantNodes` goes recursively through all children when we know that UsingDirectives are not inside TypeDeclarations.
One option to reduce iterations is `newRoot.DescendantNodes(s => s is not BaseTypeDeclarationSyntax)`

This doesn't cope with silly developers that have two namespace declarations with usings in one file where we only should look at the namespace containing our method.

Another option is to go up from the method:
```csharp
var namespaceDeclaration = methodDeclaration.FirstAncestorOrSelf<NamespaceDeclarationSyntax>();
var compilationUnit = methodDeclaration.FirstAncestorOrSelf<CompilationUnitSyntax>();

var systemThreadingTasksUsingExists =
    compilationUnit?.Usings.Any(IsUsingSystemThreadingTasks) is true ||
    namespaceDeclaration?.Usings.Any(IsUsingSystemThreadingTasks) is true;
```

Where IsUsingSystemThreadingTasks is:
```charp
        private static bool IsUsingSystemThreadingTasks(UsingDirectiveSyntax u)
        {
            return u.Name.ToString() == SystemThreadingTasksNamespace;
        }
```

src/nunit.analyzers.tests/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeCodeFixTests.cs
```diff
@@ -0,0 +1,165 @@
...
+            {
+                var i = 4;
+                var j = 67;
+                await Task.Yield();
+                Assert.That(i, Is.EqualTo(j));
+            }
+        }");
+            RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode);
+        }
+
```
manfred-brands, line 83: I added one test where the using statement is inside the namespace declaration:
```csharp
        [Test]
        public void VerifyAssertMultipleWithAsyncDelegateAndUsingSystemThreadingInsideNamespace()
        {
            var code = @"
using NUnit.Framework;

namespace NUnit.Analyzers.Tests.Targets.UseAssertEnterMultipleScope
{
    using System.Threading.Tasks;

    public class TestClass
    {
        [Test]
        public void TestMulti()
        {
            ↓Assert.Multiple(async () =>
            {
                var i = 4;
                var j = 67;
                await System.Threading.Tasks.Task.Yield();
                Assert.That(i, Is.EqualTo(j));
            });
        }
    }
}";
            var fixedCode = @"
using NUnit.Framework;

namespace NUnit.Analyzers.Tests.Targets.UseAssertEnterMultipleScope
{
    using System.Threading.Tasks;

    public class TestClass
    {
        [Test]
        public async Task TestMulti()
        {
            {
                var i = 4;
                var j = 67;
                await System.Threading.Tasks.Task.Yield();
                Assert.That(i, Is.EqualTo(j));
            }
        }
    }
}";

            RoslynAssert.CodeFix(analyzer, fix, expectedDiagnostic, code, fixedCode);
        }
```

src/nunit.analyzers.tests/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeCodeFixTests.cs
```diff
@@ -81,6 +81,108 @@ public async Task TestMulti()
...
+    public class TestClass
+    {
+        [Test]
+        public async Task TestMulti()
+        {
+            using (Assert.EnterMultipleScope())
+            {
+                var i = 4;
+                var j = 67;
+                await System.Threading.Tasks.Task.Yield();
```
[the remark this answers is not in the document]
mikkelbu: The worst part is that I actually noticed this and thought it was on purpose. I'll remove it where it is not necessary (I also copied it)

PR #881, opened 2025-06-08
fix: NUnit4002 shouldn't suggest `Is.Default` when expected/actual types don't match

### LINKED ISSUE #879: NUnit4002 shuoldn't trigger for `T` vs `nullable<T>` struct types
```cs
class SomeType
{
    public Guid? Id { get; set; }
}

// proper assert : id != guid.empty
Assert.That(obj.Id, Is.Not.EqualTo(default(Guid)));

// suggested incorrect assert : Id != null
Assert.That(obj.Id, Is.Not.Default);
```

mikkelbu:
Fixes #879


manfred-brands (approved):
Thanks @mikkelbu, looks good.

PR #882, opened 2025-06-08
feat: Also consider `typeof(T)` as a constant

### LINKED ISSUE #880: NUnit2007 could flag typeof() as a constant first parameter
Consider the following code. `0` is a compile-time constant and `typeof(TestAnalyzer)` is a runtime constant, but passing a runtime constant like `typeof()` as the `actual` is still likely an indicator of misuse. The analyzers don't currently identify cases like this but it might be helpful to extend them to catch it.

```
[Test]
public void Test1()
{
    var result = new TestAnalyzer();

    Assert.That(typeof(TestAnalyzer), Is.EqualTo(result.GetType()));
    Assert.That(0, Is.EqualTo(result.RowVersion));
}

public class TestAnalyzer
{
    public Guid Id { get; set; } = Guid.Empty;
    public int RowVersion { get; set; } = 0;
}
```

mikkelbu:
Fixes #880


stevenaw (approved):
LGTM as well, thanks for the prompt update @mikkelbu

PR #898, opened 2025-06-17
chore: Improve tests by inserting missing `↓` to have more precise assertions

### LINKED ISSUE #894: Improve tests by adding ↓ to tests were it is missing in the source
During a review I noticed that in e.g. `AnalyzeWhenTypedCustomComparerProvided` we don't have a `↓`.

https://github.com/nunit/nunit.analyzers/blob/8bce842501d1aef1d951a431c5b14db6e69fca0f/src/nunit.analyzers.tests/SomeItemsIncompatibleTypes/SomeItemsIncompatibleTypesAnalyzerTests.cs#L143-L153

This is not a major problem - and I guess we only have a few tests like this, but to make the tests more precise we should check that all tests with `RoslynAssert.Diagnostics`, `RoslynAssert.CodeFix` or `RoslynAssert.Suppressed` contains a `↓` in the expected location.

EDIT: Full list of RoslynAssert... that we currently use
```
RoslynAssert.CodeFix
RoslynAssert.Diagnostics
RoslynAssert.FixAll
RoslynAssert.NoAnalyzerDiagnostics
RoslynAssert.NoFix
RoslynAssert.NotSuppressed
RoslynAssert.Suppressed
RoslynAssert.Valid
```

mikkelbu:
Fixes #894

### DIFF
--- a/src/nunit.analyzers.tests/ClassicModelAssertUsage/GreaterClassicModelAssertUsageCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/ClassicModelAssertUsage/GreaterClassicModelAssertUsageCodeFixTests.cs
@@ -108,7 +108,7 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         public void CodeFixPreservesLineBreakBeforeMessage()
         {
             var code = TestUtility.WrapInTestMethod(@"
-            ClassicAssert.Greater(2d, 3d,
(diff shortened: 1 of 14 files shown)


manfred-brands (approved):
Thanks @mikkelbu

Looks sensible. If the were in the wrong spot, the build would have failed.
Is there any way we can fail the build if there is no ↓ in the source?

PR #900, opened 2025-07-26
NUnit2056: Also keep leading trivia

### LINKED ISSUE #899: NUnit2056 analyzer's code fix removes comments and empty lines above it.
When applying the "Use Assert.EnterMultipleScope method" code fix for "NUnit2056: Consider using Assert.EnterMultipleScope statement instead of Assert.Multiple/Assert.MultipleAsync" it's erasing comments and empty lines above it.

NUnit.Analyzers version: 4.9.2

### Example:
```csharp
[Test]
public void Test1()
{
    // Arrange
    const int expected = 4;
    
    // Act
    int? actual = 2 + 2;
    
    // Assert
    Assert.Multiple(() => {
        Assert.That(actual, Is.Not.Null);
        Assert.That(actual, Is.EqualTo(expected));
    });
}
```

#### Actual result:
```csharp
[Test]
public void Test1()
{
    // Arrange
    const int expected = 4;
    
    // Act
    int? actual = 2 + 2;
    {
        Assert.That(actual, Is.Not.Null);
        Assert.That(actual, Is.EqualTo(expected));
    }
}
```

#### Expected result:
```csharp
[Test]
public void Test1()
{
    // Arrange
    const int expected = 4;
    
    // Act
    int? actual = 2 + 2;
    
    // Assert
    {
        Assert.That(actual, Is.Not.Null);
        Assert.That(actual, Is.EqualTo(expected));
    }
}
```

manfred-brands:
Fixes #899

### DIFF
--- a/src/nunit.analyzers.codefixes/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeCodeFix.cs
+++ b/src/nunit.analyzers.codefixes/UseAssertEnterMultipleScope/UseAssertEnterMultipleScopeCodeFix.cs
@@ -53,7 +53,8 @@ namespace NUnit.Analyzers.UseAssertEnterMultipleScope
                 return;
             }
 
-            var newBodySyntax = CreateUsingStatementSyntax(blockSyntax, expressionStatementSyntax.GetTrailingTrivia());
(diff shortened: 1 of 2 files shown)


mikkelbu (approved):
LGTM - thanks for the fix @manfred-brands

Issue #885, opened 2025-06-11
Bump to NUnit version 4.4 when this is released


mikkelbu:
See comments in PR https://github.com/nunit/nunit.analyzers/pull/883#discussion_r2136516399_

PR #905, opened 2025-08-09
NUnit4.4 alpha -> beta changes

manfred-brands:
Related to #883, #885

To ensure binary compatibility, NUnit 4.4 changed the return of `Is.EqualTo` from a generic  back to a non-generic one.


mikkelbu (approved):
LGTM

PR #913, opened 2025-09-10
Add limited support for `using var scope = Assert.EnterMultipleScope()`

### LINKED ISSUE #910: using declarations not recognized by NUnit2045
Repro:
[NUnitTest.zip](https://github.com/user-attachments/files/21774129/NUnitTest.zip)

Image

Works when this is a classic using statement

manfred-brands:
Fixes #910 

It does not detect users calling `Dispose()` on the declared variable.

I'm a bit reluctant to add this, but we have had 2 independent reports about users using this declaration.


mikkelbu (approved):
LGTM - sorry that it has taken so long for me to have time to review this. 

EDIT: Regarding your concern about adding this, then I think you are correct to add it (and then we will have to see which feedback we receives)

PR #924, opened 2025-10-05
An array TestCase parameter can be assigned to an IEnumerable<T>

### LINKED ISSUE #922: NUnit1001 - false positive for arguments with generic parameters
Test values for arguments with generic parameters are incorrectly reported as unassignable, even when they actually match, e.g.:
```cs
[Test]
[TestCase(arg: new[] { 1, 2, 3, })]
[TestCase(arg: new[] { "a", "b", "c", })]
[TestCase(arg: new[] { typeof(int), typeof(string), typeof(double), })]
[TestCase(arg: new[] { 1ul, 2ul, 3ul, })]
public void TestCollectionMethod<T>(IEnumerable<T> collection)
{
	// ... test
}
```

manfred-brands:
Fixes #922


mikkelbu (approved):
Thanks for the fix @manfred-brands. I'll merge it now

PR #942, opened 2025-10-30
Detect 'Is.NaN' with unsupported types.

### LINKED ISSUE #929: Alert when using `Is.NaN` or `Is.Not.NaN` on an incompatible type
I noticed recently that we don't alert to misuse of the `NaNConstraint`. The constrain is only coded to work for `double` or `float` (though I think "Single" will also work as `{single} is float` will evaluate to true).

Other types, such as guids, ints, etc, will always fail (or always pass in the case of `Is.Not.NaN`). Either case indicates a misuse that could be alerted about. Analyzers have previously triggered an "error" for these cases where it would never pass. Would a "warning" be appropriate for cases which always pass?

manfred-brands:
Fixes #929

### DIFF
--- a/src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
@@ -1123,7 +1123,7 @@ using System.Collections.Generic;");
         {
             var testCode = TestUtility.WrapInTestMethod(@$"
                 var actual = {actual};
-                Assert.That(actual, Is.Not.NaN);
+                Assert.That(actual, ↓Is.Not.NaN);
             ");
 
             RoslynAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
@@ -1136,7 +1136,7 @@ using System.Collections.Generic;");
         {
             var testCode = TestUtility.WrapInTestMethod(@$"
                 var actual = {actual};
-                Assert.That(actual, Is.Positive.And.↓Not.NaN);
+                Assert.That(actual, Is.Positive.Or.↓NaN);
             ");
 
             RoslynAssert.Diagnostics(analyzer, expectedDiagnostic, testCode);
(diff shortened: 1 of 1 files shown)


src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
```diff
@@ -1091,5 +1091,66 @@ public AnotherRecord(string name)
...
+        }
+
+        [TestCase("3.3m")]
+        [TestCase("4")]
+        [TestCase("string.Empty")]
+        public void AnalyzeIsNotNaNOnUnsupportedTypes(string actual)
+        {
+            var testCode = TestUtility.WrapInTestMethod(@$"
+                var actual = {actual};
+                Assert.That(actual, Is.Not.NaN);
```
mikkelbu:
```suggestion
                Assert.That(actual, Is.↓Not.NaN);
```

src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
```diff
@@ -1091,5 +1091,66 @@ public AnotherRecord(string name)
...
+        }
+
+        [TestCase("3.3m")]
+        [TestCase("4")]
+        [TestCase("string.Empty")]
+        public void AnalyzeIsNaNOnUnsupportedTypes(string actual)
+        {
+            var testCode = TestUtility.WrapInTestMethod(@$"
+                var actual = {actual};
+                Assert.That(actual, Is.Positive.And.↓Not.NaN);
```
mikkelbu: The test name `AnalyzeIsNaNOnUnsupportedTypes` says `IsNan` - so should we remove the `Not` from the assert

mikkelbu (approved):
Nice and simple solution (simpler than I would have expected). I've only added a couple of comments, but I'm happy to merge it when you are

src/nunit.analyzers.tests/EqualToIncompatibleTypes/EqualToIncompatibleTypesAnalyzerTests.cs
manfred-brands: Done.

PR #944, opened 2025-10-31
New analyzer detecting misused 'Is.Not.Null.Or.Empty'

### LINKED ISSUE #189: Warning when Is.Not.Null.Or.Empty used
Operators are a bit confusing in NUnit. Personally I don't fully understand how they work.

I've seen multiple times that constraint `Is.Not.Null.Or.Empty` is used to validate that string is not empty, which is wrong - it should be either `Is.Not.Null.And.Not.Empty`, or `Is.Not.Null & Is.Not.Empty`, or `!Is.Null.Or.Empty`.
But it might be missed because `Not` operator precedence is not obvious (one of the reasons I prefer using `&`/`|` operators instead of `.And`/`.Or`).

Since `Is.Not.Null.Or.Empty` constraint makes no sense (empty string or collection is always not null), I suggest always marking it as warning.

Any ideas for other commonly misused constraints?

manfred-brands:
Fixes #189

### DIFF
--- a/src/nunit.analyzers.codefixes/MisusedConstraints/MisusedConstraintsCodeFix.cs
+++ b/src/nunit.analyzers.codefixes/MisusedConstraints/MisusedConstraintsCodeFix.cs
@@ -40,23 +40,13 @@ namespace NUnit.Analyzers.MisusedConstraints
             if (argument is null)
                 return;
 
-            ExpressionSyntax expression = IdentifierName(NUnitFrameworkConstants.NameOfIs);
-
-            expression = MemberAccessExpression(SyntaxKind.SimpleMemberAccessExpression,
-                expression,
-                IdentifierName(NUnitFrameworkConstants.NameOfIsNot));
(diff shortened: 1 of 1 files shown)


src/nunit.analyzers/MisusedConstraints/MisusedConstraintsAnalyzer.cs
```diff
@@ -0,0 +1,96 @@
...
+                return;
+
+            // Consider the whole constraint expression, not just individual parts,
+            if (IsNotNullOrEmptyOperation(constraintExpression.Operation))
+            {
+                context.ReportDiagnostic(Diagnostic.Create(
+                        descriptor,
+                        constraintExpression.Operation.Syntax.GetLocation(),
+                        $"This is equivalent to {IsNotNull}, did you mean {IsNotNullAndNotEmpty}?"));
+                return;
```
mikkelbu:
```suggestion
```

PR #946, opened 2025-11-01
Fix for UnnecessaryDelegateConstraint

### LINKED ISSUE #945: [NUnit2057] Reported for Throws assert
Version 4.11.1 (no reports in 4.11.0)

From rule description I expect such asserts not rise analyzer warning, but probably something doesn't work as expected.

Some examples where I have it:

```cs
Assert.That(() =>
{
    ...
}, Throws.InstanceOf<ArgumentOutOfRangeException>().With.Property(nameof(ArgumentOutOfRangeException.ParamName)).EqualTo("exponentialBase"));

Assert.That(
				() => db.GetTable<Child>().ToList(),
				Throws.TypeOf<ArgumentException>()
					.Or.TypeOf<InvalidOperationException>()
					.Or.TypeOf<MySqlException>());


Assert.That(() => query.ToList(), Throws.InvalidOperationException.With.Message.EqualTo("Cannot construct object 'Tests.Linq.EagerLoadingTests+EntityA'. Following members are not assignable: ObjectBRO."));
```

As all reports are quite complex, I suspect analyzer ignores only simple Throws expressions.

manfred-brands:
Fixes #945

### DIFF
--- a/src/nunit.analyzers.tests/DelegateUnnecessary/DelegateUnnecessaryAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/DelegateUnnecessary/DelegateUnnecessaryAnalyzerTests.cs
@@ -44,6 +44,19 @@ namespace NUnit.Analyzers.Tests.DelegateUnnecessary
             RoslynAssert.Valid(analyzer, testCode);
         }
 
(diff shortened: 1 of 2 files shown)


mikkelbu (approved):
LGTM

PR #954, opened 2025-11-08
NUnit2005 was to aggressive comparing 'XXX.Empty' with 'Is.Empty'.

### LINKED ISSUE #952: Code fix for `NUnit2005` tries to use `Is.Empty` constraint on incompatible types.
The `Is.Empty` constraint only works on certain types like `string`, `Guid`, `IEnumerable`, ect.
But the code fix for `NUnit2005` will still try to use it if the original code compared against a field called `Empty`, even if the type is not compatible.

Example:
```cs
using NUnit.Framework;
using NUnit.Framework.Legacy;

public enum Range
{
    Empty,
    Full
}

public class Test
{
    [Test]
    public void SampleTest()
    {
        var rangeValue = Range.Empty;
        ClassicAssert.AreEqual(Range.Empty, rangeValue);
    }
}
```
The code fix (e.g., `dotnet format --diagnostics NUnit2005`) will replace the assertion with:
```cs
Assert.That(rangeValue, Is.Empty);
```

This test will then fail with:
```
System.ArgumentException : The actual value must be not-null, a string, Guid, have an int Count property, IEnumerable or DirectoryInfo. The value passed was of type Range.
Parameter name: actual
  at NUnit.Framework.Constraints.EmptyConstraint.ApplyTo[TActual](TActual actual)
  at NUnit.Framework.Assert.That[TActual](TActual actual, IResolveConstraint expression, NUnitString message, String actualExpression, String constraintExpression)
  at Test.SampleTest()
```

Version: 4.11.2

manfred-brands:
Fixes #952

### DIFF
--- a/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
+++ b/src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
@@ -413,7 +413,7 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         [Test]
         public void CodeFixUsesIsEmpty()
         {
-            const string usingSystemCollectionsImmutable = "using System.Collections.Generic;";
+            const string usingSystemCollectionsGeneric = "using System.Collections.Generic;";
 
             var code = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
         [Test]
@@ -427,7 +427,7 @@ namespace NUnit.Analyzers.Tests.ClassicModelAssertUsage
         {{
             public static readonly MyCollection Empty = new MyCollection();
         }}",
-            usingSystemCollectionsImmutable);
+            usingSystemCollectionsGeneric);
 
             var fixedCode = TestUtility.WrapMethodInClassNamespaceAndAddUsings($@"
         [Test]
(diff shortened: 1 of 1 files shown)


src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
```diff
@@ -409,5 +409,89 @@ public void CodeFixUsesIsEmpty(string expected)
...
+        {{
+            public static readonly MyCollection Empty = new MyCollection();
+        }}",
+            usingSystemCollectionsImmutable);
+
+            IEnumerable<MetadataReference> existingReferences = Settings.Default.MetadataReferences ?? Enumerable.Empty<MetadataReference>();
+
+            Settings settings = Settings.Default
+                                        .WithMetadataReferences(existingReferences.Concat(MetadataReferences.Transitive(typeof(ImmutableArray<>))));
+
```
mikkelbu: Is this necessary (and the same comment to `settings: settings` below ?

mikkelbu (approved):
Looks good @manfred-brands. I've only two nitpick comments

src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
manfred-brands: Thanks @mikkelbu. No it isn't. Cut&Paste from the method above and I forgot to clean up.

src/nunit.analyzers.tests/ClassicModelAssertUsage/AreEqualClassicModelAssertUsageCodeFixTests.cs
```diff
@@ -409,5 +409,89 @@ public void CodeFixUsesIsEmpty(string expected)
                 fixTitle: ClassicModelAssertUsageCodeFix.TransformToConstraintModelDescription,
                 settings: settings);
         }
+
+        [Test]
+        public void CodeFixUsesIsEmpty()
+        {
+            const string usingSystemCollectionsImmutable = "using System.Collections.Generic;";
```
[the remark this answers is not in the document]
manfred-brands: Yes indeed, same cut & paste error as above.

Issue #957, opened 2025-11-18
Relax NUnit1029  for TestCaseSource where method accepts a single "params" array


stevenaw:
Params support for TestCaseSource will be added in http://github.com/nunit/nunit/pull/5057
The NUnit1029 rule will need to be relaxed to not alert when a method has _only_ a `params` arg. See https://github.com/nunit/nunit/blob/04461658b1e5c9a16091c2a03d63aba25aba6e95/src/NUnitFramework/tests/Attributes/TestCaseSourceTests.cs#L406 for more

PR #958, opened 2025-11-19
Adjust NUnit1029 to account for TestCaseSource support for params and optional arguments

stevenaw:
Fixes https://github.com/nunit/nunit.analyzers/issues/957

### DIFF
--- a/src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
+++ b/src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
@@ -561,7 +561,7 @@ namespace NUnit.Analyzers.Tests.TestCaseSourceUsage
         [TestCaseSource(nameof(TestData))]
         public void ShortName(params int[] x)
         {
-            Assert.That(x.Length, Is.GreaterThanOrEqualTo(0));
+            Assert.That(x, Has.Length.GreaterThanOrEqualTo(0));
         }
 
         static IEnumerable<int> TestData()


stevenaw:
I think this is ready for review. I ended up avoiding generic method scenarios since it looks like the existing code also option to avoid explicit type inference + validation for those cases

src/nunit.analyzers.tests/TestCaseSourceUsage/TestCaseSourceUsesStringAnalyzerTests.cs
```diff
@@ -551,6 +551,109 @@ public class InnerClass
...
+        }
+
+        static IEnumerable<int> TestData()
+        {
+            for (int i = 1; i <= 3; i++)
+            {
+                yield return i;
+            }
+        }
+    }", additionalUsings: "using System.Collections.Generic;");
```
manfred-brands: NUnit itself fails to run this test with _Unable to determine type arguments for method_.

I always extract the code from the string to see if NUnit will actually run it.

manfred-brands (commented):
I had some small remarks on the use of RoslynAssert.
Your example test code didn't get to the analyzer as there were compile errors.

I pushed the necessary fixes up, but the underlying test code doesn't work when run in NUnit 4.4.0.
I'm trying it now on master and see if it is related to the [bug](https://github.com/nunit/nunit/issues/5061) you raised

manfred-brands:
> I'll quickly remove the remaining places the file uses it to avoid future copy/paste issues.

No. Only the tests that have `RoslynAssert.Valid`.
The ones that expect a diagnostic use the location of the marker to match line numbers.

Please revert your last commit!

mikkelbu:
> Coming back to this after the 4.5 release of the framework. 
> @manfred-brands @mikkelbu Are there any other changes needed here?

Sorry about the slow response. I think I'll have time to look at it tonight.

mikkelbu (approved):
LGTM - only minor comments / questions

EDIT: I've resolved some of the CoPilot and old comments that are no longer relevant

stevenaw:
Thanks for your review @mikkelbu 
I've committed your `Has.Length` suggestion. I'm happy to merge this once the CI build completes

mikkelbu:
Thanks for the contribution @stevenaw - I might a release out tomorrow - or Friday

PR #984, opened 2026-04-11
ActualValueDelegate and Func are always evaluated

manfred-brands:
Fixes #982

We need to consider their return type.
TestDelegate and Action are never evaluated for their return type as they don't have one.

The checks are not taking the NUnit version into account.
We could do that, but I was thinking to pre-warn users of `Func<T>` being evaluated in future.


OsirisTerje:
@mikkelbu Can you review this?  Would be nice to get this in before NUnit 4.6 is out.

mikkelbu:
Thanks for the ping @OsirisTerje - I've been so busy lately that I completely overlooked this PR. I'll see if I've time tonight. Alternatively, I should have time Friday evening.

mikkelbu (approved):
I've not followed the discussion in NUnit about this, so I've just reviewed the changes without much context, but it looks good and I trust the judgement of @manfred-brands, so I'm happy to merge this when @manfred-brands is
(and then I'll make a release)

PR #989, opened 2026-05-28
Add TaskReturnShouldBeUsed analyzer and codefix

### LINKED ISSUE #988: Recognize that Assert.ThrowsAsync etc now return a Task that needs awaiting.
In NUnit 5, the `Async` named method are actually converted to `async` and now return a `Task`

This rule is to ensure that users `await` the task instead of ignoring it meaning the code isn't actually executed.

```csharp
        [Test]
        public void GenericThrowsAsyncSucceedsWithDelegate()
        {
            Assert.ThrowsAsync<ArgumentException>(AsyncTestDelegates.ThrowsArgumentException);
        }
```

The analyzer should detect this and offer a code fix to not only `await` the method but also potentially change the return type to `async Task`:

```csharp
        [Test]
        public async Task GenericThrowsAsyncSucceedsWithDelegate()
        {
            await Assert.ThrowsAsync<ArgumentException>(AsyncTestDelegates.ThrowsArgumentException);
        }
```

manfred-brands:
Fixes #988 

Also add support for NUnit5 where Async method are now truly async: https://github.com/nunit/nunit/issues/4384

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "10.0.100",
+    "version": "8.0.100",
     "rollForward" : "latestFeature",
     "allowPrerelease": false
   }
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.100",
+    "version": "10.0.100",
     "rollForward" : "latestFeature",
     "allowPrerelease": false
   }
--- a/src/nunit.analyzers.codefixes/Helpers/CodeFixHelper.cs
+++ b/src/nunit.analyzers.codefixes/Helpers/CodeFixHelper.cs
@@ -292,12 +292,11 @@ namespace NUnit.Analyzers.Helpers
                                     predefinedType.Keyword.IsKind(SyntaxKind.VoidKeyword);
 
             var task = SyntaxFactory.Identifier(nameof(Task));
-            TypeSyntax updatedReturnType =
-                isVoidReturnType ?
-                SyntaxFactory.IdentifierName(task) :
-                SyntaxFactory.GenericName(task,
-                SyntaxFactory.TypeArgumentList(
-                    SyntaxFactory.SeparatedList(new TypeSyntax[] { methodReturnType })));
+            TypeSyntax updatedReturnType = isVoidReturnType
+                ? SyntaxFactory.IdentifierName(task)
+                : SyntaxFactory.GenericName(task,
+                    SyntaxFactory.TypeArgumentList(
+                        SyntaxFactory.SeparatedList([methodReturnType])));
 
             if (systemThreadingTasksUsingExists)
             {
--- a/src/nunit.analyzers.codefixes/Helpers/CodeFixHelper.cs
+++ b/src/nunit.analyzers.codefixes/Helpers/CodeFixHelper.cs
@@ -1,6 +1,7 @@
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Threading.Tasks;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -260,7 +261,7 @@ namespace NUnit.Analyzers.Helpers
                 compilationUnit?.Usings.Any(IsUsingSystemThreadingTasks) is true ||
                 namespaceDeclaration?.Usings.Any(IsUsingSystemThreadingTasks) is true;
 
-            var taskTypeName = GetTaskTypeSyntax(systemThreadingTasksUsingExists);
+            var taskTypeName = GetTaskTypeSyntax(methodDeclaration.ReturnType, systemThreadingTasksUsingExists);
 
             var newMethodDeclaration = methodDeclaration
                 .WithModifiers(methodDeclaration.Modifiers.Add(SyntaxFactory.Token(SyntaxKind.AsyncKeyword)))
@@ -270,25 +271,35 @@ namespace NUnit.Analyzers.Helpers
         }
 
         internal const string SystemThreadingTasksNamespace = "System.Threading.Tasks";
-        internal const string TaskTypeName = "Task";
 
         private static bool IsUsingSystemThreadingTasks(UsingDirectiveSyntax u) =>
             u.Name.ToString() == SystemThreadingTasksNamespace;
 
-        private static TypeSyntax GetTaskTypeSyntax(bool systemThreadingTasksUsingExists)
-            => systemThreadingTasksUsingExists
-                ? SyntaxFactory.ParseTypeName(TaskTypeName)
-                : QualifiedNameFromParts("System", "Threading", "Tasks", TaskTypeName);
-
-        private static NameSyntax QualifiedNameFromParts(params string[] parts)
+        private static TypeSyntax GetTaskTypeSyntax(TypeSyntax methodReturnType, bool systemThreadingTasksUsingExists)
         {
-            NameSyntax name = SyntaxFactory.IdentifierName(parts[0]);
-            for (int i = 1; i < parts.Length; i++)
+            bool isVoidReturnType = methodReturnType is PredefinedTypeSyntax predefinedType &&
+                                    predefinedType.Keyword.IsKind(SyntaxKind.VoidKeyword);
+
+            var task = SyntaxFactory.Identifier(nameof(Task));
+            TypeSyntax updatedReturnType =
+                isVoidReturnType ?
+                SyntaxFactory.IdentifierName(task) :
+                SyntaxFactory.GenericName(task,
+                SyntaxFactory.TypeArgumentList(
+                    SyntaxFactory.SeparatedList(new TypeSyntax[] { methodReturnType })));
+
+            if (systemThreadingTasksUsingExists)
             {
-                name = SyntaxFactory.QualifiedName(name, SyntaxFactory.IdentifierName(parts[i]));
+                return updatedReturnType;
             }
 
-            return name;
+            NameSyntax tasks = SyntaxFactory.QualifiedName(
+                    SyntaxFactory.QualifiedName(
+                        SyntaxFactory.IdentifierName("System"),
+                        SyntaxFactory.IdentifierName("Threading")),
+                    SyntaxFactory.IdentifierName("Tasks"));
+
+            return SyntaxFactory.QualifiedName(tasks, (SimpleNameSyntax)updatedReturnType);
         }
     }
 }
--- a/src/nunit.analyzers.codefixes/Helpers/CodeFixHelper.cs
+++ b/src/nunit.analyzers.codefixes/Helpers/CodeFixHelper.cs
@@ -277,6 +277,17 @@ namespace NUnit.Analyzers.Helpers
 
         private static TypeSyntax GetTaskTypeSyntax(TypeSyntax methodReturnType, bool systemThreadingTasksUsingExists)
         {
+            bool isTaskReturnType = methodReturnType is IdentifierNameSyntax identifierName &&
+                                    identifierName.Identifier.Text == nameof(Task);
+            bool isTaskOfTReturnType = methodReturnType is GenericNameSyntax genericName &&
+                                        genericName.Identifier.Text == nameof(Task) &&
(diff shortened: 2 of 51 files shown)


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "8.0.100",
+    "version": "10.0.100",
     "rollForward" : "latestFeature",
     "allowPrerelease": false
   }
```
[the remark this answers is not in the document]
manfred-brands: 10.0.100 is not pre-release and has been out for half a year.
As rollForward is set any .NET 10 SDK is ok

mikkelbu:
Thanks for the PR @manfred-brands - I'll try to take a look at it tonight (I already looked through some of the smaller yesterday, but I think I'm only halfway)

src/nunit.analyzers/Constants/NUnitPreV5FrameworkConstants.cs
```diff
@@ -0,0 +1,16 @@
+namespace NUnit.Analyzers.Constants
+{
+    /// <summary>
+    /// String constants for relevant NUnit concepts (classes, fields, properties etc.)
+    /// so that we do not need have a dependency on NUnit from the analyzer project.
+    /// These are constants for NUnit v3 and later.
```
mikkelbu: Should we also mention that the constants are removed in V5?

src/nunit.analyzers/SameAsOnValueTypes/SameAsOnValueTypesAnalyzer.cs
```diff
@@ -54,7 +60,8 @@ protected override void AnalyzeAssertInvocation(OperationAnalysisContext context
                     var constraintMethod = constraintPartExpression.GetConstraintMethod();
 
                     if (constraintMethod?.Name != NUnitFrameworkConstants.NameOfIsSameAs
-                        || constraintMethod.ReturnType?.GetFullMetadataName() != NUnitFrameworkConstants.FullNameOfSameAsConstraint)
+                        || constraintMethod.ReturnType?.GetFullMetadataName()
+                                                       .StartsWith(NUnitPreV5FrameworkConstants.FullNameOfSameAsConstraint, StringComparison.Ordinal) is not true)
```
mikkelbu, line 64: Just curious. Is there a reason for changing `!=` to "not" .StartsWith ? Is it the stringcomparison or is there another reason?

src/nunit.analyzers.codefixes/Helpers/CodeFixHelper.cs
```diff
@@ -270,25 +271,46 @@ internal static SyntaxNode UpdateTestMethodSignatureIfNecessary(
...
+            bool isVoidReturnType = methodReturnType is PredefinedTypeSyntax predefinedType &&
+                                    predefinedType.Keyword.IsKind(SyntaxKind.VoidKeyword);
+
+            var task = SyntaxFactory.Identifier(nameof(Task));
+            TypeSyntax updatedReturnType =
+                isVoidReturnType ?
+                SyntaxFactory.IdentifierName(task) :
+                SyntaxFactory.GenericName(task,
+                SyntaxFactory.TypeArgumentList(
+                    SyntaxFactory.SeparatedList(new TypeSyntax[] { methodReturnType })));
```
mikkelbu: Personally, I think I would prefer that the `?` and the `:` align - so something like

```suggestion
            TypeSyntax updatedReturnType = isVoidReturnType
                ? SyntaxFactory.IdentifierName(task)
                : SyntaxFactory.GenericName(task,
                    SyntaxFactory.TypeArgumentList(
                        SyntaxFactory.SeparatedList([methodReturnType])));
```

src/nunit.analyzers.codefixes/TaskReturnShouldBeUsed/TaskReturnShouldBeUsedCodeFix.cs
```diff
@@ -0,0 +1,136 @@
...
+            else if (parent is AssignmentExpressionSyntax assignmentExpression)
+            {
+                updatedParent = assignmentExpression.WithRight(updatedExpression);
+            }
+            else
+            {
+                updatedParent = parent.ReplaceNode(invocationExpression, updatedExpression);
+            }
+            */
+
```
mikkelbu: Is this a leftover? 

```suggestion
```

mikkelbu (commented):
I still need to review the new analyzer and codefix and their associated test files, but I ran out of time today. So far only very minor comments

src/nunit.analyzers/Constants/NUnitPreV5FrameworkConstants.cs
manfred-brands: Changed to: `These are constants for NUnit v3 and v4 but were removed from v5.`

src/nunit.analyzers/SameAsOnValueTypes/SameAsOnValueTypesAnalyzer.cs
manfred-brands: The reason is that in V5 the `SameAsConstraint` is only generic, which has a name "SameAsConstraint\`1".
To work with both V3-4 and V4-5, I choose to use StartsWith("SameAsConstraint") to cover the trailing "`1"
The Generic version was introduces in V4.6

src/nunit.analyzers.codefixes/Helpers/CodeFixHelper.cs
manfred-brands: I have no problem with that,

src/nunit.analyzers.codefixes/TaskReturnShouldBeUsed/TaskReturnShouldBeUsedCodeFix.cs
manfred-brands: Yes it was when I thought I had to do this in 2 steps.

src/nunit.analyzers.codefixes/TaskReturnShouldBeUsed/TaskReturnShouldBeUsedCodeFix.cs
```diff
@@ -0,0 +1,118 @@
...
+        private static SyntaxNode UseTaskResult(SyntaxNode root, InvocationExpressionSyntax invocationExpression, ImmutableDictionary<string, string?> properties)
+        {
+            ExpressionSyntax updatedExpression;
+
+            if (properties.ContainsKey(AnalyzerPropertyKeys.IsTaskT))
+            {
+                updatedExpression = SyntaxFactory.MemberAccessExpression(
+                    SyntaxKind.SimpleMemberAccessExpression,
+                    invocationExpression,
+                    SyntaxFactory.IdentifierName(nameof(Task<string>.Result)));
```
mikkelbu, line 98:
```suggestion
                    SyntaxFactory.IdentifierName(nameof(Task<>.Result)));
```

src/nunit.analyzers/TaskReturnShouldBeUsed/TaskReturnShouldBeUsedAnalyzer.cs
```diff
@@ -0,0 +1,169 @@
...
+                                       or IInvocationOperation
+                                       or IPropertyReferenceOperation
+                                       or IAssignmentOperation
+                                       or IReturnOperation)
+            {
+                // The method either does not return Task/Task<T>,
+                // or the invocation is already being awaited,
+                // or the invocation is having .Wait() or .Result called on it.
+                // or the invocation is being assigned to a variable of type Task/Task<T>.
+                // either way we don't need to report a diagnostic.
```
mikkelbu, line 58: Should we also mention the returnoperation here for completeness

src/nunit.analyzers/TaskReturnShouldBeUsed/TaskReturnShouldBeUsedAnalyzer.cs
```diff
@@ -0,0 +1,169 @@
...
+                    // We don't want to report diagnostics for unlikely cases like:
+                    //      var task = Assert.ThrowsAsync<Exception>(() => SomeMethodAsync());
+                    //      Exception? exception = await task;
+                    AnalyzerConfigOptions options = context.Options.AnalyzerConfigOptionsProvider.GetOptions(variableDeclarationSyntax.SyntaxTree);
+                    if (!options.TryGetValue("dotnet_diagnostic.NUnit2059.raise_for_var_declaration_containing", out string? include))
+                    {
+                        // This is an heuristic to fix cases like:
+                        // var ex = Assert.ThrowsAsync<Exception>(() => SomeMethodAsync());
+                        //      var exceptionTask = Assert.ThrowsAsync<Exception>(() => SomeMethodAsync());
+                        //      Exception? exception = await exceptionTask;
```
mikkelbu: Should both cases be here? If so can we make it more obvious that it is 2 cases

src/nunit.analyzers/TaskReturnShouldBeUsed/TaskReturnShouldBeUsedAnalyzer.cs
```diff
@@ -0,0 +1,169 @@
...
+
+                    if (!options.TryGetValue("dotnet_diagnostic.NUnit2059.do_not_raise_for_var_declaration_containing", out string? exclude))
+                    {
+                        // This is an heuristic to fix cases like:
+                        //      var exceptionTask = Assert.ThrowsAsync<Exception>(() => SomeMethodAsync());
+                        //      Exception? exception = await exceptionTask;
+                        exclude = "task";
+                    }
+
+                    // No case-insentive contains on netstandard2.0, nut IndexOf has an overload with StringComparison argument
```
mikkelbu:
```suggestion
                    // No case-insentive contains on netstandard2.0, but IndexOf has an overload with StringComparison argument
```

mikkelbu (approved):
Again, only nitpick and questions. Otherwise it looks great and the comments are very helpful when reading the code 👍. Happy to merge this when you are

src/nunit.analyzers.codefixes/TaskReturnShouldBeUsed/TaskReturnShouldBeUsedCodeFix.cs
manfred-brands: [Unbound Generic in nameof](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-14.0/unbound-generic-types-in-nameof) is only supported in C#14.
We are still using the .NET 8 SDK due to errors in nunit test when moving to .NET 10 SDK.
And also because of appveyor.

src/nunit.analyzers/TaskReturnShouldBeUsed/TaskReturnShouldBeUsedAnalyzer.cs
manfred-brands: Thanks. Yes, the return operation was as result of a Copilot review, but I forgot to update the comments.

src/nunit.analyzers/TaskReturnShouldBeUsed/TaskReturnShouldBeUsedAnalyzer.cs
manfred-brands: Thanks. I had copied the comment when I added the additional configuration, but forgot to remove the 2nd case.

src/nunit.analyzers/TaskReturnShouldBeUsed/TaskReturnShouldBeUsedAnalyzer.cs
manfred-brands: Done.

src/nunit.analyzers.codefixes/TaskReturnShouldBeUsed/TaskReturnShouldBeUsedCodeFix.cs
manfred-brands: Yes VS suggest it because VS ignores `global.json` and always uses the SDK as shipped with the install.
I had a similar issues for the `Compare(string, StringComparison)` which somehow is available for netstandard2.0 when compiling with .NET 10 SDK (It pick an `IReadOnlySpan<char>` overload from System.Memory) but not from the CLI.

manfred-brands:
@mikkelbu as always thanks for your review.

PR #1000, opened 2026-06-29
Analyze more statements to validate variables asserted to be not null

### LINKED ISSUE #998: Assert.EnterMultipleScope clears nullability suppress info
I have a sequence in test:
1. Check model is not null
2. Check multiple model properties
3. Do some other checks on model

When I use `Assert.EnterMultipleScope()` for checking multiple properties in step 2, NUnit thinks that model can be null in step 3.

Thus I get the `warning CS8602: Dereference of a possibly null reference.` warning, although it's not possible. It only happens when there is  `Assert.EnterMultipleScope()` section between null check and subsequent access to the property. The warning is displayed only in build, not in Visual Studio. As I understand it, the NUnit analyzer NUnit3001 should suppress it but for some reason doesn't.

Image

It doesn't happen when:
* I use `Assert.Multiple(() => {})`.
* Move checks from step 3 before step 2

I am using the most recent versions:
* Microsoft.NET.Test.Sdk: 18.6.0
* NUnit: 4.6.1
* NUnit.Analyzers: 4.14.0
* NUnit3TestAdapter: 6.2.0

Here is the demo code:
```csharp
public class Model
{
    public required int Property { get; init; }
}

[TestFixture]
public class NullableTests
{
    [Test]
    public void IssueDemonstration()
    {
        var model = GetSut();

        Assert.That(model, Is.Not.Null);

        {
        }

        Assert.That(model.Property, Is.GreaterThan(1));
    }

    public Model? GetSut()
    {
        return new Model { Property = 5 };
    }
}
```

manfred-brands:
Fixes #998

### DIFF
--- a/src/nunit.analyzers.tests/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressorTests.cs
+++ b/src/nunit.analyzers.tests/DiagnosticSuppressors/DereferencePossiblyNullReferenceSuppressorTests.cs
@@ -794,6 +794,7 @@ namespace NUnit.Analyzers.Tests.DiagnosticSuppressors
                     var e = GetPossibleException();
                     try
                     {
+                        MethodThatMightThrowAnException();
                         Assert.That(e, Is.Not.Null);
                     }
                     catch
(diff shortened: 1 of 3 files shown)


mikkelbu (approved):
LGTM - thanks for the improvement @manfred-brands. I'm happy to merge this when you are

PR #1014, opened 2026-08-05
NonNullableFieldOrPropertyIsUninitializedSuppressor crashed on non-class declarations

### LINKED ISSUE #1013: Use of `records` crash the `NonNullableFieldOrPropertyIsUninitializedSuppressor` analyzer
I have `NUnit` along with `NUnit.Analyzers` version `4.14.0` in a .NET 8 project. I am getting the following in the output when building:

```
6>CSC : warning AD0001: Analyzer 'NUnit.Analyzers.DiagnosticSuppressors.NonNullableFieldOrPropertyIsUninitializedSuppressor' threw an exception of type 'System.InvalidOperationException' with message 'Sequence contains no elements'.
```

It appears to be because of the project's use of `record`, and records that aren't directly related to the test fixture. Adding the following to `NonNullableFieldOrPropertyIsUninitializedSuppressorTests` demonstrates this:

```cs
        [Test]
        public async Task FieldInNonTestRecordDoesNotThrowException()
        {
            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing(@"
            #nullable enable

            public record DataRecord
            {
                public string Field;
            }");

            Exception? analyzerException = null;
            var options = new CompilationWithAnalyzersOptions(
                new AnalyzerOptions(ImmutableArray<AdditionalText>.Empty),
                (exception, _, _) => analyzerException = exception,
                concurrentAnalysis: false,
                logAnalyzerExecutionTime: false,
                reportSuppressedDiagnostics: true);

            Compilation compilation = TestHelpers.CreateCompilation(testCode);
            await compilation
                .WithAnalyzers(ImmutableArray.Create<DiagnosticAnalyzer>(suppressor), options)
                .GetAllDiagnosticsAsync()
                .ConfigureAwait(true);

            Assert.That(analyzerException, Is.Null);
        }
```

A possible fix for 

https://github.com/nunit/nunit.analyzers/blob/8a97c92d93a365b61a177d689bc0ab31da06cc88/src/nunit.analyzers/DiagnosticSuppressors/NonNullableFieldOrPropertyIsUninitializedSuppressor.cs#L73

would be:

```cs
                TypeDeclarationSyntax? typeDeclaration = node.Ancestors().OfType<TypeDeclarationSyntax>().FirstOrDefault();
                if (typeDeclaration is not ClassDeclarationSyntax classDeclaration)
                {
                    continue;
                }
```

The `TypeDeclarationSyntax` instead of `ClassDeclarationSyntax` will keep the walk up the ancestors from skipping over the actual enclosing type of a field. Supporting `TestFixtures` attribute on a `record` would require more than this fix.

manfred-brands:
Fixes #1013 

The analyzer now works on `record` fixtures.

Also updates NonTestMethodAccessibilityLevelAnalyzer to ignore compiler generated public methods.


src/nunit.analyzers.tests/DiagnosticSuppressors/NonNullableFieldOrPropertyIsUninitializedSuppressorTests.cs
```diff
@@ -522,5 +533,18 @@ public void Test()
             Settings settings = Settings.Default.WithAnalyzerConfig(analyzerConfig);
             RoslynAssert.Suppressed(suppressor, testCode, settings);
         }
+
+        [TestCaseSource(nameof(TypeDeclarations))]
+        public void ShouldIgnoreUnrelatedTypeDeclarations(string type)
+        {
+            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing($$"""
+                public {{type}} TestType
```
mikkelbu: For consistency with the other tests

```suggestion
        public void ShouldIgnoreUnrelatedTypeDeclarations(string typeDeclaration)
        {
            var testCode = TestUtility.WrapClassInNamespaceAndAddUsing($$"""
                public {{typeDeclaration}} TestType
```

mikkelbu (approved):
LGTM - only the smallest of suggestions. I'm happy to merge this when you are

Issue #1019, opened 2026-09-09
NUnit1027 does not see [CancelAfter] inherited from a base fixture


ramonsmits:
NUnit 4 allows `[CancelAfter]` on a class and marks it `Inherited = true`, and the framework applies it to every test in derived fixtures: `CompositeWorkItem` picks it up through `Test.GetCustomAttributes<IApplyToContext>(true)`, so `TestExecutionContext.TestCaseTimeout` and `UseCancellation` flow down and each test method gets a `CancelAfterCommand` and a filled-in `CancellationToken` parameter.

The analyzer only looks at the method and its direct containing type:

```csharp
// TestMethodUsageAnalyzer.cs, TestCaseUsageAnalyzer.cs (4.14.0)
var hasCancelAfterAttribute = methodAttributes.Any(a => SymbolEqualityComparer.Default.Equals(a.AttributeClass, cancelAfterType)) ||
    methodSymbol.ContainingType.GetAttributes().Any(a => SymbolEqualityComparer.Default.Equals(a.AttributeClass, cancelAfterType));
```

So with the attribute on a base class, every derived test with a trailing `CancellationToken` parameter reports NUnit1027 (and NUnit1029 for `[TestCase]` methods), even though NUnit supplies the token at runtime.

**Repro**

```csharp
[CancelAfter(180_000)]
public abstract class AcceptanceTest { }

public class MyTests : AcceptanceTest
{
    [Test]
    public async Task Works(CancellationToken cancellationToken) // NUnit1027: The test method has '1' parameter(s), but only '0' argument(s) are supplied by attributes
    {
        await Task.Delay(10, cancellationToken);
    }

    [TestCase(true)]
    public async Task Works_with_cases(bool flag, CancellationToken cancellationToken) // NUnit1029
    {
        await Task.Delay(10, cancellationToken);
    }
}
```

Both tests run and receive a cancellable token bounded at 180 s.

**Expected**

The lookup walks the base types, matching `CancelAfterAttribute`'s `Inherited = true`, so a base-class `[CancelAfter]` counts the same as one on the fixture itself.

**Versions**

NUnit 4.6.1, NUnit.Analyzers 4.14.0, .NET 10.

**Workaround**

Suppress NUnit1027 in the affected projects, or repeat `[CancelAfter]` on every derived fixture.

PR #1020, opened 2026-09-09
Recognize [CancelAfter] inherited from a base class

ramonsmits:
Fixes:

- #1019

`CancelAfterAttribute` is `Inherited = true` and NUnit applies it to every test in derived fixtures, but `TestMethodUsageAnalyzer` and `TestCaseUsageAnalyzer` only checked the method and its direct containing type. A test with a `CancellationToken` parameter therefore reported NUnit1027 or NUnit1029 when the attribute sat on a base class.

Both lookups now use the existing `GetAllAttributes()`, which walks the base types the same way the `[TestFixture]` and `[FixtureLifeCycle]` checks already do. One test per analyzer added; both fail without the change.


manfred-brands (approved):
Thanks @ramonsmits for both your ticket and the fix.

